PR #230, opened 2013-05-27
Support HTTP 201 Semantics

adriancole:
[HTTP 201 Created](http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html) returns the `Location` header of a new resource.  This pull request allows us to receive and use URI data in response to restful services that use 201 semantics.

ex.
```
/**
 * returns id of the new resource created.
 */
@POST("/") URI create();
@GET("/") Thing get(URI id);
```

### DIFF
--- a/retrofit/src/main/java/retrofit/RestAdapter.java
+++ b/retrofit/src/main/java/retrofit/RestAdapter.java
@@ -21,6 +21,7 @@ import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
 import java.lang.reflect.Type;
+import java.net.URI;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.concurrent.Executor;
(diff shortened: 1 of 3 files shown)


gitamith:
Sorry to comment on an old PR, but is there a way to handle HTTP 201 with the 2.0.0 version?

JakeWharton:
Read the header, define a `@Url` parallel method for the follow-up request.

gitamith:
`JsonReader` throws an EOFException when the response is received since the 201 has an empty content body and we dont get a chance to read the headers. I saw that empty responses where handled for HTTP 204 and 205 elsewhere. Is there a workaround for 201s?

JakeWharton:
Gotcha. Nope. We can fix it though

JakeWharton:
Does the RFC mandate empty response bodies for 201?

gitamith:
> The response SHOULD include an entity containing a list of resource characteristics and location(s) from which the user or user agent can choose the one most appropriate. The entity format is specified by the media type given in the Content-Type header field.

SHOULD as opposed to MUST for all the mandatory stuff

JakeWharton:
I'll try to look at it this week! Thanks.

gitamith:
OK..thanks!

JakeWharton:
I didn't even realize this was a PR you were commenting on (I was replying via email).

If the response has no body you can use the `Void` type which will immediately close it without trying to go through any converter. I'll note that this was added after the first beta so it will be in the second one (hopefully next week). Everything else from this PR is already supported in v2.

gitamith:
Ok. I was trying with the first beta. `Void` would solve my problem. Thanks!

PR #268, opened 2013-07-11
implement Retryer for transient error resolution

codefromthecrypt:
Allows you to throw a `RetryableError` from your ErrorHandler when a response implies a retry.

Ex.

```
new ErrorHandler() {
  @Override public Throwable handleError(RetrofitError cause) {
    Response r = cause.getResponse();
      if (r.getStatus() == 500) {
        return new RetryableError(cause);
      }
      return cause;
    }
};
```

Default `Retryer` exponentially backs off up to 5 times, unless the `retryAfter` field is set in the `RetryableError`.


WingSafari5:
Any status on this?

PR #406, opened 2014-02-08
@Timeout added

celiovasconcelos:
Now It’s possible to setup a fine grained request timeout in a method level. Small change, huge feature.
If the annotation is not present, everything works normally as before.


erseno:
Hello! I came across this looking to implement separate timeouts for specific requests. What was the outcome of this in the end? Is there an example of using an interceptor to set a timeout for a certain request? Thanks!

PR #613, opened 2014-09-30
Support multivalues in @QueryMap and @FieldMap parameters.

s-urbaniak:
Hi,

We propose the following pull-request which adds the possibility to use a Map<?, Iterable<?>> for query and/or field parameters.

Background: We have quite complex HTTP parameters and thus use Guava's Multimaps for manipulating those. Passing such values using Multimap.asMap() to Retrofit would be very convienient.

Retrofit already supports Iterables for regular Query parameters. This pull-request extends this semantics for QueryMaps too.

Thanks,
Sergiusz Urbaniak & Alex Thurm


matiwinnetou:
+1

Carthan:
+1

PR #655, opened 2014-11-12
Add SimpleCallback for lazzy

yongjhih:
For example:

```java
service.createUser(user, new SimpleCallback<User>()); // No need implementation if you are lazzy
```


JakeWharton:
This interface will be going away to be replaced by another mechanism shortly.

PR #674, opened 2014-12-05
Added null-check for cases when a null body is being logged

Egorand:
I came across a NullPointerException which resulted from an attempt to log a null body. When using Retrofit with Rx, it results in whole request failing and onError() being called; turning logs off solved this problem.


JakeWharton:
Logging has changed substantially since this PR as we are hard at work on version 2.0. Thanks for identifying and submitting the fix, though!

Egorand:
No problem, happy to contribute!

PR #676, opened 2014-12-11
Greeno/extend interface option

greeno:
Addresses issue in https://github.com/square/retrofit/issues/504

### DIFF
--- a/retrofit-mock/src/main/java/retrofit/MockRestAdapter.java
+++ b/retrofit-mock/src/main/java/retrofit/MockRestAdapter.java
@@ -209,7 +209,7 @@ public final class MockRestAdapter {
    */
   @SuppressWarnings("unchecked")
   public <T> T create(Class<T> service, T mockService) {
-    Utils.validateServiceClass(service);
+    Utils.validateServiceClass(false, service);
     return (T) Proxy.newProxyInstance(service.getClassLoader(), new Class<?>[] { service },
         new MockHandler(mockService, restAdapter.getMethodInfoCache(service)));
   }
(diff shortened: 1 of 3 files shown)


podollb:
This would be great!

Nthalk:
Agreed! I have a large api that I was hoping to use this for, and the lack of composable interfaces make this a deal breaker for me!

JakeWharton:
@Nthalk Retrofit favors composition. This PR adds inheritance.

Nthalk:
We are talking about interfaces here. Especially a situation where the body methods are provided by a declarative convention.

I find it hard to imagine that any of the inheritance vs composition pitfalls are here.

Also, because of the requirement that the source must be an interface, this also precludes delegation at the definition.

Do you really have a single adapter with a ton of proxies?

JakeWharton:
Yes. They are generated from a service declaration in protos. One interface per service.

carek:
In my opinion this validation breaks Single Responsible Principal. Could you please reconsider this? :)

swankjesse:
I don't think this is a good fit for Retrofit's API.

ashneo76:
What is the reason for closing this PR without even explaining a possible solution using your proposed approach?

kennymacleod:
Please reconsider this.

My use case is that I have an interface that defines a set of operations without any Retrofit annotations. This interface is agnostic to the remoting layer. I then have two extensions of this interface, which annotates the methods with either Retrofit or RemotingTechnologyX.

Retrofit currently forbids this, for no reason other than composition should be encouraged over inheritance. While I agree with the principle, enforcing it like this is a blunt hammer than ignores the perfectly valid use cases for using interface inheritance.

JakeWharton:
And, you know, the fact that it doesn't work on some versions of Android.

PR #694, opened 2015-01-04
Bump Gson dependency

alexxv:
Update Gson to 2.3.1 to fix
https://code.google.com/p/google-gson/issues/detail?id=596


JakeWharton:
This is fine but there won't be a release for months. Bump the dependency in your builds if you need a newer Gson.

ChristianBecker:
@JakeWharton What about the upcoming 1.9 release? Will this be included? (https://twitter.com/JakeWharton/status/550145655499272193)

PR #712, opened 2015-01-21
Including Ekstazi (ekstazi.org) profile to optimize execution of the tests

andzaytsev:
Ekstazi allows to speed up regression testing by only executing tests that were affected by the changes that were made since the last test run. To run the tests with Ekstazi with my patch to pom.xml, one should use ```mvn test -Pekstazi```.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -169,6 +169,38 @@
     </dependencies>
   </dependencyManagement>
 
+  <profiles>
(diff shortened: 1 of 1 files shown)


JakeWharton:
Our tests are small in scope and all of them run in a few seconds. I simply don't think there's enough value here, especially since it's opt-in, and I don't want to have to maintain two system. Thanks for the PR and introducing the product though.

PR #718, opened 2015-01-27
Changing constructor and internal class visibility to protected

pthomain:
Currently the constructor is private to enforce the use of the builder, however this prevents the extension of the class, making it de facto final. Changing the visibility to protected would still require to use the builder while allowing for developers to extend the class to provide custom caching solutions for instance.


JakeWharton:
This class was not at all designed for extensibility and I'm not sure why it's not `final` (probably legacy cruft). If you want custom caching it should be done at the application layer (above Retrofit) or in the `Client` or HTTP client (below Retrofit). The freedom to change the implementation details of this class is important (and has been important over the last three years).

PR #721, opened 2015-01-29
Switch the Client to control the threading model for execution.

JakeWharton:
This makes all requests async. Synchronous requests simply block on their thread.


JakeWharton:
I haven't self-reviewed this in its entirety so LGTMs only, please.

adriancole:
LGTM Basically this trades the simplicity of a single method interface for micro-optimizations which might materialize from having a client support both sync and async invocation.

pforhan:
LGTM

PR #724, opened 2015-01-30
Remove the need for two Endpoint classes.

JakeWharton:


retrofit/src/main/java/retrofit/Utils.java
```diff
@@ -29,6 +29,12 @@
 final class Utils {
   private static final int BUFFER_SIZE = 0x1000;
 
+  static <T> T checkNotNull(T object, String message, Object... args) {
+    if (args.length > 0) message = String.format(message, args);
```
mattprecious: Don't need to format the message unless `object` is null.

mattprecious:
LGTM

PR #727, opened 2015-01-30
Pass request body type to converter for serialization.

### LINKED ISSUE #571: not convert interface Object to JSON
I have a interface object `Location`,  provide `getNation`, `getProvince`, `getDistrict` method, only getter, no setter. But I can't seriealize the class Implemented the interface to the server.

LocationAdapter.java
```
public class LocationAdapter extends TypeAdapter<Location>{

	@Override
	public void write(JsonWriter out, Location value) throws IOException {
		System.out.print(".... write object....");
		out.beginObject()
		.name("city").value(value.getCity())
		.name("district").value(value.getDistrict())
		.name("nation").value(value.getNation())
		.name("provider").value(value.getProvider())
		.endObject();
	}
}
```

API define
```
@POST("/user/geo")
public void updateLocation(@Body Location location, Callback<Response> callback);
```

Create service
```
Gson gson = new GsonBuilder()
		.registerTypeAdapter(Location.class, new LocationAdapter().nullSafe())
		.create();


	RestAdapter restAdapter = new RestAdapter.Builder()
  	    .setConverter(new GsonConverter(gson))
	    .build();

```

### LINKED ISSUE #713: JacksonConverter get the wrong type
class Foo<T> {
	   T bars;
	}

	class FooBarList extends Foo<List<Bar>>

	class Bar extends IBar {
	   ...
	}

	abstract class IBar extends Comparator<IBar>{
	   ...
	}

The exception below is thrown when I try to deserialize my json to `FooBarList`
   
    com.fasterxml.jackson.databind.JsonMappingException: Can not construct instance of IBar, problem: abstract types either need to be mapped to concrete types, have custom deserializer, or be instantiated with additional type information

    (through reference chain: FooBarList["bars"])

I found out that the problem is because of my `IBar extends Comparator<IBar>`.

Jackson detects that the class type of `Bar` is `IBar` which is in the `Comparator<IBar>` and it tries to use `IBar` to construct instance and fails.

When I switch `Comparator<IBar>` to `Comparator<AnotherAbstractClass>`, error of `com.fasterxml.jackson.databind.JsonMappingException: Can not construct instance of AnotherAbstractClass` is thrown. 

When I remove the `Comparator`, it works. So, I am sure that this is the cause.

I have a thread in stackoverflow, http://stackoverflow.com/questions/28043518/jackson-jsonmappingexception-can-not-construct-instance-abstractclass-due-to-com

JakeWharton:
@swankjesse 

Closes #571. Closes #713.

### DIFF
--- a/retrofit-converters/jackson/src/main/java/retrofit/converter/JacksonConverter.java
+++ b/retrofit-converters/jackson/src/main/java/retrofit/converter/JacksonConverter.java
@@ -6,6 +6,7 @@ import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Type;
 import retrofit.mime.TypedByteArray;
@@ -31,21 +32,31 @@ public class JacksonConverter implements Converter {
   }
 
   @Override public Object fromBody(TypedInput body, Type type) throws ConversionException {
+    InputStream in = null;
     try {
       JavaType javaType = objectMapper.getTypeFactory().constructType(type);
-      return objectMapper.readValue(body.in(), javaType);
+      in = body.in();
+      return objectMapper.readValue(in, javaType);
     } catch (JsonParseException e) {
       throw new ConversionException(e);
     } catch (JsonMappingException e) {
       throw new ConversionException(e);
     } catch (IOException e) {
       throw new ConversionException(e);
+    } finally {
+      try {
+        if (in != null) {
+          in.close();
+        }
+      } catch (IOException ignored) {
+      }
     }
   }
 
-  @Override public TypedOutput toBody(Object object) {
+  @Override public TypedOutput toBody(Object object, Type type) {
     try {
-      String json = objectMapper.writeValueAsString(object);
+      JavaType javaType = objectMapper.getTypeFactory().constructType(type);
+      String json = objectMapper.writerWithType(javaType).writeValueAsString(object);
       return new TypedByteArray(MIME_TYPE, json.getBytes("UTF-8"));
     } catch (JsonProcessingException e) {
       throw new AssertionError(e);
--- a/retrofit/src/main/java/retrofit/RestMethodInfo.java
+++ b/retrofit/src/main/java/retrofit/RestMethodInfo.java
@@ -71,6 +71,7 @@ final class RestMethodInfo {
   // Method-level details
   final ExecutionType executionType;
   Type responseObjectType;
+  Type requestObjectType;
   RequestType requestType = RequestType.SIMPLE;
   String requestMethod;
   boolean requestHasBody;
@@ -291,7 +292,7 @@ final class RestMethodInfo {
    * Loads {@link #requestParamAnnotations}. Must be called after {@link #parseMethodAnnotations()}.
    */
   private void parseParameters() {
-    Class<?>[] methodParameterTypes = method.getParameterTypes();
+    Type[] methodParameterTypes = method.getGenericParameterTypes();
 
     Annotation[][] methodParameterAnnotationArrays = method.getParameterAnnotations();
     int count = methodParameterAnnotationArrays.length;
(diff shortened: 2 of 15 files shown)


retrofit/src/main/java/retrofit/converter/Converter.java
```diff
@@ -44,5 +44,5 @@
    * @param object Object instance to convert.
    * @return Representation of the specified object as bytes.
    */
-  TypedOutput toBody(Object object);
+  TypedOutput toBody(Object object, Type type);
```
swankjesse: Big API change here.

retrofit-converters/jackson/src/main/java/retrofit/converter/JacksonConverter.java
```diff
@@ -31,21 +32,31 @@ public JacksonConverter(ObjectMapper objectMapper) {
...
     } catch (JsonParseException e) {
       throw new ConversionException(e);
     } catch (JsonMappingException e) {
       throw new ConversionException(e);
     } catch (IOException e) {
       throw new ConversionException(e);
+    } finally {
+      try {
+        if (in != null) {
+          in.close();
```
swankjesse: yay!

retrofit/src/main/java/retrofit/RestMethodInfo.java
```diff
@@ -291,7 +292,7 @@ private static Type getParameterUpperBound(ParameterizedType type) {
    * Loads {@link #requestParamAnnotations}. Must be called after {@link #parseMethodAnnotations()}.
    */
   private void parseParameters() {
-    Class<?>[] methodParameterTypes = method.getParameterTypes();
+    Type[] methodParameterTypes = method.getGenericParameterTypes();
```
swankjesse: Nice.

retrofit/src/main/java/retrofit/converter/GsonConverter.java
```diff
@@ -75,9 +75,10 @@ public GsonConverter(Gson gson, String charset) {
     }
   }
 
-  @Override public TypedOutput toBody(Object object) {
+  @Override public TypedOutput toBody(Object object, Type type) {
     try {
-      return new JsonTypedOutput(gson.toJson(object).getBytes(charset), charset);
+      String json = gson.toJson(object, type);
```
swankjesse: Slight potential for a behavior change here. For example, if `Type` is an interface, Gson may very well freak out attempting to find a converter.

Regardless, I think this is the right behavior.

retrofit/src/test/java/retrofit/converter/GsonConverterTest.java
```diff
@@ -0,0 +1,75 @@
...
+import org.junit.Test;
+import retrofit.mime.TypedOutput;
+import retrofit.mime.TypedString;
+
+import static org.junit.Assert.assertEquals;
+
+public final class GsonConverterTest {
+  private Converter converter;
+
+  interface Example {
```
swankjesse: Nice.

swankjesse:
Very nice change. I only wish it'd worked like this from the beginning. You'll need to warn users when they take this update; it has the potential to inflict big pain!

```
    assertEquals("\"foo\"", new Gson().toJson("foo", Object.class));
    assertEquals("\"foo\"", new Gson().toJson("foo", String.class));
    assertEquals("{}", new Gson().toJson("foo", CharSequence.class));
    assertEquals("{}", new Gson().toJson("foo", Comparable.class));
```

Yes, I agree that this is absurd, and makes Gson look criminally dumb.

One other thing we could think about; perhaps here or perhaps for 2.0: Gson has a mechanism to lookup a `TypeAdapter` for a given type ahead of time. Would be nice to do that work, which saves some map lookups and permits fail-fast if the type adapter isn't available.

swankjesse:
LGTM

JakeWharton:
> perhaps here or perhaps for 2.0

This is towards 2.0!

> Would be nice to do that work, which saves some map lookups and permits fail-fast if the type adapter isn't available.

The problem there is the converter agnostic behavior of Retrofit. We could certainly add APIs to `Converter` for eager validation of body type, though.

JakeWharton:
> makes Gson look criminally dumb

We have the power to fix this in [redacted] :grinning:

PR #729, opened 2015-01-31
Get rid of our own Header class.

JakeWharton:
@swankjesse


pforhan:
LGTM.

retrofit/src/main/java/retrofit/client/Request.java
```diff
@@ -56,8 +51,8 @@ public String getUrl() {
     return url;
   }
 
-  /** Returns an unmodifiable list of headers, never {@code null}. */
-  public List<Header> getHeaders() {
+  /** Returns an unmodifiable list of headers. May be {@code null}. */
```
swankjesse: nits: its _always_ unmodifiable, and it's not a list

non-nit: return empty headers instead of null?

swankjesse:
Nice!

retrofit/src/main/java/retrofit/client/Request.java
JakeWharton: Yep it's never null. Will fix.

JakeWharton: Actually this class is doing to die. So I won't fix, I'll nuke the whole
thing from orbit.

PR #731, opened 2015-02-01
Simplify custom HTTP method implementation.

JakeWharton:
This is less cute, and more conducive to annotation processing.


adriancole:
lgtm

PR #757, opened 2015-02-20
Fixing comment to address current support of ErrorHandler

dallasgutauckis:
Related to square/retrofit#439 and square/retrofit#482

### DIFF
--- a/retrofit/src/main/java/retrofit/RestAdapter.java
+++ b/retrofit/src/main/java/retrofit/RestAdapter.java
@@ -415,8 +415,8 @@ public class RestAdapter {
     }
 
     /**
-     * The error handler allows you to customize the type of exception thrown for errors on
-     * synchronous requests.
+     * The error handler allows you to customize
+     * {@link retrofit.client.Response response} exceptions.
      */
     public Builder setErrorHandler(ErrorHandler errorHandler) {
       if (errorHandler == null) {


retrofit/src/main/java/retrofit/RestAdapter.java
```diff
@@ -415,8 +415,8 @@ public Builder setConverter(Converter converter) {
     }
 
     /**
-     * The error handler allows you to customize the type of exception thrown for errors on
-     * synchronous requests.
+     * The error handler allows you to customize
+     * {@link retrofit.client.Response response} exceptions.
```
JakeWharton: I don't think this is accurate either. I'm pretty sure we now always apply the error handler, regardless of whether it's synchronous, async, or Rx.

dallasgutauckis: Do you have preferred verbiage? I'd rather fix it in all places and have it be completely clear.

dallasgutauckis: - Looks like `Response` is now from OkHttp, and not in Retrofit.
 - I took a deeper look at the project and it does seem that the errorHandler is used for everything (synchronous, Rx, `Callback`) that at least makes its way to OkHttp which I think hints at how to word this. It seems like any HTTP/Conversion/IO error will be passed through to the internal `handleError` method which calls out to the `ErrorHandler`.

Succinctly: 
> The error handler allows you to customize the type of `Throwable` returned for any `RetrofitError`.

This also happens to be a way one would describe the method in `ErrorHandler` ;-)

>  Throwable handleError(RetrofitError cause);

Thoughts?

PR #766, opened 2015-03-03
DataPreprocessor interface added to RestAdapter to preprocess Objects

k-kojak:
In case of encrypted communication between client and server, Java Generics cannot be used to customize and create type-safe Body-param objects and return values, because encrypted data always will look like the same. (For example a base64 encoded String.)

With this small interface you can wrap all of your Body-param objects before sending to server, and unwrap all of your results before returning it - if you want - and results and params remain type-safe and self-explanatory as intended.

### DIFF
--- a/retrofit/src/main/java/retrofit/DataPreprocessor.java
+++ b/retrofit/src/main/java/retrofit/DataPreprocessor.java
@@ -1,0 +1,47 @@
+package retrofit;
+
+import com.squareup.okhttp.ResponseBody;
+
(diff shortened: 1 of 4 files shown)


egor-n:
Can't this be done with a custom [Converter](https://github.com/square/retrofit/blob/master/retrofit/src/main/java/retrofit/converter/Converter.java)?

k-kojak:
Yes, you are right :)
Sorry for wasting your time...

PR #788, opened 2015-03-19
Changes to Javadoc to remove doclint errors and warnings

hekar:
Noticed the lack of [2.0.0 snapshots](https://oss.sonatype.org/content/repositories/snapshots/com/squareup/retrofit/retrofit/) since Jan. 30th, due to the migration to JDK 8.

The addition of [doclint to JDK 8](http://openjdk.java.net/jeps/172) causes an early failure of the Maven deployment. [See build #1545.2](https://travis-ci.org/square/retrofit/jobs/49275303#L911).

Resolved the `doclint` errors and some warnings. Unfortunately, there are still quite a few warnings. Some of them are due to poor HTML and they are rather easy to fix. Others are due to missing `@param` and `@return` statements. This is frustrating besides it goes against the DRY principle when writing documentation for simple functions.

I think the best solution would be to reconfigure `doclint` and [enable only the necessary groups](https://gist.github.com/hekar/62772af5b0fe0b6731bb). Maybe `html`, `syntax` and `reference`?

Any configuration would have to be done with Maven Profiles, because [the -Xdoclint argument cannot be passed to JDK 7 or lower](http://stackoverflow.com/questions/15886209/maven-is-not-working-in-java-8-when-javadoc-tags-are-incomplete/22981151#22981151).

Does anyone else have any opinions on this?

Thanks
Hekar

### DIFF
--- a/retrofit/src/main/java/retrofit/http/Body.java
+++ b/retrofit/src/main/java/retrofit/http/Body.java
@@ -25,10 +25,7 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
 /**
  * Use this annotation on a service method param when you want to directly control the request body
  * of a POST/PUT request (instead of sending in as request parameters or form-style request
- * body). If the value of the parameter implements {@link retrofit.mime.TypedOutput TypedOutput},
- * the request body will be written exactly as specified by
- * {@link retrofit.mime.TypedOutput#writeTo(java.io.OutputStream)}. If the value does not implement
- * TypedOutput, the object will be serialized using the {@link retrofit.RestAdapter RestAdapter}'s
+ * body). The object will be serialized using the {@link retrofit.RestAdapter RestAdapter}'s
  * {@link retrofit.converter.Converter Converter} and the result will be set directly as the
  * request body.
  * <p>
--- a/retrofit/src/main/java/retrofit/ErrorHandler.java
+++ b/retrofit/src/main/java/retrofit/ErrorHandler.java
@@ -16,7 +16,7 @@
 package retrofit;
 
 /**
- * A hook allowing clients to customize {@link retrofit.client.Response response} exceptions.
+ * A hook allowing clients to customize {@link RetrofitError RetrofitError}.
  *
  * @author Sam Beran sberan@gmail.com
  */
--- a/retrofit/src/main/java/retrofit/RestAdapter.java
+++ b/retrofit/src/main/java/retrofit/RestAdapter.java
@@ -63,9 +63,8 @@ import retrofit.http.Header;
  * <p>
  * The body of a request is denoted by the {@link retrofit.http.Body @Body} annotation. The object
  * will be converted to request representation by a call to
- * {@link retrofit.converter.Converter#toBody(Object) toBody} on the supplied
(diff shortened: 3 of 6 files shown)


JakeWharton:
Great investigation. I wasn't aware that failures in the script would not trigger a failure of the build (it's because the script happens after success was determined).

The only ones that need fixed are the errors. The warnings are just logged but do not fail the upload.

I'm fundamentally opposed to blindly adding `@returns` and `@param` just for the sake of having them. If they are there, they should be providing value that the normal Javadoc description does not otherwise provide.

I think there's two things that should happen here:
 * This pull request is update to only fix the errors (the missing type references)
 * I'll look into changing the script so that failures are reported somewhere.

hekar:
Updated to just fix the errors.

The deployment script does return an exit code of `1`, but it's Travis-CI that's not failing the build.

http://docs.travis-ci.com/user/build-configuration/#Build-Lifecycle
```
The outcome of any of these commands (except after_success, after_failure or after_script) indicates whether or not this build has failed or passed. The standard Unix exit code of "0" means the build passed; everything else is treated as failure.
```

JakeWharton:
Wow, ok. I assumed the exit code would not affect the build result at all. I'll ping Travis CI people.

hekar:
The exit code of scripts run in `after_success` does _not_ affect the build result at all.

If you _want_ the build to fail, then the deployment should be under `script`.

Though, it'd probably make more sense not to fail the build, since only snapshot artifacts are being deployed.

JakeWharton:
Oh yeah. I misread. I actually wouldn't mind it failing the build because we are otherwise unaware of the failure. I'll have a look later.

hekar:
Thought about it some more... And agreed. It makes far more sense to have the build fail when the deployment fails.

I assume adding a `script` section to [travis.yml](https://github.com/square/retrofit/blob/master/.travis.yml) will do the trick:
```
script:
  - mvn test
  - .buildscript/deploy_snapshot.sh
```

By default the only `script` executed for a 'java' project is `mvn test`.

retrofit/src/main/java/retrofit/http/Body.java
```diff
@@ -25,14 +25,12 @@
...
- * body). If the value of the parameter implements {@link retrofit.mime.TypedOutput TypedOutput},
- * the request body will be written exactly as specified by
- * {@link retrofit.mime.TypedOutput#writeTo(java.io.OutputStream)}. If the value does not implement
- * TypedOutput, the object will be serialized using the {@link retrofit.RestAdapter RestAdapter}'s
+ * body). The object will be serialized using the {@link retrofit.RestAdapter RestAdapter}'s
  * {@link retrofit.converter.Converter Converter} and the result will be set directly as the
  * request body.
  * <p>
  * Body parameters may not be {@code null}.
+ * </p>
```
JakeWharton: Delete. This isn't needed.

JakeWharton:
Yep. I think that's the easiest way. Do you want to make that change in this PR?

JakeWharton:
Actually, don't make the Travis change in this PR. I need to make it across about 30 projects so I'd rather it be done all at once with matching commits.

JakeWharton:
Thanks!

PR #810, opened 2015-04-15
Brings back Multipart from the dead!

felipecsl:
All tests should be passing. Re-enabled all Multipart related tests. Adapted old `RequestBuilder` class to work with new `OkHttp` classes. Everything seems to be working fine.
One minor details is that I wasnt able to assert on the Request body `endsWith()`, so I replaced them with `contains()` since we now just have a string instead of a list of parts. Maybe we can inject the boundary into `MultipartBuilder` somehow. Let me know!

### DIFF
--- a/retrofit/src/main/java/retrofit/RequestBuilder.java
+++ b/retrofit/src/main/java/retrofit/RequestBuilder.java
@@ -17,18 +17,23 @@ package retrofit;
 
 import com.squareup.okhttp.Headers;
 import com.squareup.okhttp.MediaType;
+import com.squareup.okhttp.MultipartBuilder;
 import com.squareup.okhttp.Request;
 import com.squareup.okhttp.RequestBody;
+
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Array;
 import java.net.URLEncoder;
 import java.util.Map;
+
 import okio.BufferedSink;
 import retrofit.converter.Converter;
 import retrofit.http.Body;
 import retrofit.http.Header;
+import retrofit.http.Part;
+import retrofit.http.PartMap;
 import retrofit.http.Path;
 import retrofit.http.Query;
 import retrofit.http.QueryMap;
@@ -42,6 +47,7 @@ final class RequestBuilder implements RequestInterceptor.RequestFacade {
   private final boolean async;
   private final String apiUrl;
 
+  private MultipartBuilder multipartBody;
   private RequestBody body;
 
   private String relativeUrl;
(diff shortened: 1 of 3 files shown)


retrofit/src/main/java/retrofit/RequestBuilder.java
```diff
@@ -240,87 +266,90 @@ void setArguments(Object[] args) {
...
+//              throw new IllegalArgumentException(
+//                  "Parameter #" + (i + 1) + " field map contained null key.");
+//            }
+//            Object entryValue = entry.getValue();
+//            if (entryValue != null) { // Skip null values.
+//              formBody.addField(entryKey.toString(), encodeNames, entryValue.toString(),
+//                  encodeValues);
+//            }
+//          }
+//        }
```
JakeWharton: Can you revert the comment changes here? It adds needless noise to the commit which messes with history and prevent cherry-picking and reverting (not that I expect those to need to happen).

retrofit/src/test/java/retrofit/RequestBuilderTest.java
```diff
@@ -1218,372 +1224,370 @@ Response method(@Path("ping") String ping, @Body Object body, @Path("kit") Strin
...
+//        return null;
+//      }
+//    }
+//    try {
+//      buildRequest(Example.class);
+//      fail();
+//    } catch (IllegalArgumentException e) {
+//      assertThat(e).hasMessage("Example.method: @FieldMap parameter type must be Map. (parameter #1)");
+//    }
+//  }
```
JakeWharton: Same here.

JakeWharton:
I think this looks good. I'm not at home for a detailed review, but nothing jumped out.

felipecsl:
Done. 
Build seems to be failing for this reason: `/home/travis/build/square/retrofit/retrofit/src/main/java/retrofit/RequestBuilder.java:218:3: Method length is 151 lines (max allowed is 150).` Should I do anything about it?
Also, do you want me to squash the commits in a single one?

JakeWharton:
@jacobtabak comment on the PR diff view not the commit. No one gets emails for commit comments so I didn't even realize you saw this.

@felipecsl ignore it. I'll disable when I merge. You can squash or I can when I merge.

retrofit/src/test/java/retrofit/RequestBuilderTest.java
```diff
@@ -1218,192 +1224,190 @@ Response method(@Path("ping") String ping, @Body Object body, @Path("kit") Strin
...
+
+    assertThat(bodyString)
+        .contains("name=\"kit\"")
+        .contains("\r\nkat\r\n--");
+  }
+
+  @Test public void multipartWithEncoding() throws IOException {
+    class Example {
+      @Multipart //
+      @POST("/foo/bar/") //
```
jacobtabak: What's up with these empty comments at the end of the line?  I see they're carried over from a previous commit.

JakeWharton: They prevent dumb automatic formatting (collapsing together) when you reformat the whole file.

retrofit/src/main/java/retrofit/RequestBuilder.java
```diff
@@ -282,45 +308,49 @@ void setArguments(Object[] args) {
...
-      //              converter.toBody(entryValue, entryValue.getClass()));
-      //        }
-      //      }
-      //    }
-      //  }
+      } else if (annotationType == Part.class) {
+        if (value != null) { // Skip null values.
+          String name = ((Part) annotation).value();
+          String transferEncoding = ((Part) annotation).encoding();
+          Headers headers = Headers.of(
```
JakeWharton: Nice

JakeWharton:
Fixed, squashed, merged. Thanks!

felipecsl:
:+1: :+1:

PR #812, opened 2015-04-16
Restore Form URL encoding support.

JakeWharton:
@swankjesse @mattprecious

### DIFF
--- a/retrofit/src/main/java/retrofit/FormEncodingBuilder.java
+++ b/retrofit/src/main/java/retrofit/FormEncodingBuilder.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright (C) 2014 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package retrofit;
+
+import com.squareup.okhttp.MediaType;
+import com.squareup.okhttp.RequestBody;
+import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import okio.Buffer;
+import okio.BufferedSink;
+import okio.ByteString;
+
+final class FormEncodingBuilder {
+  private final Buffer content = new Buffer();
+
+  FormEncodingBuilder add(String name, boolean encodeName, String value, boolean encodeValue) {
+    if (content.size() > 0) {
+      content.writeByte('&');
+    }
+    try {
+      if (encodeName) {
+        name = URLEncoder.encode(name, "UTF-8");
+      }
+      if (encodeValue) {
+        value = URLEncoder.encode(value, "UTF-8");
+      }
+    } catch (UnsupportedEncodingException e) {
+      throw new AssertionError(e);
+    }
+    content.writeUtf8(name);
+    content.writeByte('=');
+    content.writeUtf8(value);
+    return this;
+  }
+
+  RequestBody build() {
+    if (content.size() == 0) {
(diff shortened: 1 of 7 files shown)


retrofit/src/main/java/retrofit/FormEncodingBuilder.java
```diff
@@ -0,0 +1,79 @@
...
+      throw new AssertionError(e);
+    }
+    content.writeUtf8(name);
+    content.writeByte('=');
+    content.writeUtf8(value);
+    return this;
+  }
+
+  RequestBody build() {
+    if (content.size() == 0) {
```
swankjesse: Should we add `isEmpty()` ?

retrofit/src/test/java/retrofit/RequestBuilderTest.java
```diff
@@ -77,39 +78,39 @@ Response method(@Body RequestBody body) {
...
+    }
+    try {
+      buildRequest(Example.class);
+      fail();
+    } catch (IllegalArgumentException e) {
+      assertThat(e).hasMessage("Example.method: Only one encoding annotation is allowed.");
+    }
+  }
+
+  @Test public void onlyOneEncodingIsAllowedFormEncodingFirst() {
```
swankjesse: Is annotation order retained?

JakeWharton: Yes. We previously only asserted (and tested) one of these cases in the code before it was discovered that it matters.

retrofit/src/test/java/retrofit/RequestBuilderTest.java
```diff
@@ -1404,186 +1405,186 @@ Response method(@Part("ping") RequestBody ping) {
...
+  @Test public void formEncodedFieldList() {
+    class Example {
+      @FormUrlEncoded //
+      @POST("/foo") //
+      Response method(@Field("foo") List<Object> fields, @Field("kit") String kit) {
+        return null;
+      }
+    }
+
+    List<Object> values = Arrays.<Object>asList("foo", "bar", null, 3);
```
swankjesse: Nice.

swankjesse:
LGTM

retrofit/src/main/java/retrofit/FormEncodingBuilder.java
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright (C) 2014 Square, Inc.
```
mattprecious: 2015?

JakeWharton: This is copy/pasted from OkHttp.

retrofit/src/main/java/retrofit/FormEncodingBuilder.java
JakeWharton: https://github.com/square/okio/pull/140

retrofit/src/main/java/retrofit/RequestBuilder.java
```diff
@@ -74,13 +77,13 @@
     }
 
     switch (methodInfo.requestType) {
-      //case FORM_URL_ENCODED:
+      case FORM_URL_ENCODED:
       //  // Will be set to 'body' in 'build'.
```
jacobtabak: Extra `//` here

retrofit/src/main/java/retrofit/RequestBuilder.java
```diff
@@ -381,6 +379,12 @@ Request build() {
     }
 
     RequestBody body = this.body;
+    if (formEncodingBuilder != null) {
```
jacobtabak: Smells a bit weird that this is an `if`/`else` block, as if form encoding has higher priority than multipart, when in reality they're mutually exclusive.

JakeWharton: Their mutual exclusivity is why it's an `else`. If one matches I don't want to pay the cost of the field lookup for the other.

JakeWharton: What I really want is
```java
RequestBody body = this.body;
if (body == null) {
  FormEncodingBuilder formEncodingBuilder = this.formEncodingBuilder;
  if (formEncodingBuilder != null) {
    body = formEncodingBuilder.build();
  } else {
    MultipartBuilder multipartBuilder = this.multipartBuilder;
    if (multipartBuilder != null) {
      body = multipartBuilder.build();
    }
  }
}
```
but that's far too ugly.

PR #815, opened 2015-04-19
HTTP 204 & 205 should skip the converter.

JakeWharton:
@swankjesse @mattprecious

### DIFF
--- a/retrofit/src/main/java/retrofit/RestAdapter.java
+++ b/retrofit/src/main/java/retrofit/RestAdapter.java
@@ -291,7 +291,12 @@ public final class RestAdapter {
       }
 
       ResponseBody body = response.body();
-      if (body == null) {
+      if (statusCode == 204 || statusCode == 205) {
+        // HTTP 204 No Content "...response MUST NOT include a message-body"
+        // HTTP 205 Reset Content "...response MUST NOT include an entity"
+        if (body.contentLength() > 0) {
+          throw new IllegalStateException(statusCode + " response must not include body.");
+        }
         return null;
       }
 
--- a/retrofit/src/test/java/retrofit/RestAdapterTest.java
+++ b/retrofit/src/test/java/retrofit/RestAdapterTest.java
@@ -12,6 +12,8 @@ import java.util.concurrent.atomic.AtomicReference;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
+import retrofit.converter.Converter;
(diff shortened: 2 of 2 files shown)


retrofit/src/main/java/retrofit/RestAdapter.java
```diff
@@ -290,11 +300,6 @@ private Object parseResult(MethodInfo methodInfo, Response response)
         return response;
       }
 
-      ResponseBody body = response.body();
-      if (body == null) {
```
JakeWharton: This is never true of an OkHttp response.

retrofit/src/main/java/retrofit/RestAdapter.java
```diff
@@ -291,7 +291,12 @@ private Object parseResult(MethodInfo methodInfo, Response response)
       }
 
       ResponseBody body = response.body();
-      if (body == null) {
```
JakeWharton: This is never true of an OkHttp response.

retrofit/src/main/java/retrofit/RestAdapter.java
```diff
@@ -291,7 +291,12 @@ private Object parseResult(MethodInfo methodInfo, Response response)
...
+      if (statusCode == 204 || statusCode == 205) {
+        // HTTP 204 No Content "...response MUST NOT include a message-body"
+        // HTTP 205 Reset Content "...response MUST NOT include an entity"
+        if (body.contentLength() > 0) {
+          throw new IllegalStateException(statusCode + " response must not include body.");
```
swankjesse: Not a protocol exception? Somewhat harsh to crash the client when the server made the mistake.

swankjesse: (not sure what's default in Retrofit)

retrofit/src/test/java/retrofit/RestAdapterTest.java
```diff
@@ -71,6 +79,56 @@
...
+    Response response = example.direct();
+    assertThat(response.code()).isEqualTo(204);
+  }
+
+  @Test public void http204WithBodyThrows() {
+    server.enqueue(new MockResponse().setStatus("HTTP/1.1 204 Nothin").setBody("Hey"));
+    try {
+      example.something();
+      fail();
+    } catch (RetrofitError e) {
```
swankjesse: Ahhh . . . it gets wrapped. Cool.

Issue #817, opened 2015-04-20
Pluggable execution mechanism


JakeWharton:
In another attempt to front-load work in order to not have to do anything in the future, supporting swappable execution mechanisms would allow the next hot framework to come in and plug directly in without needing any core library modifications. Retrofit itself will be modified so that the first-party mechanism is simply a default one (much like other configurable behavior) and the RxJava support will be moved to it as well. This allows support of futures (native or Guava) or even something as gross as AsyncTasks if you really hated yourself through a simple configuration.

SandroMachado:
In my project I really want to use promises, so what I did was:

1º - Add the promises library - https://github.com/darylteo/rxjava-promises/tree/master/rxjava-promises

2º - Create a RetrofitPromise class that extends a Promise and implements the retrofit callback

```java
/**
 * Converts a {@link Callback<T>} into a {@link Promise<T>}.
 *
 * @param <T> The type of the {@link Promise}.
 */

public class RetrofitPromise<T> extends Promise<T> implements Callback<T> {

    /**
     * Constructor.
     */

    public RetrofitPromise() {
    }

    /**
     * Executed when the request is performed with the success.
     *
     * @param responseObject The {@link T} object.
     * @param response The {@link Response}.
     */

    @Override
    public void success(T responseObject, Response response) {
        this.fulfill(responseObject);
    }

    /**
     * Executed when the request gives an error.
     *
     * @param error The {@link retrofit.RetrofitError}.
     */

    @Override
    public void failure(RetrofitError error) {
        this.reject(error.getCause());
    }

}
```

3º Pass the promise in the callback reference

```java
RetrofitPromise<List<Contributor>> promise = new RetrofitPromise<>();
...
GitHub github = restAdapter.create(GitHub.class);
github.contributors("square", "retrofit", promise);

promise.then(....);
```

With this it is possible to user promises without changing the Retrofit implementation, maybe this can help.

skarllot:
I think that there is not enought documentation to implement @SandroMachado solution using `CallAdapter`.

PR #831, opened 2015-05-03
Add rxpromises support

SandroMachado:
This pull request adds promises support to the retrofit.

Why promises support?  Mainly, because the power of chaining multiple of them together. With multiple promises in chaining we can have one single point to handle the errors and to execute the final operation. 
```java
promise.then(....).then(...).then(...).fail(....).fin(...)
```
The library of promises used was: https://github.com/darylteo/rxjava-promises

For the example project, the code adapted to the promises is:

```java
public class GitHubClient {

    private static final String API_URL = "https://api.github.com";

    static class Contributor {
        String login;
        int contributions;
    }

    interface GitHub {
        @GET("/repos/{owner}/{repo}/contributors")
        RetrofitPromise<List<Contributor>> contributors(
                @Path("owner") String owner,
                @Path("repo") String repo
        );
    }

    public static void main(String... args) {
        // Create a very simple REST adapter which points the GitHub API endpoint.
        RestAdapter restAdapter = new RestAdapter.Builder()
                .endpoint(API_URL)
                .build();

        // Create an instance of our GitHub API interface.
        GitHub github = restAdapter.create(GitHub.class);

        // Fetch and print a list of the contributors to this library.
        RetrofitPromise<List<Contributor>> promise = github.contributors("square", "retrofit");
        promise.then(new PromiseAction<List<Contributor>>() {
            @Override
            public void call(List<Contributor> contributors) {
                for (Contributor contributor : contributors) {
                    System.out.println(contributor.login + " (" + contributor.contributions + ")");
                }
            }
        }).fail(new PromiseAction<Exception>() {
            @Override
            public void call(Exception e) {

            }
        });
    }

}
```

### DIFF
--- a/retrofit/pom.xml
+++ b/retrofit/pom.xml
@@ -33,6 +33,11 @@
       <artifactId>rxjava</artifactId>
       <optional>true</optional>
     </dependency>
+    <dependency>
+  	  <groupId>com.darylteo</groupId>
+  	  <artifactId>rxjava-promises-java</artifactId>
+  	  <version>1.2.0</version>
+ 	</dependency>
 
     <dependency>
       <groupId>junit</groupId>
--- a/retrofit/src/main/java/retrofit/MethodInfo.java
+++ b/retrofit/src/main/java/retrofit/MethodInfo.java
@@ -16,6 +16,7 @@
 package retrofit;
 
 import com.squareup.okhttp.Response;
+
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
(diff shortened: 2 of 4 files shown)


JakeWharton:
> Why promises support? Mainly, because the power of chaining multiple of them together. With multiple promises in chaining we can have one single point to handle the errors and to execute the final operation.

We already support RxJava `Observable` for this style of chaining which not only allows you to achieve the same effect, but has composition operations which are much more powerful than those offered by that promises library. It also is much more widely used library to build on than this specific promises library–the latter of which actually is built on RxJava. I'm not sure why the library doesn't just accept an `Observable` if it's based on RxJava.

As a potential consolation, we're looking at a pluggable mechanism for people to provide their own wrapper around execution which would allow you add support for this promises library without modifying Retrofit itself. See #817.

PR #839, opened 2015-05-12
Remove our RequestInterceptor.

JakeWharton:
Use OkHttp's Interceptors for adding headers, query parameters, or modifying the path.

@swankjesse


kevelbreh:
Yes! This is beautiful. @JakeWharton when is the next release for retrofit happening?

JakeWharton:
There is no timeline yet.

PR #845, opened 2015-05-14
Introduce call, call adapters, and the response type.

### LINKED ISSUE #436: Add a way to get both the raw HTTP response + converted Object at once
I would like to get both the raw HTTP response (to grab response headers) + the converted object. I'm using RX Observables.

### LINKED ISSUE #444: Read Response Header data
Hi, Is it possible to read headers in the response direclty in my object? 
Like:
```Java
@POST("/users/login")
User login(String username, String password);
```
with the user having a field "token" or "cookies" in his fields?

or should I use the Response object to read the headers?
Like:
```Java
@POST("/users/login")
Response login(String username, String password);
```
Thanks folks.

### LINKED ISSUE #461: Pass through request data to Exception handling mechanism
I'm using your RxJava hook and I'd like to be able to use their error handling mechanism `Observable<T>.onErrorFlatMap()` to clean up bad requests and retry. To make this clean I would need to have Retrofit pass me the request data that caused the problem so I could wrap it in the Exception that I produce in my Retrofit `ErrorHandler`.

I really don't want to pass this stuff elsewhere (in my constructor(s)), since it mixes object lifecycles and will mess with dependency injection. Seems the cleanest way would be to cache the request data and wrap it in the Exception that Retrofit produces.

### LINKED ISSUE #647: Support HystrixRequestContext When Using Observable
Currently the RxSupport doesn't allow me to copy the HystrixRequestContext onto the Executor thread used by RetroFit. So now any commands executed in my Hystrix-enabled HTTP client fail. I cannot see a way to copy the context to the executor thread. This isn't an issue when using the none Observable version, where an Executor is not used.

Can there be support added to access the thread, in some way, to allow putting the HystrixRequestContext onto to that thread? Hystrix does this by using a CallableWrapper.

### LINKED ISSUE #669: RetrofitError cannot be serialized, extends RuntimeException
A RuntimeException is thrown if you attempt to serialize a RetrofitError that was created with a non-null Converter.

The converter field should be marked `transient`.

```
java.lang.RuntimeException: Parcelable encountered IOException writing serializable object 
...
Caused by: java.io.NotSerializableException: retrofit.converter.GsonConverter
       at java.io.ObjectOutputStream.writeNewObject(ObjectOutputStream.java:1364)
       at java.io.ObjectOutputStream.writeObjectInternal(ObjectOutputStream.java:1671)
       at java.io.ObjectOutputStream.writeObject(ObjectOutputStream.java:1517)
       at java.io.ObjectOutputStream.writeObject(ObjectOutputStream.java:1481)
       at java.io.ObjectOutputStream.writeFieldValues(ObjectOutputStream.java:979)
       at java.io.ObjectOutputStream.defaultWriteObject(ObjectOutputStream.java:368)
       at java.io.ObjectOutputStream.writeHierarchy(ObjectOutputStream.java:1074)
       at java.io.ObjectOutputStream.writeNewObject(ObjectOutputStream.java:1404)
       at java.io.ObjectOutputStream.writeObjectInternal(ObjectOutputStream.java:1671)
       at java.io.ObjectOutputStream.writeObject(ObjectOutputStream.java:1517)
       at java.io.ObjectOutputStream.writeObject(ObjectOutputStream.java:1481)
       at java.io.ObjectOutputStream.writeFieldValues(ObjectOutputStream.java:979)
       at java.io.ObjectOutputStream.defaultWriteObject(ObjectOutputStream.java:368)
       at java.io.ObjectOutputStream.writeHierarchy(ObjectOutputStream.java:1074)
       at java.io.ObjectOutputStream.writeNewObject(ObjectOutputStream.java:1404)
       at java.io.ObjectOutputStream.writeObjectInternal(ObjectOutputStream.java:1671)
       at java.io.ObjectOutputStream.writeObject(ObjectOutputStream.java:1517)
       at java.io.ObjectOutputStream.writeObject(ObjectOutputStream.java:1481)
       at java.io.ObjectOutputStream.writeFieldValues(ObjectOutputStream.java:979)
       at java.io.ObjectOutputStream.defaultWriteObject(ObjectOutputStream.java:368)
       at java.lang.Throwable.writeObject(Throwable.java:446)
       at java.lang.reflect.Method.invokeNative(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:511)
       at java.io.ObjectOutputStream.writeHierarchy(ObjectOutputStream.java:1053)
       at java.io.ObjectOutputStream.writeNewObject(ObjectOutputStream.java:1404)
       at java.io.ObjectOutputStream.writeObjectInternal(ObjectOutputStream.java:1671)
       at java.io.ObjectOutputStream.writeObject(ObjectOutputStream.java:1517)
       at java.io.ObjectOutputStream.writeObject(ObjectOutputStream.java:1481)
       at android.os.Parcel.writeSerializable(Parcel.java:1274)
       at android.os.Parcel.writeValue(Parcel.java:1233)
       at android.os.Parcel.writeMapInternal(Parcel.java:591)
       at android.os.Bundle.writeToParcel(Bundle.java:1627)
       at android.os.Parcel.writeBundle(Parcel.java:605)
       at android.content.Intent.writeToParcel(Intent.java:6751)
       at android.app.ActivityManagerProxy.finishActivity(ActivityManagerNative.java:2074)
       at android.app.Activity.finish(Activity.java:4230)
```

### LINKED ISSUE #710: Allow a service method to return ConnectableObservable<>
If a service method returns ConnectableObservable, then the server request would not be sent until the connect() method was called.

This allows you to detach the subscription from the sending of the request.

This would allow you to subscribe multiple observers and then connect().
IMO, it is more clear to have to call connect, than to remember that an HTTP call is a side effect of subscribe sometimes. 

The client code has to be aware of this special type of Observable with side effects, otherwise they might make more server requests than intended.

You can, of course, call .publish() on the Observables returned by the current version of Retrofit to turn them into ConnectableObservables, but you may forgot to do that in some places.

Calling subscribe to initiate a server call can obfuscate what is actually causing the server call for someone unfamiliar with Retrofit.

In fact, maybe for Retrofit 2 returning a plain Observable should be dis-allowed and ConnectableObservable could be the only option.

### LINKED ISSUE #782: Next iteration of first-party RxJava support.
We know we want a wrapper type. Similar to `Observable<Response>` but parameterized and having passed through the converter.

There's also the issue of errors which are a pain in the ass to deal with. We have the same `doOnError` / `onErrorResumeNext` nonsense all of the place because we don't want the stream to be torn down just because we flatMap'd through Retrofit.

### LINKED ISSUE #817: Pluggable execution mechanism

### LINKED ISSUE #835: No response (also no error) in chained/nested rxjava callbacks
This happens if you can together multiple Api calls using operators like switchMap, flatMap or concatMap.
I get no output from Retrofit and my code is stuck waiting for a response.
By no output I mean that none of the subscriber's methods get called(onNext, onError or onComplete)

Possible solutions for others who have the same issue:   

  - use the timeout operator from rjxava, or 
  - invoke the nested api calls from other threads than the ones that retrofit provides. You can do this by using the observeOn operator (any threadpool is ok, including using the UI thread)

Sample project that reproduces this issue and shows how to use a workaround to avoid the bug
https://github.com/adi1133/Retrofit-Issue

JakeWharton:
* `Call` is the mechanism for request execution. It encapsulates a single request action either synchronously or asynchronously, and can support canceling.
* `CallAdapter` is a means of supporting a method return type other than `Call` (like RxJava's `Observable`) by adapting the `Call` instance into another instance. This change moves RxJava support into a sibling module which uses an adapter for supporting `Observable` types.
* `Response` represents the HTTP response along side the deserialized body. It allows access to both the converted body and lower-level items such as the header, response code, and underlying OkHttp request and response objects.

@swankjesse

Closes #835.
Closes #817.
Closes #782.
Closes #710.
Closes #669.
Closes #647.
Closes #461.
Closes #444.
Closes #436.


retrofit-adapters/rxjava/src/main/java/retrofit/ObservableExecutionAdapter.java
```diff
@@ -0,0 +1,76 @@
...
+              case RESPONSE:
+                if (result.isError()) {
+                  subscriber.onError(result.error());
+                } else {
+                  subscriber.onNext(result.response());
+                  subscriber.onCompleted();
+                }
+                break;
+
+              case USER:
```
swankjesse: I'm intrigued

retrofit/src/main/java/retrofit/Call.java
```diff
@@ -0,0 +1,72 @@
...
+    this.execution = execution;
+    this.callbackExecutor = callbackExecutor;
+  }
+
+  public void enqueue(final Callback<T> callback) {
+    synchronized (this) {
+      if (executed) throw new IllegalStateException("Already Executed");
+      executed = true;
+    }
+    rawCall = execution.execute(new ExecutionCallback() {
```
swankjesse: whoa

swankjesse: I'm surprised it's called `execute` and takes a callback. In OkHttp `enqueue` takes a callback; `execute` is synchronous and returns a value.

retrofit/src/main/java/retrofit/Call.java
```diff
@@ -0,0 +1,72 @@
...
+            } else {
+              callback.success(result.response());
+            }
+          }
+        });
+      }
+    });
+  }
+
+  public Response<T> execute() throws IOException {
```
swankjesse: can you add an `executed` guard?

retrofit/src/main/java/retrofit/Call.java
```diff
@@ -0,0 +1,72 @@
...
+    final AtomicReference<Result<T>> resultRef = new AtomicReference<Result<T>>();
+    com.squareup.okhttp.Call call = execution.execute(new ExecutionCallback<T>() {
+      @Override public void result(Result<T> result) {
+        resultRef.set(result);
+        latch.countDown();
+      }
+    });
+
+    try {
+      latch.await();
```
swankjesse: eep. the dead waiting thread mechanism.

retrofit/src/main/java/retrofit/Call.java
```diff
@@ -0,0 +1,72 @@
...
+    } catch (InterruptedException e) {
+      call.cancel();
+      Utils.sneakyRethrow(e);
```
swankjesse: Teehee

retrofit/src/main/java/retrofit/Execution.java
```diff
@@ -0,0 +1,98 @@
+package retrofit;
+
+import com.squareup.okhttp.Call;
+import com.squareup.okhttp.Callback;
+import com.squareup.okhttp.Request;
+import com.squareup.okhttp.ResponseBody;
+import java.io.IOException;
+
+public final class Execution {
+  public enum Classification {
```
swankjesse: Not sure I like this name. Maybe it's packaging?

swankjesse: I mean that literally, like `enum Packaging { ... }`

retrofit/src/main/java/retrofit/Response.java
```diff
@@ -0,0 +1,63 @@
+package retrofit;
+
+import com.squareup.okhttp.Headers;
+import com.squareup.okhttp.ResponseBody;
+import java.io.IOException;
+import retrofit.converter.Converter;
+
+public final class Response<T> {
```
swankjesse: I like.

retrofit/src/main/java/retrofit/Response.java
```diff
@@ -0,0 +1,63 @@
...
+    return rawResponse.code();
+  }
+
+  public Headers headers() {
+    return rawResponse.headers();
+  }
+
+  /** {@code true} if {@link #code()} is in the range [200..300). */
+  public boolean isSuccess() {
+    return raw().isSuccessful();
```
swankjesse: inline `raw()` ?

retrofit/src/main/java/retrofit/Response.java
swankjesse: This needs a little doc to differentiate _Result_ from _Response_ which are typically synonyms

samples/github-client/src/main/java/com/example/retrofit/GitHubClient.java
```diff
@@ -45,8 +46,11 @@ public static void main(String... args) {
     // Create an instance of our GitHub API interface.
     GitHub github = restAdapter.create(GitHub.class);
 
-    // Fetch and print a list of the contributors to this library.
-    List<Contributor> contributors = github.contributors("square", "retrofit");
+    // Create a call instance for looking up Retrofit contributors.
+    Call<List<Contributor>> call = github.contributors("square", "retrofit");
+
+    // Fetch and print a list of the contributors to the library.
+    List<Contributor> contributors = call.execute().body();
```
swankjesse: Neat.

retrofit/src/main/java/retrofit/Response.java
JakeWharton: For sure. Everything here needs _a lot_ of documentation.

retrofit/src/main/java/retrofit/Types.java
```diff
@@ -32,8 +29,6 @@
  * @author Jesse Wilson
```
swankjesse: Delete this while you're at it? Deleting `@author` lines in general is bad (cause you're denying somebody credit for their work) but personally I'd prefer to not be named.

retrofit/src/main/java/retrofit/Call.java
JakeWharton: done

JakeWharton:
@swankjesse PTAL again. Another iteration. Still some stuff to talk about...

retrofit-adapters/rxjava/src/main/java/retrofit/ObservableExecutionAdapter.java
```diff
@@ -0,0 +1,97 @@
...
+   * TODO
+   */
+  private static ObservableExecutionAdapter create() {
+    return new ObservableExecutionAdapter();
+  }
+
+  private ObservableExecutionAdapter() {
+  }
+
+  @Override public Type parseType(Type returnType) {
```
swankjesse: parseType seems like the wrong name for this. Maybe it's something gross and long like `returnTypeToMessageType`?

swankjesse: (or just `messageType` ?)

retrofit/src/main/java/retrofit/Call.java
```diff
@@ -0,0 +1,173 @@
...
+          response = parseResponse(rawResponse);
+        } catch (final Throwable e) {
+          if (callbackExecutor != null) {
+            callbackExecutor.execute(new Runnable() {
+              @Override public void run() {
+                callFailure(e);
+              }
+            });
+          } else {
+            callFailure(e);
```
swankjesse: Maybe `callFailure()` should take care of bouncing to the right thread?

retrofit/src/main/java/retrofit/Call.java
```diff
@@ -0,0 +1,173 @@
...
+
+    try {
+      int code = rawResponse.code();
+      if (code < 200 || code >= 300) {
+        // Buffer the entire body in the event of a non-2xx status to avoid future I/O.
+        body = Utils.readBodyToBytesIfNecessary(rawBody);
+      } else if (code == 204 || code == 205) {
+        // HTTP 204 No Content "...response MUST NOT include a message-body"
+        // HTTP 205 Reset Content "...response MUST NOT include an entity"
+        if (rawBody.contentLength() > 0) {
```
swankjesse: Should OkHttp guarantee this? Tricky.

retrofit/src/main/java/retrofit/Call.java
```diff
@@ -0,0 +1,173 @@
...
+    return new Call(restAdapter, methodInfo, args, callbackExecutor);
+  }
+
+  public void enqueue(final Callback<T> callback) {
+    synchronized (this) {
+      if (executed) throw new IllegalStateException("Already executed");
+      executed = true;
+    }
+
+    final com.squareup.okhttp.Call call = createRawCall();
```
swankjesse: I see what you're doing here. I think it might actually read better if you deliberately shadow the `rawCall` field? Ie. the local variable would be `rawCall` . . . unsure. 

(I have a hard time reasoning about volatiles, though I think that's just me)

retrofit/src/main/java/retrofit/CallCallAdapter.java
```diff
@@ -0,0 +1,39 @@
...
+      throw new IllegalStateException(
+          "Call<T> cannot use Result or Response as its generic parameter. "
+              + "Specify the user type only (e.g., Call<TweetResponse>).");
+    }
+
+    return upperBound;
+  }
+
+  @Override public <T> Call<T> adapt(Call<T> call, Packaging packaging) {
+    return call.copyWithExecutor(callbackExecutor); // TODO just mutate and save an allocation?
```
swankjesse: I don't think your TODO plays. (The number of allocations per HTTP request is high.)

retrofit-adapters/rxjava/src/main/java/retrofit/ObservableExecutionAdapter.java
```diff
@@ -0,0 +1,97 @@
...
+      return null;
+    }
+    if (returnType instanceof ParameterizedType) {
+      return Types.getParameterUpperBound((ParameterizedType) returnType);
+    }
+    throw new IllegalStateException("Observable return type must be parameterized"
+        + " as Observable<Foo> or Observable<? extends Foo>");
+  }
+
+  @Override public <T> Observable<?> adapt(final Call<T> call, final Packaging packaging) {
```
swankjesse: Here's a silly idea . . . you could your anonymous observable could be available in `Response` packaging only, and then you could use Rx to transform that to other packagings as necessary.

swankjesse:
You've still got a lot of runway here, but LGTM as an early step in a new direction.

retrofit/src/main/java/retrofit/Call.java
JakeWharton: Then I have to try/catch/log twice. Crappy situation.

retrofit/src/main/java/retrofit/Call.java
JakeWharton: It's technically a protocol violation so I'm all for it!

retrofit-adapters/rxjava/src/main/java/retrofit/ObservableExecutionAdapter.java
JakeWharton: :heart_eyes:

JakeWharton:
Proceeding with test and `MockRestAdapter` updates. Documentation will be last.

retrofit-adapters/rxjava/src/main/java/retrofit/ObservableExecutionAdapter.java
```diff
@@ -0,0 +1,107 @@
...
+            })
+            .onErrorReturn(new Func1<Throwable, Result<T>>() {
+              @Override public Result<T> call(Throwable throwable) {
+                return Result.fromError(throwable);
+              }
+            });
+
+      case NONE:
+        return responseObservable //
+            .flatMap(new Func1<Response<T>, Observable<T>>() {
```
swankjesse: I still need to learn nuances between map and flatMap

retrofit-adapters/rxjava/src/main/java/retrofit/ObservableExecutionAdapter.java
```diff
@@ -0,0 +1,107 @@
...
+    private CallOnSubcribe(Call<T> originalCall) {
+      this.originalCall = originalCall;
+    }
+
+    @Override public void call(final Subscriber<? super Response<T>> subscriber) {
+      // Since Call is a one-shot type. Clone it for each new subscriber.
+      final Call<T> call = originalCall.clone();
+
+      // Attempt to cancel the call if it is still in-flight on unsubscription.
+      subscriber.add(Subscriptions.create(new Action0() {
```
swankjesse: RxJava rant: this line's hard to reason about. Not obvious that the Action0 is the unsubscribe callable.

swankjesse: ... I'd be tempted to skip the `Subscriptions` accelerator. Add one more method isUnsubscribed(); and save an allocation and an indirection.

retrofit-adapters/rxjava/src/main/java/retrofit/ObservableExecutionAdapter.java
```diff
@@ -0,0 +1,107 @@
...
+            subscriber.onNext(response);
+          } catch (Throwable t) {
+            subscriber.onError(t);
+            return;
+          }
+          subscriber.onCompleted();
+        }
+
+        @Override public void failure(Throwable t) {
+          subscriber.onError(t);
```
swankjesse: curious about alternatives. . . is this promoting Retrofit failure to Rx java error?

retrofit-adapters/rxjava/src/main/java/retrofit/ObservableExecutionAdapter.java
JakeWharton: [`map`][1] converts each `Foo` to a `Bar`. [`flatMap`][2] converts each `Foo` to an `Observable<Bar>` and "flattens" each back into a single stream.

Normally in this case I would use `map`, but I have no way of throwing the checked `IOException` for it to show up as an `onError`. I leverage `flatMap` to either return a stream that has a single "`Bar`" value which will be flattened and thus behave like a `map` OR I return a stream that immediately calls `onError`.

 [1]: http://reactivex.io/documentation/operators/map.html
 [2]: http://reactivex.io/documentation/operators/flatmap.html

retrofit-adapters/rxjava/src/main/java/retrofit/ObservableExecutionAdapter.java
JakeWharton: Totally agree. The parameter name is `unsubscribeAction` but it's such a terrible method name that it fools you every time.

https://github.com/ReactiveX/RxJava/issues/2967

retrofit-adapters/rxjava/src/main/java/retrofit/ObservableExecutionAdapter.java
JakeWharton: Yes. If you request an `Observable<Result<T>>` we coerce it back into data with `onErrorReturn` above though.

JakeWharton:
PTAL! I'm (fairly) confident this is in a state for merge. There's a ton of TODO scattered around for documentation and the mock module is disabled to be updated separately, but it's working!

retrofit-adapters/rxjava/src/main/java/retrofit/ObservableCallAdapterFactory.java
```diff
@@ -0,0 +1,167 @@
+package retrofit;
```
swankjesse: nit: copyright

retrofit-adapters/rxjava/src/main/java/retrofit/ObservableCallAdapterFactory.java
```diff
@@ -0,0 +1,167 @@
...
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import rx.Observable;
+import rx.Subscriber;
+import rx.functions.Action0;
+import rx.functions.Func1;
+import rx.subscriptions.Subscriptions;
+
+/**
+ * TODO docs
```
swankjesse: Yeah, definitely need this. I think the thing that's important to mention is that each subscription yields a single value, computed fresh for each subscriber.

retrofit-adapters/rxjava/src/main/java/retrofit/Result.java
```diff
@@ -0,0 +1,64 @@
...
+   * The response received from executing an HTTP request. Only present when {@link #isError()} is
+   * {@code false}, {@code null} otherwise.
+   */
+  public Response<T> response() {
+    return response;
+  }
+
+  /**
+   * The error experienced while attempting to execute an HTTP request. Only present when {@link
+   * #isError()} is {@code true}, {@code null} otherwise.
```
swankjesse: The `{@code ...}` thing isn't necessary for true or null; those are plain English words that you can use in regular text.

swankjesse: (You wouldn't say, _“only present if the size is greater than {@code 5}”_)

retrofit-adapters/rxjava/src/test/java/retrofit/ObservableCallAdapterFactoryTest.java
```diff
@@ -0,0 +1,183 @@
+package retrofit;
```
swankjesse: copyright

retrofit-adapters/rxjava/src/test/java/retrofit/ObservableCallAdapterFactoryTest.java
```diff
@@ -0,0 +1,183 @@
...
+    String body = observable.toBlocking().first();
+    assertThat(body).isEqualTo("Hi");
+  }
+
+  @Test public void bodySuccess404() {
+    Type type = new TypeToken<Observable<String>>() {}.getType();
+    CallAdapter adapter = factory.get(type);
+    Observable<String> observable = (Observable<String>) adapter.adapt(new EmptyCall() {
+      @Override public void enqueue(Callback<Object> callback) {
+        callback.success(Response.fromError(404, ResponseBody.create(MediaType.parse("application/json"), "Hi"), new GsonConverter()));
```
swankjesse: Strange to have a non-JSON body here. Shouldn't it be text/plain?

(I'm assuming the converter is only exercised on 2xx responses)

retrofit-adapters/rxjava/src/test/java/retrofit/ObservableCallAdapterFactoryTest.java
```diff
@@ -0,0 +1,183 @@
...
+      @Override public void enqueue(Callback<Object> callback) {
+        callback.failure(throwable);
+      }
+    });
+    Result<String> result = observable.toBlocking().first();
+    assertThat(result.isError()).isTrue();
+    assertThat(result.error()).isSameAs(throwable);
+  }
+
+  @Test public void responseType() {
```
swankjesse: nice test

swankjesse: no `? extends ...` ?

retrofit-adapters/rxjava/src/test/java/retrofit/ResultTest.java
```diff
@@ -0,0 +1,43 @@
+package retrofit;
```
swankjesse: copyright

retrofit/src/main/java/retrofit/Call.java
```diff
@@ -0,0 +1,10 @@
+package retrofit;
```
swankjesse: copyright

retrofit/src/main/java/retrofit/Call.java
```diff
@@ -0,0 +1,10 @@
...
+
+import java.io.IOException;
+
+public interface Call<T> extends Cloneable {
```
swankjesse: off topic: and now I'm wondering whether we should have used an interface in OkHttp

retrofit/src/main/java/retrofit/CallAdapter.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package retrofit;
+
+import java.lang.reflect.Type;
+
+public interface CallAdapter<T> {
```
swankjesse: This is wonderful

retrofit/src/main/java/retrofit/DefaultCallAdapterFactory.java
```diff
@@ -0,0 +1,109 @@
...
+ */
+package retrofit;
+
+import java.io.IOException;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.concurrent.Executor;
+
+final class DefaultCallAdapterFactory implements CallAdapter.Factory {
+  private final Executor callbackExecutor;
```
swankjesse: optional?

retrofit/src/main/java/retrofit/DefaultCallAdapterFactory.java
```diff
@@ -0,0 +1,109 @@
...
+    private final Executor callbackExecutor;
+    private final Call<T> delegate;
+
+    ExecutorCallbackCall(Executor callbackExecutor, Call<T> delegate) {
+      this.callbackExecutor = callbackExecutor;
+      this.delegate = delegate;
+    }
+
+    @Override public void enqueue(Callback<T> callback) {
+      delegate.enqueue(new ExecutorCallback<>(callbackExecutor, callback));
```
swankjesse: I don't quite follow this. Why isn't the callback thread itself appropriate?

retrofit/src/main/java/retrofit/DefaultCallAdapterFactory.java
```diff
@@ -0,0 +1,109 @@
...
+    @Override public Response<T> execute() throws IOException {
+      return delegate.execute();
+    }
+
+    @Override public void cancel() {
+      delegate.cancel();
+    }
+
+    @Override public Call<T> clone() {
+      return new ExecutorCallbackCall<>(callbackExecutor, delegate.clone());
```
swankjesse: not cloning the executor is tricky. What if it's a single-threaded executor?

retrofit/src/main/java/retrofit/MethodInfo.java
```diff
@@ -223,69 +210,35 @@ private void parseHttpMethodAndPath(String method, String path, boolean hasBody)
...
 
-    lastArgType = Types.getSupertype(lastArgType, Types.getRawType(lastArgType), Callback.class);
-    if (lastArgType instanceof ParameterizedType) {
-      responseObjectType = getParameterUpperBound((ParameterizedType) lastArgType);
-      return ExecutionType.ASYNC;
+    StringBuilder builder = new StringBuilder() //
+        .append("No registered call adapters were able to handle return type ")
+        .append(returnType)
+        .append(". Checked: [ ");
+    for (int i = 0, count = factories.size(); i < count; i++) {
```
swankjesse: You might be better off just using list's toString. If for whatever reason I've got my own configurable factory class, this prevents me from adding my own toString debugging. (Especially since you could implement toString() to return the factory class simplename in the built-in implementations.)

retrofit/src/main/java/retrofit/NetworkCall.java
```diff
@@ -0,0 +1,160 @@
...
+ */
+package retrofit;
+
+import com.squareup.okhttp.OkHttpClient;
+import com.squareup.okhttp.Request;
+import com.squareup.okhttp.ResponseBody;
+import java.io.IOException;
+import retrofit.converter.Converter;
+
+final class NetworkCall<T> implements Call<T> {
```
swankjesse: Not sure I love _network_ here. I get it, but it could be confusing due to the fact that the call may ultimately never see the network. (caching, etc.)

retrofit/src/main/java/retrofit/NetworkCall.java
```diff
@@ -0,0 +1,160 @@
...
+
+      @Override public void onResponse(com.squareup.okhttp.Response rawResponse) {
+        final Response<T> response;
+        try {
+          response = parseResponse(rawResponse);
+        } catch (final Throwable e) {
+          callFailure(e);
+          return;
+        }
+        callSuccess(response);
```
swankjesse: I would have expected no need to trampoline off of this thread in the executor delegate.

retrofit/src/main/java/retrofit/NetworkCall.java
```diff
@@ -0,0 +1,160 @@
...
+import com.squareup.okhttp.ResponseBody;
+import java.io.IOException;
+import retrofit.converter.Converter;
+
+final class NetworkCall<T> implements Call<T> {
+  private final Endpoint endpoint;
+  private final Converter converter;
+  private final OkHttpClient client;
+  private final MethodInfo methodInfo;
+  private final Object[] args;
```
swankjesse: slightly annoyingly, these can be mutable, which makes executing `call()` multiple times dangerous. (Such as if this is an InputStream)

retrofit/src/main/java/retrofit/Utils.java
```diff
@@ -77,6 +67,51 @@ static Response readBodyToBytesIfNecessary(Response response) throws IOException
     }
   }
 
+  public static Type getParameterUpperBound(ParameterizedType type) {
+    Type[] types = type.getActualTypeArguments();
+    for (int i = 0; i < types.length; i++) {
```
swankjesse: Why loop? Only i == 0 matters, right?

retrofit/src/test/java/retrofit/CallTest.java
```diff
@@ -0,0 +1,320 @@
+package retrofit;
```
swankjesse: copyright

retrofit/src/test/java/retrofit/CallTest.java
```diff
@@ -0,0 +1,320 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+
+public final class CallTest {
+  @Rule public final MockWebServerRule server = new MockWebServerRule();
+
+  interface Service {
```
swankjesse: love this

retrofit/src/test/java/retrofit/CallTest.java
```diff
@@ -0,0 +1,320 @@
...
+        latch.countDown();
+      }
+    });
+    assertTrue(latch.await(2, SECONDS));
+
+    Throwable failure = failureRef.get();
+    assertThat(failure).isInstanceOf(IOException.class);
+  }
+
+  @Test public void conversionProblemOutgoingSync() throws IOException {
```
swankjesse: Love these tests

retrofit/src/test/java/retrofit/CallTest.java
```diff
@@ -0,0 +1,320 @@
...
+        latch.countDown();
+      }
+    });
+    assertTrue(latch.await(2, SECONDS));
+
+    assertThat(failureRef.get()).isInstanceOf(UnsupportedOperationException.class)
+        .hasMessage("I am broken!");
+  }
+
+  @Test public void http204SkipsConverter() throws IOException {
```
swankjesse: interesting!

retrofit/src/test/java/retrofit/DefaultCallAdapterFactoryTest.java
```diff
@@ -0,0 +1,135 @@
+package retrofit;
```
swankjesse: copyright

retrofit/src/test/java/retrofit/DefaultCallAdapterFactoryTest.java
```diff
@@ -0,0 +1,135 @@
...
+    } catch (IllegalArgumentException e) {
+      assertThat(e).hasMessage("Call<T> cannot use Response as its generic parameter. "
+          + "Specify the response body type only (e.g., Call<TweetResponse>).");
+    }
+  }
+
+  @Test public void responseType() {
+    Type classType = new TypeToken<Call<String>>() {}.getType();
+    assertThat(factory.get(classType).responseType()).isEqualTo(String.class);
+    Type wilcardType = new TypeToken<Call<? extends String>>() {}.getType();
```
swankjesse: dig

samples/src/main/java/com/example/retrofit/CustomCallAdapter.java
```diff
@@ -0,0 +1,100 @@
+package com.example.retrofit;
```
swankjesse: copyright

samples/src/main/java/com/example/retrofit/CustomCallAdapter.java
```diff
@@ -0,0 +1,100 @@
...
+  interface HttpBinService {
+    @GET("/ip")
+    ListenableFuture<Ip> getIp();
+  }
+
+  static class Ip {
+    String origin;
+  }
+
+  public static void main(String... args) {
```
swankjesse: fun!

swankjesse:
# LGTM

Shipit

retrofit-adapters/rxjava/src/main/java/retrofit/Result.java
JakeWharton: I'm skeptical of the reasoning only because these two are keywords whereas a number literal is not. But I get your argument and will change. We should really write a style guide proper...

retrofit-adapters/rxjava/src/test/java/retrofit/ObservableCallAdapterFactoryTest.java
JakeWharton: Extends case is second.

retrofit/src/main/java/retrofit/DefaultCallAdapterFactory.java
JakeWharton: Required. The `NetworkCall` calls callbacks on the OkHttp callback thread. This bounces it onto the appropriate thread for the platform (e.g., main thread on Android).

retrofit/src/main/java/retrofit/DefaultCallAdapterFactory.java
JakeWharton: The executor is global to all requests. The Android main thread can be thought of as a single-threaded executor.

retrofit/src/main/java/retrofit/DefaultCallAdapterFactory.java
JakeWharton: Retrofit has always delivered callbacks to the main thread. We do all the work on OkHttp's callback thread and then bounce to the main thread to deliver the final, immutable result.

retrofit/src/main/java/retrofit/ExceptionCatchingRequestBody.java
```diff
@@ -49,6 +49,10 @@
     });
   }
 
+  @Override public void close() throws IOException {
+    delegate.close();
```
JakeWharton: I think this might have been the StrictMode leak from the disk cache we've seen...

retrofit/src/main/java/retrofit/MethodInfo.java
JakeWharton: Good call, done.

retrofit/src/main/java/retrofit/NetworkCall.java
JakeWharton: I tried a bunch of other names and am open to suggestions. OkHttpCall, RetrofitCall, RealCall...

retrofit-adapters/rxjava/src/main/java/retrofit/Result.java
swankjesse: Hey @kevinb9n, any chance you think this is something you could slip into the Google Java Style guide?

retrofit/src/main/java/retrofit/DefaultCallAdapterFactory.java
swankjesse: Makes sense.

retrofit/src/main/java/retrofit/NetworkCall.java
swankjesse: I think OkHttpCall is good. It's the underlying layer.

charbgr:
Can ```Call``` or ```CallAdapter``` be used with ```Espresso 2.0``` ?

JakeWharton:
Yes. I don't see how the two are at all related.

charbgr:
I mean with [IdlingResource](http://developer.android.com/reference/android/support/test/espresso/IdlingResource.html).

JakeWharton:
You'll have to attach an idling resource to OkHttp's executor inside the
dispatcher as well as pass Retrofit your own callback executor which is
attached to an idling resource.

We'll put an example in https://github.com/JakeWharton/u2020 once 2.0 ships.

stevecstian:
@JakeWharton How about using `rx.Observable#doOnUnsubscribe` ?

JakeWharton:
For the idling resource? I think you'd want `doOnTerminate` instead.

stevecstian:
To replace below?
```
      // Attempt to cancel the call if it is still in-flight on unsubscription.
      subscriber.add(Subscriptions.create(new Action0() {
        @Override public void call() {
          call.cancel();
        }
      }));
```

JakeWharton:
the implementation of `doOnSubscribe` is exactly what you just posted

stevecstian:
Yeah, I know.  Just for semantic clarity?

JakeWharton:
I don't want to add another operator allocation on top of the already-heavy requirements of RxJava. If the subscription factory had a better name it would be a lot more clear: https://github.com/ReactiveX/RxJava/issues/2967

stevecstian:
Good point! :metal:

adriancole:
You should be very proud of this work, Jake. The call adapter registry is
simple and effective. You've managed to elegantly decouple rx from core
without watering it down or compromising. Really. Good job.

PR #848, opened 2015-05-17
Switch to OkHttp's built-in form body builder.

JakeWharton:


retrofit/src/test/java/retrofit/RequestBuilderTest.java
```diff
@@ -1008,22 +1009,22 @@ Response method(
...
-    assertThat(request.urlString()).isEqualTo("http://example.com/foo/bar/?k+it=k%20t&pi+ng=p%20g");
+    assertThat(request.urlString()).isEqualTo("http://example.com/foo/bar/?k%20it=k%20t&pi%20ng=p%20g");
     assertThat(request.body()).isNull();
   }
 
   @Test public void normalPostWithPathParam() {
     class Example {
       @POST("/foo/bar/{ping}/") //
-      Response method(@Path("ping") String ping) {
+      Response method(@Path("ping") String ping, @Body Object body) {
```
JakeWharton: I have no idea how this passed before...

JakeWharton:
Ping @swankjesse

retrofit/src/main/java/retrofit/http/Field.java
```diff
@@ -57,9 +57,6 @@
 public @interface Field {
   String value();
 
-  /** Specifies whether {@link #value()} is URL encoded. */
-  boolean encodeName() default true;
-
-  /** Specifies whether the argument value to the annotated method parameter is URL encoded. */
-  boolean encodeValue() default true;
+  /** Specifies whether the {@linkplain #value() name} and value are URL encoded. */
```
swankjesse: are _already_ URL encoded ?

swankjesse: too bad about `{@linkplain #value() name}`. That's how it goes though.

retrofit/src/main/java/retrofit/http/Field.java
```diff
@@ -57,9 +57,6 @@
...
+  boolean encode() default true;
```
swankjesse: _encode_ means "the value needs Retrofit to encode it" or "the value doesn't need Retrofit to encode it" ?

swankjesse:
Deleting code is good.

retrofit/src/main/java/retrofit/http/Field.java
JakeWharton: Might be best to stick to the "encoded" terminology here. Both for consistency and it seems more clear.

swankjesse: yes, I like that better.

retrofit/src/test/java/retrofit/RequestBuilderTest.java
roman-mazur: okhttp 2.3 allows to pass `null` when `RequestBody` is constructed, which is done by retrofit 1.9.
With okhttp 2.4 an empty body must be defined explicitly (e.g. with an empty array).
This change removes a test for post requests with empty body. Will they be supported?

swankjesse: Callers that want to post an empty body will need to pass an empty string.

roman-mazur: So is it a deliberate decision and you want to prevent declaration of `@POST` methods without `@Body`? Or would you accept a change that revives such declarations?

JakeWharton: Retrofit can coerce the absence of body into the empty body.

PR #849, opened 2015-05-18
Normalize specifying encoded values via parameter annotations.

JakeWharton:


JakeWharton:
Ping @swankjesse. Follow-up from #848.

PR #856, opened 2015-05-21
Fix memory leak caused by MockRxSupport

grandstaish:
subscribeOn(httpScheduler) is causing occasional memory leak to show up in LeakCanary. Is there a reason for having this here? Everything seems to work fine without it and it fixes the memory leak issues I am running into.

### DIFF
--- a/retrofit-mock/src/main/java/retrofit/MockRestAdapter.java
+++ b/retrofit-mock/src/main/java/retrofit/MockRestAdapter.java
@@ -494,15 +494,17 @@ public final class MockRestAdapter {
       });
     }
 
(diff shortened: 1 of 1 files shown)


grandstaish:
Closed. This was actually due to the default http scheduler in v1.9 (which has been completely removed in 2.0):

```
@Override Executor defaultHttpExecutor() {
      return Executors.newCachedThreadPool(new ThreadFactory() {
        @Override public Thread newThread(final Runnable r) {
          return new Thread(new Runnable() {
            @Override public void run() {
              Process.setThreadPriority(THREAD_PRIORITY_BACKGROUND);
              r.run();
            }
          }, RestAdapter.IDLE_THREAD_NAME);
        }
      });
    }
```

Issue #528, opened 2014-06-10
Mock Client implementation


ccmtaylor:
Hi,

inspired by the answers to [a question on StackOverflow](http://stackoverflow.com/questions/17544751/square-retrofit-server-mock-for-testing), I wrote [retromock](https://github.com/ccmtaylor/retromock/), a mock implementation of `retrofit.client.Client` and a couple of associated hamcrest matchers that let you specify a mock API endpoint for testing with a fluent API reminiscent of [wiremock](http://wiremock.org/stubbing.html) or [rest-driver](https://github.com/rest-driver/rest-driver/wiki), but without forcing you to start up a web server for your tests.

As of now, I'm still working on adding functionality, tests and documentation, but you can find a sketch of how to use the library in [this test](https://github.com/ccmtaylor/retromock/blob/master/src/test/java/retromock/MockClientTest.java).

Would you be interested in including this in retrofit itself? I couldn't find a mailing list or forum for retrofit-related discussion, so if I should submit this somewhere else, please let me know.

JakeWharton:
`Client` has died. OkHttp is now used directly for v2. The right thing to do is either mock the interface, script OkHttp via an interceptor, or use MockWebServer.

PR #861, opened 2015-05-26
Add a default mock service implementation

mullender:
This request is similar to #528, but only requires pure retrofit.

By providing a default mock service implementation, the need for manually implementing the retrofit service api is eliminated. Making offline testing a lot easier.
The mock service implementation takes a MocksService.Dispatcher to allow dispatching of custom responses.

unit tests that illustrate the usage are provided

### DIFF
--- a/retrofit-mock/src/main/java/retrofit/MockService.java
+++ b/retrofit-mock/src/main/java/retrofit/MockService.java
@@ -1,0 +1,80 @@
+package retrofit;
+
+import rx.Observable;
+
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
+import java.util.HashMap;
+import java.util.Map;
+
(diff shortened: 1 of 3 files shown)


JakeWharton:
> manually implementing the retrofit service api is eliminated

We emphasize this rather than shy away from it. One of the points of having our endpoint declarations be an interface is because you can do this. That said, I don't think we want to conflate what the mock module does with this endpoint-based dispatcher implementation.

If you want to do something like this, there's a few approaches which I think are better:

 1. Just implement your own `InvocationHandler` via `Proxy`. You get basically the same callback with the method to read information to switch on.
 2. Use an OkHttp interceptor and never call down the chain. Simply dispatch the request somewhere and return the fake response. This has the advantage of access to the full URL and headers but you also have to deal in raw bodies. You could also run MockWebServer locally and connect to it instead of the real server and use its dispatcher for endpoint-based handling.

PR #868, opened 2015-05-28
Remove okhttp-urlconnection Gradle dependency

commonquail:
The previous commit removed the dependency from the Maven declaration but left it in the Gradle declaration.


JakeWharton:
The website is maintained on the `master` branch alongside the code. Please make the PR there.

PR #874, opened 2015-06-01
Resolve mixed-content resources

commonquail:
jQuery is served over HTTP, which causes issues with [blocked mixed content](https://developer.mozilla.org/en-US/docs/Security/MixedContent), in particular preventing loading the latest version number. This PR ensures external content is served over HTTPS.

Three HTTP anchors to GitHub-owned domains are left unchanged. Let me know if you want those updated as well.


JakeWharton:
Thanks!

PR #875, opened 2015-06-01
Coerce the absence of body into the empty body

roman-mazur:
Related discussion https://github.com/square/retrofit/pull/848#discussion_r30467570
This raises a couple of questions though. See inline comments.

### DIFF
--- a/retrofit/src/main/java/retrofit/RequestBuilder.java
+++ b/retrofit/src/main/java/retrofit/RequestBuilder.java
@@ -41,10 +41,12 @@ import retrofit.http.QueryMap;
 
 final class RequestBuilder {
   private static final Headers NO_HEADERS = Headers.of();
+  private static final byte[] NO_BODY = new byte[0];
 
   private final Converter converter;
   private final Annotation[] paramAnnotations;
   private final String requestMethod;
+  private final boolean requestHasBody;
   private final String apiUrl;
 
   private MultipartBuilder multipartBuilder;
@@ -62,6 +64,7 @@ final class RequestBuilder {
 
     paramAnnotations = methodInfo.requestParamAnnotations;
     requestMethod = methodInfo.requestMethod;
+    requestHasBody = methodInfo.requestHasBody;
 
     if (methodInfo.headers != null) {
       headers = methodInfo.headers.newBuilder();
@@ -366,6 +369,9 @@ final class RequestBuilder {
         body = formEncodingBuilder.build();
       } else if (multipartBuilder != null) {
         body = multipartBuilder.build();
+      } else if (requestHasBody) {
+        // Body is absent, make an empty body.
+        body = RequestBody.create(null, NO_BODY);
       }
     }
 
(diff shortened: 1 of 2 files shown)


retrofit/src/main/java/retrofit/Utils.java
```diff
@@ -39,6 +39,12 @@
     return object;
   }
 
+  static boolean methodRequiresRequestBody(final String method) {
```
roman-mazur: Should/can we use internal package of okhttp?

retrofit/src/main/java/retrofit/RequestBuilder.java
```diff
@@ -366,6 +367,9 @@ Request build() {
         body = formEncodingBuilder.build();
       } else if (multipartBuilder != null) {
         body = multipartBuilder.build();
+      } else if (Utils.methodRequiresRequestBody(requestMethod)) {
+        // Body is absent, make an empty body.
+        body = RequestBody.create(MediaType.parse("text/plain"), NO_BODY);
```
roman-mazur: MIME type is a question. Converter know about it. But it's not capable of producing an empty body. If we delefate this to a Converter, it will have to provide something like `RequestBody emptyBody()`.
On the other hand, if API accepts an empty body, content type should really matter.

retrofit/src/main/java/retrofit/Utils.java
JakeWharton: No. Also this doesn't handle arbitrary HTTP methods defined by the user with `hasBody = true`.

roman-mazur: The problem is that okhttp will not allow you to make a custom request with body...
You'll get
```
java.lang.IllegalArgumentException: method CUSTOM must not have a request body.
	at com.squareup.okhttp.Request$Builder.method(Request.java:241)
```
because its `permitsRequestBody` will return false.

JakeWharton: That's not Retrofit's problem. And there's an issue tracking that behavior (https://github.com/square/okhttp/issues/229).

roman-mazur:
Pushed an update to illustrate the problem with custom methods.

retrofit/src/test/java/retrofit/RequestBuilderTest.java
```diff
@@ -860,6 +860,34 @@
...
+  }
+
+  @Test public void customMethodemptyBody() {
+    class Example {
+      @HTTP(method = "CUSTOM", path = "/foo/bar/", hasBody = true) //
+      Call<Object> method() {
+        return null;
+      }
+    }
+    Request request = buildRequest(Example.class);
```
roman-mazur: This fails because okhttp does not permit custom methods to have a body.

roman-mazur:
Let me know what you think about MIME type.

retrofit/src/test/java/retrofit/RequestBuilderTest.java
JakeWharton: Suppress like the other one in this file.

roman-mazur: Done.

JakeWharton:
I don't think we can force a mime type onto the request. That's asking for trouble. Servers route requests to different handlers based on things like this. Why not pass `null` instead?

roman-mazur:
Now we pass `null`. Just haven't though about `null` content type as an option for some reason.

JakeWharton:
Thanks!

PR #876, opened 2015-06-02
Move converter into the main package, remove Gson from core.

JakeWharton:
This change means that only RequestBody and ResponseBody are the types which are allowed for their respective objects by default. If you want serialization of any kind, you need to opt-in with either your own converter or one of the provided ones.


JakeWharton:
Shameless @swankjesse review mention because I know he's online...

retrofit/src/main/java/retrofit/DefaultConverter.java
```diff
@@ -0,0 +1,18 @@
+package retrofit;
```
swankjesse: nit: copyright

retrofit/src/main/java/retrofit/DefaultConverter.java
```diff
@@ -0,0 +1,18 @@
...
+
+import com.squareup.okhttp.RequestBody;
+import com.squareup.okhttp.ResponseBody;
+import java.io.IOException;
+import java.lang.reflect.Type;
+
+final class DefaultConverter implements Converter {
```
swankjesse: Hmm . . . wouldn't it be better to blow up at rest adapter creation time?

retrofit/src/test/java/retrofit/RestAdapterTest.java
```diff
@@ -129,4 +143,62 @@
...
+    RestAdapter ra = new RestAdapter.Builder()
+        .endpoint(server.getUrl("/").toString())
+        .build();
+    CallMethod example = ra.create(CallMethod.class);
+    Call<String> call = example.disallowed("Hi!");
+    try {
+      call.execute();
+      fail();
+    } catch (UnsupportedOperationException e) {
+      assertThat(e).hasMessage("No converter installed to handle serializing class java.lang.String to body");
```
swankjesse: I think the words here are flipped? 

_No converter installed to handle serializing class java.lang.String to body_
vs.
_No converter installed to handle serializing body to class java.lang.String_

swankjesse: . . . ahh, I get it. It's `body` that's distracting me. Nevermind.

swankjesse: Maybe drop `to body` here and below? That's implicit

swankjesse:
LGTM

retrofit/src/main/java/retrofit/DefaultConverter.java
JakeWharton: Are you saying a converter is then required? As it stands now it'd be perfectly fine to use Retrofit with `RequestBody` and `ResponseBody` as your only non-primitive, non-String types.

retrofit/src/main/java/retrofit/DefaultConverter.java
JakeWharton: done

retrofit/src/main/java/retrofit/DefaultConverter.java
swankjesse: Works for me.

swankjesse: ... though it still seems like that check should happen earlier. When I build a rest adapter that's when it should find the converter if it needs one, and if that's missing it should blow up.

JakeWharton: True. We can do that.

retrofit/src/test/java/retrofit/RestAdapterTest.java
```diff
@@ -129,4 +144,60 @@
...
     }
   }
+
+  @Test public void missingConverterThrowsOnNonRequestBody() throws IOException {
+    RestAdapter ra = new RestAdapter.Builder()
+        .endpoint(server.getUrl("/").toString())
+        .build();
+    CallMethod example = ra.create(CallMethod.class);
+    try {
+      example.disallowed("Hi!");
```
JakeWharton: This test and the one below now fail on service method invocation rather than call execution with the converter validation changes :+1:

retrofit/src/test/java/retrofit/StringConverter.java
```diff
@@ -0,0 +1,17 @@
+package retrofit;
```
swankjesse: nit: copyright

retrofit/src/test/java/retrofit/RestAdapterTest.java
```diff
@@ -1,19 +1,33 @@
 // Copyright 2013 Square, Inc.
```
swankjesse: nit: Apache copyright

PR #880, opened 2015-06-03
Remove the mock preference value change listener.

JakeWharton:
Strange concept. Not sure what I was thinking here. I was being lazy on figuring out a way to persist values but that's your job, not the MockRestAdapter's.


JakeWharton:
@swankjesse

PR #881, opened 2015-06-03
Correct Simple XML converter's type casing.

JakeWharton:
@swankjesse


swankjesse:
This one is a tricky one for git!

PR #882, opened 2015-06-03
Rewrite RxJava call adapter tests to be more understandable.

JakeWharton:
@swankjesse


adriancole:
LG - agree tests now focus more on what's being asserted vs dealing with
the incidental.

swankjesse:
LGTM

PR #885, opened 2015-06-04
Forbid type variable use in response types.

### LINKED ISSUE #433: Generic API interfaces
Hello guys,

First of all, thank you for this amazing project. It saves a lot of coding for me!

Now, I came across a small problem. 

My API is simple and virtually all trafficked objects are similar. Therefore all have the same parent class. However, I can not create an API interface based on that parent class, or using generics. For example:

```java
public class BaseModel {
    // ...
}

public class TodoItem extends BaseModel {
    // ...
}

// ------------

public interface SyncApi {
    @GET("/some/url/with/{id}")
    TodoItem get(@Path("id") long id);  // This works well :)

    @GET("/some/url/with/{id}")
    BaseModel get(@Path("id") long id);  // This fail because GsonConverter does not have an known de/serializer for BaseModel

    @GET("/some/url/with/{id}")
    <T extends BaseModel> T get(@Path("id") long id); // This return a LinkedTreeMap from Gson
}

// ---------

public interface SyncApi<T extends BaseModel> {
    @GET("/some/url/with/{id}")
    T get(@Path("id") long id);
}

public interface TodoItemApi extends SyncApi<TodoItem> {
    // This raises exception from Retrofit validation about heritance
}
```

Have you some suggestion?

### LINKED ISSUE #807: Can't get the genericity?
#my service
```
public interface BaseHttpService {
    @POST("/{requestUrl}")
    <T> void post(@Path("requestUrl") String requestUrl, @QueryMap Map<String, Object> map, Callback<T> callback);
}
```

#my request
```
 BaseHttpService service = restAdapter.create(BaseHttpService.class);
 Map<String, Object> params = new HashMap<>();
 addParams(0, params);
 service.post("xxx", params, callback);
```

#my callback
```
 private Callback<DishListResponse<DishDetail>> callback = new Callback<DishListResponse<DishDetail>>() {
        @Override
        public void success(DishListResponse<DishDetail> dishListResponse, Response response) {
        }

        @Override
        public void failure(RetrofitError retrofitError) {
        }
    };
```

#the exception
```
    java.lang.ClassCastException: java.util.LinkedHashMap cannot be cast to xxx.DishListResponse
            at xxx$2.success(DishKitchenFragment.java:116)
            at retrofit.CallbackRunnable$1.run(CallbackRunnable.java:45)
            at android.os.Handler.handleCallback(Handler.java:615)
            at android.os.Handler.dispatchMessage(Handler.java:92)
            at android.os.Looper.loop(Looper.java:137)
            at android.app.ActivityThread.main(ActivityThread.java:4898)
            at java.lang.reflect.Method.invokeNative(Native Method)
            at java.lang.reflect.Method.invoke(Method.java:511)
            at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1008)
            at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:775)
            at dalvik.system.NativeStart.main(Native Method)
```

#but change callback like this , it's ok.
```
@POST("/xxx")
void listDish(@QueryMap Map<String, Object> map, Callback<DishListResponse<DishDetail>> callback);
```

JakeWharton:
@swankjesse 

Closes #807
Closes #433

### DIFF
--- a/retrofit/src/main/java/retrofit/Utils.java
+++ b/retrofit/src/main/java/retrofit/Utils.java
@@ -81,6 +81,33 @@ final class Utils {
     return paramType;
   }
 
+  public static boolean hasTypeVariable(Type type) {
+    if (type instanceof Class<?>) {
+      return false;
+    }
+    if (type instanceof ParameterizedType) {
+      ParameterizedType parameterizedType = (ParameterizedType) type;
+      for (Type typeArgument : parameterizedType.getActualTypeArguments()) {
+        if (hasTypeVariable(typeArgument)) {
+          return true;
+        }
+      }
+      return false;
+    }
+    if (type instanceof GenericArrayType) {
+      return hasTypeVariable(((GenericArrayType) type).getGenericComponentType());
+    }
+    if (type instanceof TypeVariable) {
+      return true;
+    }
+    if (type instanceof WildcardType) {
+      return true;
+    }
+    String className = type == null ? "null" : type.getClass().getName();
+    throw new IllegalArgumentException("Expected a Class, ParameterizedType, or "
+        + "GenericArrayType, but <" + type + "> is of type " + className);
+  }
+
   // This method is copyright 2008 Google Inc. and is taken from Gson under the Apache 2.0 license.
   public static Class<?> getRawType(Type type) {
     if (type instanceof Class<?>) {
--- a/retrofit/src/main/java/retrofit/MethodInfo.java
+++ b/retrofit/src/main/java/retrofit/MethodInfo.java
@@ -236,6 +236,9 @@ final class MethodInfo {
           + factories);
     }
     Type responseType = adapter.responseType();
+    if (Utils.hasTypeVariable(responseType)) {
+      throw methodError("Method response type must not include a type variable.");
+    }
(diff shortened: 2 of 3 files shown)


JakeWharton:
Actually now that I'm thinking about this, we may want to disallow all type variables. There will never be a way of resolving the actual type the person wants, even with the bound. At best we could get an upper bound, but that's not helpful.

floating-cat:
Closes #443 -> Closes #433 ?

JakeWharton:
@floating-cat fixed. thanks.

retrofit/src/main/java/retrofit/Utils.java
```diff
@@ -81,6 +81,33 @@ public static Type getSingleParameterUpperBound(ParameterizedType type) {
...
+      return false;
+    }
+    if (type instanceof GenericArrayType) {
+      return hasTypeVariable(((GenericArrayType) type).getGenericComponentType());
+    }
+    if (type instanceof TypeVariable) {
+      return true;
+    }
+    if (type instanceof WildcardType) {
+      return true;
```
swankjesse: Not a type variable

PR #887, opened 2015-06-05
Adds retrofit-typedrequest optional module for builder style requests

felipecsl:
This pull request separates `MethodInfo` and `RestAdapter` from `Retrofit` (formerly `RestAdapter`), allowing different ways of declaring services. A new module retrofit-typedrequest has been added with a different implementation.
I know this is a pretty radical change and might not be within the interest of you guys to integrate into the library, however I figured that this kind of abstraction/extensibility could be useful in other scenarios.
In my particular case, the app already has **a lot** of "Volley style" request classes, each with their own classes and logic. This makes it virtually impossible to migrate to Retrofit given the way it requires you to define your "service" using an interface and annotations.
This change gives us that ability by simply implementing the new `Request` interface, which is just `CallAdapter` with an additional method.

### DIFF
--- a/retrofit/src/test/java/retrofit/RestAdapterTest.java
+++ b/retrofit/src/test/java/retrofit/RestAdapterTest.java
@@ -8,6 +8,10 @@ import com.squareup.okhttp.RequestBody;
 import com.squareup.okhttp.ResponseBody;
 import com.squareup.okhttp.mockwebserver.MockResponse;
 import com.squareup.okhttp.mockwebserver.MockWebServer;
+
+import org.junit.Rule;
+import org.junit.Test;
+
 import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.lang.annotation.Retention;
@@ -22,8 +26,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
-import org.junit.Rule;
-import org.junit.Test;
+
 import retrofit.http.Body;
 import retrofit.http.GET;
 import retrofit.http.POST;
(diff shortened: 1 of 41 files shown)


felipecsl:
can't figure out why Travis is complaining, though. This works fine running locally: `mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V`

swankjesse:
I don't think we're ready for a change like this right now. We're in the middle of a long sequence of changes in preparation for Retrofit 2, and this breaks some of that. Do you mind putting this on hold for a couple weeks until that's all landed?

swankjesse:
(No promises that we'll merge it at that time either, unfortunately)

felipecsl:
Sure, that makes sense! I'll keep rebasing against master to keep it in
sync in the meantime

felipecsl:
Rebased/squashed against master after #896 has been merged

retrofit/src/test/java/retrofit/RestAdapterTest.java
```diff
@@ -53,7 +56,7 @@
     Retrofit retrofit = new Retrofit.Builder()
         .endpoint(server.getUrl("/").toString())
         .build();
-    CallMethod example = retrofit.create(CallMethod.class);
+    CallMethod example = new RestAdapter(retrofit).create(CallMethod.class);
```
felipecsl: I'm not really happy with this API, open to suggestions

felipecsl:
i'm keeping this in sync with upstream in case there's any chance for this to be merged

JakeWharton:
We have no plans for a module like this at this time. Two completely separate mechanisms for making requests would only serve to confuse users. A standalone library depending solely on OkHttp is probably your best bet here.

felipecsl:
fair enough, however since this leverages most of the existing retrofit setup, we might just keep it as a fork for now.
thanks for the update

PR #890, opened 2015-06-06
Add a test to verify unused and potentially invalid path params are allowed.

### LINKED ISSUE #754: Feature request: Escaping { and } in url
I am using retrofit to access the Facebook API. Because Facebook uses curly braces {} within their query syntax (and retrofit does too), it is quite confusing to write against the Facebook API.

https://developers.facebook.com/docs/graph-api/using-graph-api/v2.2#fieldexpansion

This is how a Facebook request looks like:

    @GET("/me?fields=id,email,first_name,gender,last_name,link,locale,name,timezone,verified,picture.type(large){is_silhouette,url},age_range") JsonObject getUser(@Query("access_token") String userToken);

But because retrofit expects a parameter within {} I need to escape the curly braces meant for Facebook (which works, but looks kind of ugly and is hard to read):

    @GET("/me?fields=id,email,first_name,gender,last_name,link,locale,name,timezone,verified,picture.type(large)%7Bis_silhouette,url%7D,age_range") JsonObject getUser(@Query("access_token") String userToken);

Suggestion:

If you don't see any negative side effects, I would suggest allowing to escape { with {{ and } with }} if the curly braces are meant for the API and not for retrofit.

You could run a simple pre-check which replaces {{ with %7B and }} with %7D and the user would have a better readable endpoint definition:

    @GET("/me?fields=id,email,first_name,gender,last_name,link,locale,name,timezone,verified,picture.type(large){{is_silhouette,url}},age_range") JsonObject getUser(@Query("access_token") String userToken);

I admit it's a special use-case, but Facebook is not nobody ;-)

JakeWharton:
Closes #754.

### DIFF
--- a/retrofit/src/test/java/retrofit/RequestBuilderTest.java
+++ b/retrofit/src/test/java/retrofit/RequestBuilderTest.java
@@ -585,6 +585,20 @@ public final class RequestBuilderTest {
     assertThat(request.body()).isNull();
   }
 
+  @Test public void getWithUnusedAndInvalidNamedPathParam() {
+    class Example {
+      @GET("/foo/bar/{ping}/{kit,kat}/") //
+      Call<Object> method(@Path("ping") String ping) {
+        return null;
+      }
+    }
+    Request request = buildRequest(Example.class, "pong");
+    assertThat(request.method()).isEqualTo("GET");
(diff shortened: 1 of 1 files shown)


swankjesse:
Slightly dangerous, though I don't have better ideas.

PR #891, opened 2015-06-06
Require Endpoint as a constructor argument to RestAdapter.Builder

f2prateek:
Given that Endpoint is a required parameter, this makes it a required parameter when creating RestAdapter.Builder.

Also simplifies the construction for the dominant use case from `new RestAdapter.Builder().endpoint(ENDPOINT).build();` to `RestAdapter.builder(ENDPOINT).build();`.

### DIFF
--- a/retrofit/src/main/java/retrofit/Endpoint.java
+++ b/retrofit/src/main/java/retrofit/Endpoint.java
@@ -9,6 +9,9 @@ public abstract class Endpoint {
   /** Create an endpoint with the provided {@code url}. */
   public static Endpoint createFixed(String url) {
     checkNotNull(url, "url == null");
+    if (url.trim().length() == 0) {
+      throw new IllegalArgumentException("Empty URL");
+    }
     final HttpUrl httpUrl = HttpUrl.parse(url);
     if (httpUrl == null) {
       throw new IllegalArgumentException("Invalid URL: " + url);
--- a/retrofit/src/main/java/retrofit/RestAdapter.java
+++ b/retrofit/src/main/java/retrofit/RestAdapter.java
@@ -91,6 +91,15 @@ public final class RestAdapter {
   private final Converter converter;
   private final CallAdapter.Factory adapterFactory;
 
+  public static Builder builder(Endpoint endpoint) {
+    checkNotNull(endpoint, "endpoint == null");
+    return new Builder(endpoint);
+  }
+
+  public static Builder builder(String url) {
+    return new Builder(Endpoint.createFixed(url));
+  }
+
   private RestAdapter(OkHttpClient client, Endpoint endpoint, Converter converter,
       CallAdapter.Factory adapterFactory) {
     this.client = client;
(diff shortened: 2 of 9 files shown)


f2prateek:
If the endpoint is something that may change between multiple incovations of `build()`, we could make it an argument passed to the `build` method.

retrofit/src/main/java/retrofit/RestAdapter.java
```diff
@@ -170,28 +179,21 @@ public Converter converter() {
    * are optional.
```
JakeWharton: Javadoc needs updated. You can remove this whole paragraph.

retrofit/src/test/java/retrofit/EndpointsTest.java
```diff
@@ -10,4 +10,20 @@
     Endpoint endpoint = Endpoint.createFixed("http://example.com");
     assertThat(endpoint.url()).isEqualTo("http://example.com");
   }
+
+  @Test public void emptyUrlThrows() {
+    try {
+      Endpoint.createFixed(null);
```
JakeWharton: Missing `fail()`

retrofit/src/test/java/retrofit/EndpointsTest.java
```diff
@@ -10,4 +10,20 @@
...
+    } catch (NullPointerException expected) {
```
JakeWharton: Normally we name these just `e` when the type is used. `expected` is for when we expect it but don't want to waste the comment in the empty `catch` block.

retrofit/src/test/java/retrofit/EndpointsTest.java
```diff
@@ -10,4 +10,20 @@
...
+      assertThat(expected).hasMessage("url == null");
+    }
+  }
+
+  @Test public void nullUrlThrows() {
+    try {
+      Endpoint.createFixed(" ");
```
JakeWharton: Missing `fail()`

JakeWharton:
Also needs a rebase.

f2prateek:
Patched and rebased.

retrofit/src/main/java/retrofit/Endpoint.java
```diff
@@ -8,7 +8,10 @@
 public abstract class Endpoint {
   /** Create an endpoint with the provided {@code url}. */
   public static Endpoint createFixed(String url) {
-    checkNotNull(url, "url == null");
+    checkNotNull(url, "URL == null");
```
JakeWharton: This was lowercase to denote the method parameter name. The others referenced the uppercase acronym as a noun.

retrofit/src/main/java/retrofit/Endpoint.java
```diff
@@ -8,7 +8,10 @@
...
+    if (url.trim().length() == 0) {
+      throw new IllegalArgumentException("URL is empty");
```
f2prateek: @JakeWharton From aboe should this be lowercase as well or should I leave it as is?

JakeWharton: This can stay upper.

f2prateek:
With the new HttpUrl stuff, does it make sense to add a `RestAdapter.builder(HttpUrl)` overload as well?

JakeWharton:
Sorry for the delay. I've been mulling the ramifications of this over. I realize what motivates this change. The change seems obvious to me as well on first look, but after some thought I've come down against it. Here's some reasons why:

 * While the endpoint is currently the only required configuration, this was not always the case nor can we guarantee that it will always be the case. Rather than deal with deprecating the static methods and creating more overloads I prefer the flexibility of methods on the builder class.
 * A static method named `builder` does not describe the argument that it takes. Ideally the method name would talk about its behavior but `newBuilderForEndpoint` or `builderForEndpoint` is a bit too verbose for my liking.
 * It splits the APIs around building a `RestAdapter` between the `RestAdapter` type and the `RestAdapter.Builder`. These are not in the same place in the documentation which is unfortunate.

That said, an `HttpUrl` overload for the builder method would be welcome.

f2prateek:
Those sound like fair points, thanks for explaining your reasoning!

PR #893, opened 2015-06-08
Use HttpUrl for Endpoint and building URL.

JakeWharton:


retrofit/src/test/java/retrofit/EndpointsTest.java
```diff
@@ -1,13 +1,24 @@
...
 import org.junit.Test;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.fail;
 
 public final class EndpointsTest {
   @Test public void endpoint() {
     Endpoint endpoint = Endpoint.createFixed("http://example.com");
-    assertThat(endpoint.url()).isEqualTo("http://example.com");
+    assertThat(endpoint.url()).isEqualTo(HttpUrl.parse("http://example.com"));
```
swankjesse: slightly more correct to use a trailing slash, like

`http://example.com/`

PR #898, opened 2015-06-11
Accept an HttpUrl in the endpoint builder method.

JakeWharton:
Also add tests for all the builder methods and their counterparts in the final Retrofit class.


JakeWharton:
Refs #891.

retrofit/src/main/java/retrofit/Endpoint.java
```diff
@@ -2,29 +2,13 @@
...
-      throw new IllegalArgumentException("Invalid URL: " + url);
-    }
-    return new Endpoint() {
-      @Override public HttpUrl url() {
-        return httpUrl;
-      }
-    };
-  }
-
+public interface Endpoint {
```
swankjesse: Do we need an interface? We could just use the `HttpUrl`?

swankjesse: (though then it isn't mutable)

swankjesse: (and cannot be computed on-demand)

JakeWharton: Yeah it needs to be able to change over time. We use it internally and there's a few external consumers of the behavior.

PR #902, opened 2015-06-15
Added fileName member to TypedByteArray

### LINKED ISSUE #787: Allow to specify filename as optional parameter to TypedByteArray
I would expect to be able to use `new TypedByteArray(type, bytes, "myfilename.foo")` to upload a file multi-part using RetroFit from an in-memory file instead of a `File` object.

Right now one needs to subclass using work-around described here: https://github.com/square/retrofit/issues/178#issuecomment-34470409

pepijntb:
This would fix #787


RoyMontoya:
looks good!

JakeWharton:
TypedByteArray is long gone. As to how filename is specified now, this is being tracked by #1140

PR #903, opened 2015-06-16
Include a request body for content type and length.

### LINKED ISSUE #698: Conversion error on empty body in response
Retrofit version 1.8.0.

I have a query which in some cases returns a response body and other cases has an empty response body. Currently it seems that Retrofit RestAdapter does not correctly handle empty response body. I would expect a null object to be returned by RestAdapter perhaps around line 353 but currently the empty body is being passed to a converter on line 362 (JacksonConverter in my case) which then throws a JsonMappingException since the InputStream is empty. Maybe line 353 should be changed to also check the length of the body and if it is 0 then return null? I also thought about creating a local JacksonConverter to return null on content with length 0 but then logResponseBody on line 363 would throw a NullPointerException when logging is turned on.

JakeWharton:
Second commit closes #698.

### DIFF
--- a/retrofit/src/main/java/retrofit/NoContentResponseBody.java
+++ b/retrofit/src/main/java/retrofit/NoContentResponseBody.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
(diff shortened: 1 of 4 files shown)


swankjesse:
Dig.

PR #908, opened 2015-06-19
Introduce factory for converter concept.

JakeWharton:
This is still highly-inefficient in its internal use and there are no tests proving that we now fail earlier in configuring a service method. Both of those will come in follow up commits


JakeWharton:
If you see old copyright dates in what looks like a new file, it's most likely a git mismatch or a file that split into two.

retrofit-adapters/rxjava/src/test/java/retrofit/ObservableCallAdapterFactoryTest.java
```diff
@@ -195,13 +195,17 @@
     }
   }
 
-  static class StringConverter implements Converter {
-    @Override public Object fromBody(ResponseBody body, Type type) throws IOException {
-      return body.string();
-    }
+  static class StringConverterFactory implements Converter.Factory {
```
swankjesse: oooh interesting, no type arguments.

retrofit-converters/gson/src/main/java/retrofit/GsonConverter.java
```diff
@@ -15,71 +15,55 @@
...
 
-  @Override public Object fromBody(ResponseBody body, Type type) throws IOException {
+  @Override public Object fromBody(ResponseBody body) throws IOException {
     Charset charset = this.charset;
     if (body.contentType() != null) {
       charset = body.contentType().charset(charset);
     }
 
-    InputStream is = body.byteStream();
+    Reader in = new InputStreamReader(body.byteStream(), charset);
```
swankjesse: Can we get closer with `body.charStream()` ? It already uses the HTTP response's charset

swankjesse: (though it doesn't accept a default to be passed in; it allows only UTF-8 as default)

swankjesse: (we should offer `ResponseBody.charStream(Charset defaultCharset)` if we think that that's a thing)

retrofit-converters/gson/src/main/java/retrofit/GsonConverter.java
```diff
@@ -15,71 +15,55 @@
...
       }
     }
   }
 
-  @Override public RequestBody toBody(Object object, Type type) {
-    String json = gson.toJson(object, type);
-    return RequestBody.create(mediaType, json);
+  @Override public RequestBody toBody(Object object) {
+    Buffer buffer = new Buffer();
+    Writer writer = new OutputStreamWriter(buffer.outputStream(), charset);
```
swankjesse: Should we add `BufferedSource.writer(Charset)` ? It could do UTF-8 natively.

retrofit-converters/gson/src/main/java/retrofit/GsonConverter.java
```diff
@@ -15,71 +15,55 @@
...
+    try {
+      typeAdapter.toJson(writer, object);
+      writer.flush();
+    } catch (IOException e) {
+      throw new AssertionError(e); // Writing to String does no I/O.
```
swankjesse: Writing to a _buffer_ does no I/O.

retrofit-converters/gson/src/main/java/retrofit/GsonConverterFactory.java
```diff
@@ -0,0 +1,67 @@
...
+
+import com.google.gson.Gson;
+import com.google.gson.TypeAdapter;
+import com.google.gson.reflect.TypeToken;
+import com.squareup.okhttp.MediaType;
+import java.lang.reflect.Type;
+import java.nio.charset.Charset;
+
+/** A {@linkplain Converter.Factory converter} which uses Gson for JSON. */
+public final class GsonConverterFactory implements Converter.Factory {
```
swankjesse: does this want to be a static method on GsonConverter?

```
ConverterFactory asciiGson = GsonConverter.factory(Charsets.ASCII);
```

retrofit-converters/jackson/src/main/java/retrofit/JacksonConverterFactory.java
```diff
@@ -0,0 +1,49 @@
...
+package retrofit;
+
+import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectReader;
+import com.fasterxml.jackson.databind.ObjectWriter;
+import java.lang.reflect.Type;
+
+/** A {@linkplain Converter.Factory converter} which uses Jackson. */
+public final class JacksonConverterFactory implements Converter.Factory {
```
swankjesse: Similarly, does this really need its own class?

swankjesse: (we'd never refer to it by type, only by the interface it implements)

JakeWharton: To be fair, we also don't care about the `JacksonConverter` type, only the fact that it is a `Converter` instance.

swankjesse:
LGTM

retrofit-adapters/rxjava/src/test/java/retrofit/ObservableCallAdapterFactoryTest.java
JakeWharton: Added to `Converter`

PR #911, opened 2015-06-20
Refactor the core of request builder into parameter actions.

JakeWharton:
This pushes a ton of work forward into the method parsing step so that when a request is needed its assembly has almost no logic.

Still work to do around killing `MethodInfo`'s mutableness.

### DIFF
--- a/retrofit/src/main/java/retrofit/OkHttpCall.java
+++ b/retrofit/src/main/java/retrofit/OkHttpCall.java
@@ -111,7 +111,15 @@ final class OkHttpCall<T> implements Call<T> {
   private com.squareup.okhttp.Call createRawCall() {
     HttpUrl url = endpoint.url();
     RequestBuilder requestBuilder = new RequestBuilder(url, methodInfo);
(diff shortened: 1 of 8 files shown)


retrofit/src/main/java/retrofit/OkHttpCall.java
```diff
@@ -111,7 +111,15 @@ private void callSuccess(Response<T> response) {
   private com.squareup.okhttp.Call createRawCall() {
     HttpUrl url = endpoint.url();
     RequestBuilder requestBuilder = new RequestBuilder(url, methodInfo);
-    requestBuilder.setArguments(args);
+
+    Object[] args = this.args;
+    if (args != null) {
+      RequestBuilderAction[] actions = methodInfo.requestBuilderActions;
+      for (int i = 0, count = args.length; i < count; i++) {
+        actions[i].perform(requestBuilder, args[i]);
```
swankjesse: awesome

swankjesse:
This is great.

Issue #907, opened 2015-06-18
Retrofit drops context path from endpoint url


unguiculus:
I'm currently checking out Retrofit with the new Moshi converter. 

Retrofit 2.0.0-SNAPSHOT
OkHttp 2.4.0
Okio 1.4.0
Moshi 0.9.0

Say, I have the following service:

```Java
public interface MyService {
  @POST("/")
  Call<MyResult> post(@Body MyRequest request);
}
```

The request is executed as follows:

```Java
Retrofit retrofit = new Retrofit.Builder()
    .endpoint("http://localhost:8778/mycontext/")
    .converter(new MoshiConverter())
    .build();

Call<MyResult> call = retrofit.create(MyService.class).post(new MyRequest());
Response<MyResult> response = call.execute();
```

The result is a `404` because the context (`/mycontext`) is dropped from the URL and the post goes to `http://localhost:8778/`. This happens in https://github.com/square/retrofit/blob/master/retrofit/src/main/java/retrofit/RequestBuilder.java#L333 when [encodedPath(String)](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java#L731) is called. Shouldn't [addEncodedPathSegment(String)](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java#L691) be called instead?

swankjesse:
Maybe we should trim/forbid the leading slash on the relative path? If we treat the whole thing as a relative URL there are interesting consequences. For example, this works:
```
  @POST("https://host/path/{parameter}")
```
And this:
```
  @POST("../path/{parameter}")
```
Though looking at it, I think I'm convincing myself that what we have in retrofit 1.x is simpler. Maybe just require a trailing shash in the endpoint and forbid it in the annotation?

gustavohenke:
So, this issue has been fixed in #927.
But how is the fix supposed to work? I'm new to Android/Retrofit world.

- Base URL is `http://server.com/api`
- API interface has `@GET("/user")`
- Result: the request goes to `http://server.com/user`.

I'm considering using an interceptor to rewrite the URL (if that's even possible...).

I'm using Retrofit `2.0.0-beta1`.

JakeWharton:
Add trailing slash to base URL, remove leading slash from relative URL.

gustavohenke:
OK, thank you! I was using a Builder to create that URL, so I added an empty path segment.

PR #912, opened 2015-06-21
Switch the architecture of method parsing and invocation.

JakeWharton:
This makes all of the mutability during parsing temporary and creates the immutable MethodInvoker.

### DIFF
--- a/retrofit/src/main/java/retrofit/RequestBuilder.java
+++ b/retrofit/src/main/java/retrofit/RequestBuilder.java
@@ -28,51 +28,44 @@ import java.net.URLEncoder;
 import okio.BufferedSink;
 
 final class RequestBuilder {
-  private final String requestMethod;
-  private final boolean requestHasBody;
+  private final String method;
   private final HttpUrl.Builder urlBuilder;
-  private final Request.Builder requestBuilder;
+  private String pathUrl;
 
+  private final Request.Builder requestBuilder;
+  private MediaType mediaType;
+
+  private final boolean hasBody;
   private MultipartBuilder multipartBuilder;
   private FormEncodingBuilder formEncodingBuilder;
   private RequestBody body;
 
-  private String relativeUrl;
-  private MediaType mediaType;
+  RequestBuilder(String method, HttpUrl url, String pathUrl, String queryParams, Headers headers,
+      MediaType mediaType, boolean hasBody, boolean isFormEncoded, boolean isMultipart) {
+    this.method = method;
 
-  RequestBuilder(HttpUrl url, MethodInfo methodInfo) {
-    requestMethod = methodInfo.requestMethod;
-    requestHasBody = methodInfo.requestHasBody;
-    mediaType = methodInfo.mediaType;
-    relativeUrl = methodInfo.requestUrl;
+    HttpUrl.Builder urlBuilder = url.newBuilder();
+    if (queryParams != null) {
+      urlBuilder.query(queryParams);
+    }
+    this.urlBuilder = urlBuilder;
+    this.pathUrl = pathUrl;
 
-    urlBuilder = url.newBuilder();
-    requestBuilder = new Request.Builder();
-
-    Headers headers = methodInfo.headers;
+    Request.Builder requestBuilder = new Request.Builder();
     if (headers != null) {
       requestBuilder.headers(headers);
     }
+    this.requestBuilder = requestBuilder;
+    this.mediaType = mediaType;
 
-    String requestQuery = methodInfo.requestQuery;
-    if (requestQuery != null) {
-      urlBuilder.query(requestQuery);
-    }
+    this.hasBody = hasBody;
 
-    switch (methodInfo.bodyEncoding) {
-      case FORM_URL_ENCODED:
-        // Will be set to 'body' in 'build'.
-        formEncodingBuilder = new FormEncodingBuilder();
-        break;
-      case MULTIPART:
-        // Will be set to 'body' in 'build'.
-        multipartBuilder = new MultipartBuilder();
-        break;
-      case NONE:
-        // If present, 'body' will be set in 'setArguments' call.
-        break;
-      default:
-        throw new IllegalArgumentException("Unknown request type: " + methodInfo.bodyEncoding);
+    if (isFormEncoded) {
+      // Will be set to 'body' in 'build'.
+      formEncodingBuilder = new FormEncodingBuilder();
+    } else if (isMultipart) {
+      // Will be set to 'body' in 'build'.
+      multipartBuilder = new MultipartBuilder();
     }
   }
 
@@ -92,9 +85,9 @@ final class RequestBuilder {
         // encode spaces rather than +. Query encoding difference specified in HTML spec.
         // Any remaining plus signs represent spaces as already URLEncoded.
         encodedValue = encodedValue.replace("+", "%20");
-        relativeUrl = relativeUrl.replace("{" + name + "}", encodedValue);
+        pathUrl = pathUrl.replace("{" + name + "}", encodedValue);
       } else {
-        relativeUrl = relativeUrl.replace("{" + name + "}", String.valueOf(value));
+        pathUrl = pathUrl.replace("{" + name + "}", String.valueOf(value));
       }
     } catch (UnsupportedEncodingException e) {
       throw new RuntimeException(
@@ -127,7 +120,8 @@ final class RequestBuilder {
   }
 
   Request build() {
-    urlBuilder.encodedPath(relativeUrl);
+    // TODO this should append, not replace.
+    HttpUrl url = urlBuilder.encodedPath(pathUrl).build();
 
     RequestBody body = this.body;
     if (body == null) {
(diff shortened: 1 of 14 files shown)


retrofit/src/main/java/retrofit/MethodInvoker.java
```diff
@@ -0,0 +1,97 @@
...
+
+import com.squareup.okhttp.OkHttpClient;
+import com.squareup.okhttp.ResponseBody;
+import java.lang.reflect.Method;
+import java.lang.reflect.Type;
+import retrofit.http.Streaming;
+
+import static retrofit.Utils.methodError;
+
+final class MethodInvoker<T> {
```
JakeWharton: This class probably shouldn't be parameterized, but it allows storing the fact that the `CallAdapter` and `Converter` are of the same parameterized type which makes instantiating the `OkHttpCall` simple. Open to suggestions here.

retrofit/src/main/java/retrofit/RequestBuilder.java
```diff
@@ -127,7 +133,8 @@ void setBody(RequestBody body) {
   }
 
   Request build() {
-    urlBuilder.encodedPath(relativeUrl);
+    // TODO this should append, not replace.
```
JakeWharton: This is #907.

retrofit/src/main/java/retrofit/MethodInvoker.java
JakeWharton: Also not sure about the name since it's not invoking a method... `MethodHandler` maybe? `InvocationHandler` is about ideal but it's taken.

retrofit/src/main/java/retrofit/MethodInvoker.java
```diff
@@ -0,0 +1,96 @@
...
+      throw methodError(method, "Method response type is "
+          + responseType
+          + " but no converter registered. "
+          + "Either add a converter to the Retrofit instance or use ResponseBody.");
+    }
+
+    Converter<?> converter = converterFactory.get(responseType);
+    if (converter == null) {
+      throw methodError(method,
+          "Registered converter factory was unable to handle response type " + responseType);
```
swankjesse: include the factory in the message?

swankjesse: maybe also include the method for context ?

swankjesse: derp, methodError does that! nice.

retrofit/src/main/java/retrofit/MethodInvoker.java
```diff
@@ -0,0 +1,96 @@
...
+  private MethodInvoker(OkHttpClient client, RequestFactory requestFactory,
+      CallAdapter<T> callAdapter, Converter<T> responseConverter) {
+    this.client = client;
+    this.requestFactory = requestFactory;
+    this.callAdapter = callAdapter;
+    this.responseConverter = responseConverter;
+  }
+
+  Object invoke(Object... args) {
+    return callAdapter.adapt(new OkHttpCall<>(client, requestFactory, responseConverter, args));
```
swankjesse: this is awesome

retrofit/src/main/java/retrofit/RequestFactoryParser.java
```diff
@@ -45,155 +44,143 @@
...
-    parseParameters();
   }
 
-  private RuntimeException methodError(String message, Object... args) {
-    if (args.length > 0) {
-      message = String.format(message, args);
-    }
-    return new IllegalArgumentException(
-        method.getDeclaringClass().getSimpleName() + "." + method.getName() + ": " + message);
+  private RequestFactory toMethodInfo(Endpoint endpoint) {
```
swankjesse: toRequestFactory ?

retrofit/src/main/java/retrofit/Retrofit.java
```diff
@@ -108,35 +108,29 @@ private Retrofit(OkHttpClient client, Endpoint endpoint, Converter.Factory conve
...
-
-  private MethodInfo loadMethodInfo(Method method) {
-    MethodInfo methodInfo = methodInfoCache.get(method);
-    if (methodInfo == null) {
-      synchronized (methodInfoCache) {
-        methodInfo = methodInfoCache.get(method);
-        if (methodInfo == null) {
-          methodInfo = new MethodInfo(method, adapterFactory, converterFactory);
-          methodInfoCache.put(method, methodInfo);
+  // Package-private avoids synthetic accessor method from InvocationHandler.
```
swankjesse: This comment really shouldn't be necessary right here. (it's irrelevant)

retrofit/src/main/java/retrofit/Retrofit.java
```diff
@@ -108,35 +108,29 @@ private Retrofit(OkHttpClient client, Endpoint endpoint, Converter.Factory conve
...
+  MethodInvoker<?> loadMethodInvoker(Method method) {
+    MethodInvoker<?> methodInvoker = methodInvokerCache.get(method);
```
swankjesse: reading a LinkedHashMap while another thread is writing it is sketcy

retrofit/src/main/java/retrofit/Utils.java
```diff
@@ -151,6 +152,14 @@ public static boolean hasUnresolvableType(Type type) {
     }
   }
 
+  static RuntimeException methodError(Method method, String message, Object... args) {
+    if (args.length > 0) {
+      message = String.format(message, args);
```
swankjesse: Do this unconditionally? Otherwise things like `%n` are weird

retrofit/src/main/java/retrofit/MethodInvoker.java
```diff
@@ -0,0 +1,96 @@
...
+  private static Converter<?> createResponseConverter(Method method, Type responseType,
+      Converter.Factory converterFactory) {
+    if (responseType == ResponseBody.class) {
+      boolean isStreaming = method.isAnnotationPresent(Streaming.class);
+      return new OkHttpResponseBodyConverter(isStreaming);
+    }
+
+    if (converterFactory == null) {
+      throw methodError(method, "Method response type is "
+          + responseType
```
swankjesse: use `%s` here

retrofit/src/main/java/retrofit/MethodInvoker.java
swankjesse: Use %s here

retrofit/src/main/java/retrofit/MethodInvoker.java
```diff
@@ -0,0 +1,96 @@
...
+            converterFactory);
+    RequestFactory requestFactory = RequestFactoryParser.parse(method, endpoint, converterFactory);
+    return new MethodInvoker<>(client, requestFactory, callAdapter, responseConverter);
+  }
+
+  private static CallAdapter<?> createCallAdapter(Method method,
+      CallAdapter.Factory adapterFactory) {
+    Type returnType = method.getGenericReturnType();
+    if (Utils.hasUnresolvableType(returnType)) {
+      throw methodError(method, "Method return type must not include a type variable or wildcard.");
```
swankjesse: include the return type!

retrofit/src/main/java/retrofit/MethodInvoker.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    if (returnType == void.class) {
+      throw methodError(method, "Service methods cannot return void.");
+    }
+
+    CallAdapter<?> adapter = adapterFactory.get(returnType);
+    if (adapter == null) {
+      throw methodError(method,
+          "Registered call adapter factory was unable to handle return type " + returnType);
```
swankjesse: `%s`

retrofit/src/main/java/retrofit/Utils.java
```diff
@@ -151,6 +152,14 @@ public static boolean hasUnresolvableType(Type type) {
...
+    }
+    return new IllegalArgumentException(
+        method.getDeclaringClass().getSimpleName() + "." + method.getName() + ": " + message);
```
swankjesse: Would this be better wrapping the method name afterwards?

```
IllegalArgumentException: GitHubService.listProjects: @Headers value must be in the form \"Name: Value\". Found: \"Authorization; key\"
```
vs.

```
IllegalArgumentException: @Headers value must be in the form \"Name: Value\". Found: \"Authorization; key\"
    for method GitHubService.listProjects()
```

retrofit/src/main/java/retrofit/Retrofit.java
swankjesse: Just compute 'em all eagerly?

retrofit/src/main/java/retrofit/MethodInvoker.java
swankjesse: `MethodInvocationHandler`, `MethodHandler`, `MethodAdapter`

swankjesse: It's definitely not a `MethodInvoker`, that's a thing that calls methods, not a thing that responds to method calls

swankjesse:
LGTM

This is landing in a very good place.

retrofit/src/main/java/retrofit/Retrofit.java
JakeWharton: Eagerly will cause slowdown on things like app boot. We're already moving validation into the synchronous call path of a method whereas before we would defer it to happen on the background thread on first invocation.

What about a `ConcurrentMap`, no synchronization, and using `putIfAbsent` instead?

swankjesse: Just use `synchronized`. It's fast unless there's contention, and there isn't going to be contention.

PR #915, opened 2015-06-22
Fm/add cookie headers

FilippoMito:
Added a convenient annotation to manage the "Cookie" header.

### DIFF
--- a/retrofit/src/main/java/retrofit/RequestBuilder.java
+++ b/retrofit/src/main/java/retrofit/RequestBuilder.java
@@ -32,7 +32,7 @@ import retrofit.client.Request;
 import retrofit.converter.Converter;
 import retrofit.http.Body;
 import retrofit.http.Cookie;
-import retrofit.http.Cookies;
+import retrofit.http.CookieMap;
 import retrofit.http.EncodedPath;
 import retrofit.http.EncodedQuery;
 import retrofit.http.EncodedQueryMap;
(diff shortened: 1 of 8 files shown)


swankjesse:
Please avoid investing significant time in pull requests before consulting with the project maintainers on the features benefits and implementation.

In this case I think the entire feature is a mistake; that's not how cookies work. Instead they should be supplied by the webserver and transparent to the application.

FilippoMito:
> Please avoid investing significant time in pull requests before consulting with the project maintainers on the features benefits and implementation.

You're right, I'm sorry.

> In this case I think the entire feature is a mistake; that's not how cookies work. Instead they should be supplied by the webserver and transparent to the application.

I needed a way to avoid typing the "Cookie: " string everywhere (or like mess with the code in a custom *RequestInterceptor*) i need to send a cookie to the server. Looking at the Java EE 7 Specification APIs [The JAX-RS client API](https://docs.oracle.com/javaee/7/api/javax/ws/rs/client/Invocation.Builder.html) I found that requests can be constructed by explicitly specifying the cookie(s), even if it is just a header, so I thought it was a sensible way of handling this. 

On the plus side, it seems that other developers (at least judging by the stackoverflow questions) have been asking for this feature.

Please note that this would not be a method for automatic cookie handling (i.e. custom CookieManager): I am just providing a way to supply custom cookies *per request*.

swankjesse:
Obvious question: why provide cookies per request?

FilippoMito:
Because it is possible that some resource don't need that cookie, so providing it with a *RequestInterceptor* or *CookieManager* every time is overkill.

There are no constraints in the architectural properties of ReST that forces your client to act as an user agent in a [HTTP Cookie](https://tools.ietf.org/html/rfc6265)-style. In fact the JAX-RS client API not only allow but encourage that behavior (with a dedicated method in the invocation builder).

So, why provide cookies for **every** request? I believe that a per-method logic is always preferable.

FilippoMito:
Since this is going nowhere... Would you mind closing (or accepting) the request?

swankjesse:
Closing, mostly because I don't fully understand the use case, and worry about the consequences in interactions with cookies in lower-layers of the system, such as the HTTP client's cookie store.

PR #925, opened 2015-06-27
Allow passing in an executor on which callback methods are invoked.

JakeWharton:
This only has an effect when Call is the return type of a service method. This does not affect the callback methods of a Call passed to a CallAdapter.

We also change the non-Android call adapter factory to be a simple implementation that returns the original instance rather than needlessly wrapping it in the executor factory with a synchronous executor.

### DIFF
--- a/retrofit/src/test/java/retrofit/RetrofitTest.java
+++ b/retrofit/src/test/java/retrofit/RetrofitTest.java
@@ -14,7 +14,10 @@ import java.lang.reflect.Type;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.Executor;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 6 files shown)


JakeWharton:
There's an alternative implementation here: we could have a `callbackExecutor` method on `Retrofit.Builder` (similar to what we do now). The interesting case there is that it would only be applied to `Call` instances which are exposed in the service interface API. If you have your own call adapter, that method would be useless as the callbacks would still happen on the HTTP thread and not in your executor.

swankjesse:
Two executors: the network one, and the "UI" one. OkHttp let's us configure the executor in its dispatcher, so perhaps Retrofit could expose both for configuration.

swankjesse:
I don't think this is the right API, but I'm unsure. Feels like we should let the user pass on Executors, not factories.

JakeWharton:
Updated. PTAL.

retrofit/src/test/java/retrofit/RetrofitTest.java
```diff
@@ -448,4 +457,91 @@
...
+    } catch (NullPointerException e) {
+      assertThat(e).hasMessage("callbackExecutor == null");
+    }
+  }
+
+  @Test public void callbackExecutorNoDefault() {
+    Retrofit retrofit = new Retrofit.Builder()
+        .endpoint("http://example.com/")
+        .build();
+    assertThat(retrofit.callbackExecutor()).isNull();
```
swankjesse: Nice test.

swankjesse:
LGTM

PR #927, opened 2015-06-28
Add Url annotation for dynamic resolution.

### LINKED ISSUE #333: Follow links
Retrofit works great and probably has the cleanest api as far as rest clients goes in the java world. The one limitation is that it's really hard to use to follow links. I understand that it probably wasn't a design goal but would it be something you'd consider? I could easily call the url directly but I really want to be able to use the error handling, logging, converters and everything else configured in the RestAdapter. 

Currently you can do:
```
@GET("/{path}")
SomeObject get(@EncodedPath("path") String path);
```
It works well but this means you can only follow links going to the server configured with the RestAdapter. What do you think about adding a Url annotation?
```
@GET 
SomeObject get(@Url String url);

@POST
void post(@Url String url, @Body SomeObject object);
```

### LINKED ISSUE #546: Wrong behavior when joining endpoints with paths
Retrofit disrespects a number of rules from the [RFC 1808 "Relative Uniform Resource Locators" section 4 (Resolving Relative URLs)](http://tools.ietf.org/html/rfc1808#section-4):

Retrofit treats absolute paths as relative:
```
   Step 4: If the embedded URL path is preceded by a slash "/", the
           path is not relative and we skip to Step 7.
```

Retrofit rejects empty paths, and also those not starting with a slash:
```
Step 5: If the embedded URL path is empty (and not preceded by a
           slash), then the embedded URL inherits the base URL path

Step 7: The resulting URL components, including any inherited from
           the base URL, are recombined to give the absolute form of
           the embedded URL.
```

And maybe more.

The [RFC's section 5.1](http://tools.ietf.org/html/rfc1808#section-5) contains many examples of the expected behavior.

Does Retrofit purposedly discomplies with the URL spec, or is this a bug?
If this is a bug, a simple solution would be to use Java's ```new URL(URL, String)``` instead of [manually joining URLs](https://github.com/square/retrofit/blob/e87ad3e25925bea4f484d853052fa8f63ce6fa26/retrofit/src/main/java/retrofit/RequestBuilder.java#L327).

### LINKED ISSUE #907: Retrofit drops context path from endpoint url

JakeWharton:
This also corrects an incorrect behavior of replacing any path parts in the base URL with the relative URL.

Closes #907. Closes #333. Closes #546.

### DIFF
--- a/retrofit/src/main/java/retrofit/RequestFactoryParser.java
+++ b/retrofit/src/main/java/retrofit/RequestFactoryParser.java
@@ -44,6 +44,7 @@ import retrofit.http.PartMap;
 import retrofit.http.Path;
 import retrofit.http.Query;
 import retrofit.http.QueryMap;
+import retrofit.http.Url;
 
 import static retrofit.Utils.methodError;
 
@@ -67,21 +68,20 @@ final class RequestFactoryParser {
   private boolean hasBody;
   private boolean isFormEncoded;
   private boolean isMultipart;
-  private String pathUrl;
-  private String queryParams;
+  private String relativeUrl;
   private com.squareup.okhttp.Headers headers;
   private MediaType mediaType;
   private RequestBuilderAction[] requestBuilderActions;
 
-  private Set<String> pathUrlParamNames;
+  private Set<String> relativeUrlParamNames;
 
   private RequestFactoryParser(Method method) {
     this.method = method;
   }
 
   private RequestFactory toRequestFactory(BaseUrl baseUrl) {
-    return new RequestFactory(httpMethod, baseUrl, pathUrl, queryParams, headers, mediaType,
-        hasBody, isFormEncoded, isMultipart, requestBuilderActions);
+    return new RequestFactory(httpMethod, baseUrl, relativeUrl, headers, mediaType, hasBody,
+        isFormEncoded, isMultipart, requestBuilderActions);
   }
 
   private RuntimeException parameterError(int index, String message, Object... args) {
@@ -145,32 +145,27 @@ final class RequestFactoryParser {
       throw methodError(method, "Only one HTTP method is allowed. Found: %s and %s.",
           this.httpMethod, httpMethod);
     }
-    if (value == null || value.length() == 0 || value.charAt(0) != '/') {
-      throw methodError(method, "URL path \"%s\" must start with '/'.", value);
+    this.httpMethod = httpMethod;
+    this.hasBody = hasBody;
+
+    if (value.isEmpty()) {
+      return;
     }
 
     // Get the relative URL path and existing query string, if present.
-    String pathUrl = value;
-    String queryParams = null;
     int question = value.indexOf('?');
     if (question != -1 && question < value.length() - 1) {
-      pathUrl = value.substring(0, question);
-      queryParams = value.substring(question + 1);
-
       // Ensure the query string does not have any named parameters.
+      String queryParams = value.substring(question + 1);
       Matcher queryParamMatcher = PARAM_URL_REGEX.matcher(queryParams);
       if (queryParamMatcher.find()) {
-        throw methodError(method,
-            "URL query string \"%s\" must not have replace block. For dynamic query"
-                + " parameters use @Query.", queryParams);
+        throw methodError(method, "URL query string \"%s\" must not have replace block. "
+            + "For dynamic query parameters use @Query.", queryParams);
       }
     }
 
-    this.httpMethod = httpMethod;
-    this.hasBody = hasBody;
-    this.pathUrl = pathUrl;
-    this.pathUrlParamNames = parsePathParameters(pathUrl);
-    this.queryParams = queryParams;
+    this.relativeUrl = value;
+    this.relativeUrlParamNames = parsePathParameters(value);
   }
 
   private com.squareup.okhttp.Headers parseHeaders(String[] headers) {
@@ -199,6 +194,9 @@ final class RequestFactoryParser {
     boolean gotField = false;
     boolean gotPart = false;
     boolean gotBody = false;
+    boolean gotPath = false;
+    boolean gotQuery = false;
+    boolean gotUrl = false;
 
     int count = methodParameterAnnotationArrays.length;
     RequestBuilderAction[] requestBuilderActions = new RequestBuilderAction[count];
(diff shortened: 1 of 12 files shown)


retrofit/src/main/java/retrofit/RequestFactoryParser.java
```diff
@@ -211,7 +209,38 @@ private void parseParameters(Converter.Factory converterFactory) {
             throw parameterError(i, "Multiple Retrofit annotations found, only one allowed.");
           }
 
-          if (methodParameterAnnotation instanceof Path) {
+          if (methodParameterAnnotation instanceof Url) {
+            if (gotUrl) {
+              throw parameterError(i, "Multiple @Url method annotations found.");
+            }
+            if (gotPath) {
```
swankjesse: Is there an order-of-parameters problem here? Ie. if I have `@Path` and then `@Url`, or the opposite?

retrofit/src/main/java/retrofit/RequestFactoryParser.java
```diff
@@ -211,7 +209,38 @@ private void parseParameters(Converter.Factory converterFactory) {
...
+              throw parameterError(i, "@Url cannot be used with @%s URL", httpMethod);
+            }
+            gotUrl = true;
+            requestBuilderActions[i] = new RequestBuilderAction.Url();
+
+          } else if (methodParameterAnnotation instanceof Path) {
+            if (gotQuery) {
+              throw parameterError(i, "A @Path parameter must not come after a @Query.");
+            }
+            if (gotUrl) {
```
swankjesse: aahhhh, here it is.

retrofit/src/test/java/retrofit/RequestBuilderTest.java
```diff
@@ -864,6 +864,147 @@
...
 
+  @Test public void getWithUrl() {
+    class Example {
+      @GET
+      Call<Object> method(@Url String url) {
+        return null;
+      }
+    }
+
+    Request request = buildRequest(Example.class, "foo/bar/");
```
swankjesse: dig

retrofit/src/test/java/retrofit/RequestBuilderTest.java
```diff
@@ -864,6 +864,147 @@
...
+  @Test public void getWithNonStringUrlThrows() {
+    class Example {
+      @GET
+      Call<Object> method(@Url Object url) {
+        return null;
+      }
+    }
+
+    try {
+      buildRequest(Example.class, "foo/bar");
```
swankjesse: fail

retrofit/src/test/java/retrofit/RequestBuilderTest.java
```diff
@@ -731,6 +715,22 @@
...
+  @Test public void getWithQueryThenPathThrows() {
+    class Example {
+      @GET("/foo/bar/{ping}/") //
+      Call<Object> method(@Query("kit") String kit, @Path("ping") String ping) {
+        return null;
+      }
+    }
+
+    try {
+      buildRequest(Example.class, "kat", "pong");
```
swankjesse: fail

retrofit/src/test/java/retrofit/RequestBuilderTest.java
```diff
@@ -864,6 +864,147 @@
...
+  @Test public void getUrlAndUrlParamThrows() {
+    class Example {
+      @GET("foo/bar")
+      Call<Object> method(@Url Object url) {
+        return null;
+      }
+    }
+
+    try {
+      buildRequest(Example.class, "foo/bar");
```
swankjesse: fail

retrofit/src/test/java/retrofit/RequestBuilderTest.java
```diff
@@ -864,6 +864,147 @@
...
+  @Test public void getWithoutUrlThrows() {
+    class Example {
+      @GET
+      Call<Object> method() {
+        return null;
+      }
+    }
+
+    try {
+      buildRequest(Example.class);
```
swankjesse: fail

retrofit/src/test/java/retrofit/RequestBuilderTest.java
```diff
@@ -864,6 +864,147 @@
...
+  @Test public void getWithUrlThenPathThrows() {
+    class Example {
+      @GET
+      Call<Object> method(@Url String url, @Path("hey") String hey) {
+        return null;
+      }
+    }
+
+    try {
+      buildRequest(Example.class, "foo/bar");
```
swankjesse: fail

retrofit/src/test/java/retrofit/RequestBuilderTest.java
```diff
@@ -864,6 +864,147 @@
...
+    } catch (IllegalArgumentException e) {
+      assertThat(e).hasMessage("@Path parameters may not be used with @Url. (parameter #2)\n"
+          + "    for method Example.method");
+    }
+  }
+
+  @Test public void getWithPathThenUrlThrows() {
```
swankjesse: Nice.

retrofit/src/test/java/retrofit/RequestBuilderTest.java
```diff
@@ -864,6 +864,147 @@
...
+    class Example {
+      @GET
+      Call<Object> method(@Path("hey") String hey, @Url Object url) {
+        return null;
+      }
+    }
+
+    try {
+      buildRequest(Example.class, "foo/bar");
```
swankjesse: fial

retrofit/src/test/java/retrofit/RequestBuilderTest.java
```diff
@@ -864,6 +864,147 @@
...
+  @Test public void getWithQueryThenUrlThrows() {
+    class Example {
+      @GET("foo/bar")
+      Call<Object> method(@Query("hey") String hey, @Url Object url) {
+        return null;
+      }
+    }
+
+    try {
+      buildRequest(Example.class, "hey", "foo/bar/");
```
swankjesse: fail

retrofit/src/test/java/retrofit/RequestBuilderTest.java
```diff
@@ -864,6 +864,147 @@
...
+
+    try {
+      buildRequest(Example.class, "foo/bar");
+    } catch (IllegalArgumentException e) {
+      assertThat(e).hasMessage("@Path can only be used with relative url on @GET (parameter #1)\n"
+          + "    for method Example.method");
+    }
+  }
+
+  @Test public void getWithQueryThenUrlThrows() {
```
swankjesse: test UrlThenQuery ?

swankjesse:
LGTM

PR #929, opened 2015-06-29
Also refer to the latest OkHttp version.

swankjesse:


website/index.html
```diff
@@ -338,6 +338,11 @@ <h3 id="license">License</h3>
           $('.version-tag').text('v' + version);
           $('.version-href').attr('href', url);
         });
+        $.fn.artifactVersion('com.squareup.okhttp', 'okhttp', function(version, url) {
+          $('.okhttpversion').text(version);
+          $('.okhttpversion-tag').text('v' + version);
+          $('.okhttpversion-href').attr('href', url);
```
JakeWharton: These last two lines aren't needed as they don't do anything.

JakeWharton:
LGTM

website/index.html
swankjesse: Fixed.

PR #930, opened 2015-06-29
Additional coverage for Url annotation

benoitdion:
Assert Url works with the POST annotation and with a different host.

### DIFF
--- a/retrofit/src/test/java/retrofit/RequestBuilderTest.java
+++ b/retrofit/src/test/java/retrofit/RequestBuilderTest.java
@@ -882,7 +882,7 @@ public final class RequestBuilderTest {
 
   @Test public void getAbsoluteUrl() {
     class Example {
-      @GET("http://example.org/foo/bar/")
+      @GET("http://example2.com/foo/bar/")
       Call<Object> method() {
         return null;
       }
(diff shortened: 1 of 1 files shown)


retrofit/src/test/java/retrofit/RequestBuilderTest.java
```diff
@@ -910,6 +910,21 @@
     assertThat(request.body()).isNull();
   }
 
+  @Test public void getWithUrlAbsoluteDifferentHost() {
```
JakeWharton: This is already tested directly above. The base URL authority is 'example.com' and we switch it to 'example.org'. I realize they look very similar which can cause confusion. Maybe just change the above two tests to 'example2.com`?

benoitdion: That's fine, didn't see the .org difference. Any value in a test with the same authority?

JakeWharton: Sure. Can't hurt.

benoitdion:
Made additional changes based on review comments

retrofit/src/test/java/retrofit/RequestBuilderTest.java
```diff
@@ -903,10 +903,25 @@
       }
     }
 
-    Request request = buildRequest(Example.class, "http://example.org/foo/bar/");
+    Request request = buildRequest(Example.class, "https://example2.com/foo/bar/");
```
benoitdion: Note the `https` change

PR #938, opened 2015-07-04
Support iterable and array parts.

### LINKED ISSUE #685: Passing array of values which have the same parameter name in Multipart request
I want to send multipart POST request exactly like GET request as below
`param[]=1&param[]=2&param[]=3,....`
currently I think have to construct MultipartTypedOutput to send arbitrary number of values of the same name. For example,
```java
 MultipartTypedOutput body = new MultipartTypedOutput();
 for (String value : values) {
        body.addPart("param[]", new TypedString(value));
  }
```
Please correct me if I am in wrong direction.

So I wonder if it is good for Retrofit to have an annotation for this pattern, something like @PartArray,
and the concept is implemented [here](https://github.com/skonb/retrofit/commit/21a40a85ad6c8177062651f43303541a6ec81bab) WITHOUT any tests...

Usage is,
```java
   @Multipart
   @POST("/")
   Observable sendSomething(@PartArray("params[]") List<String> params, @Part("image") TypedFile image);
```

If it is worth considering, I will write tests for this.
Thank you.

JakeWharton:
Closes #685.

### DIFF
--- a/retrofit/src/main/java/retrofit/RequestBuilderAction.java
+++ b/retrofit/src/main/java/retrofit/RequestBuilderAction.java
@@ -199,8 +199,25 @@ abstract class RequestBuilderAction {
     @Override void perform(RequestBuilder builder, Object value) {
       if (value == null) return; // Skip null values.
 
-      //noinspection unchecked
-      builder.addPart(headers, converter.toBody((T) value));
+      if (value instanceof Iterable) {
+        for (Object iterableValue : (Iterable<?>) value) {
+          if (iterableValue != null) { // Skip null values.
+            //noinspection unchecked
+            builder.addPart(headers, converter.toBody((T) iterableValue));
+          }
+        }
+      } else if (value.getClass().isArray()) {
+        for (int x = 0, arrayLength = Array.getLength(value); x < arrayLength; x++) {
+          Object arrayValue = Array.get(value, x);
+          if (arrayValue != null) { // Skip null values.
+            //noinspection unchecked
+            builder.addPart(headers, converter.toBody((T) arrayValue));
+          }
+        }
+      } else {
+        //noinspection unchecked
+        builder.addPart(headers, converter.toBody((T) value));
+      }
     }
   }
 
--- a/retrofit/src/test/java/retrofit/RequestBuilderTest.java
+++ b/retrofit/src/test/java/retrofit/RequestBuilderTest.java
@@ -1336,6 +1336,58 @@ public final class RequestBuilderTest {
         .contains("\r\npong\r\n--");
   }
 
+  @Test public void multipartPartArray() throws IOException {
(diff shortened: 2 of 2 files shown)


retrofit/src/main/java/retrofit/RequestBuilderAction.java
```diff
@@ -199,8 +199,25 @@
     @Override void perform(RequestBuilder builder, Object value) {
       if (value == null) return; // Skip null values.
 
-      //noinspection unchecked
-      builder.addPart(headers, converter.toBody((T) value));
+      if (value instanceof Iterable) {
```
JakeWharton: These behaviors can be front loaded to configuration time. Another PR.

retrofit/src/main/java/retrofit/RequestBuilderAction.java
```diff
@@ -199,8 +199,25 @@
...
+          if (iterableValue != null) { // Skip null values.
+            //noinspection unchecked
+            builder.addPart(headers, converter.toBody((T) iterableValue));
+          }
+        }
+      } else if (value.getClass().isArray()) {
+        for (int x = 0, arrayLength = Array.getLength(value); x < arrayLength; x++) {
+          Object arrayValue = Array.get(value, x);
+          if (arrayValue != null) { // Skip null values.
+            //noinspection unchecked
```
swankjesse: The casts are scary!

retrofit/src/test/java/retrofit/RequestBuilderTest.java
```diff
@@ -1336,6 +1336,58 @@
...
+
+    RequestBody body = request.body();
+    Buffer buffer = new Buffer();
+    body.writeTo(buffer);
+    String bodyString = buffer.readUtf8();
+
+    assertThat(bodyString)
+        .contains("name=\"ping\"")
+        .contains("pong")
+        .contains("name=\"ping\"")
```
swankjesse: does this do what you want? ie. would it work if ping was included only once?

swankjesse:
LGTM

retrofit/src/test/java/retrofit/RequestBuilderTest.java
JakeWharton: Nope. It's copy/paste. I'll come up with something better tomorrow.

retrofit/src/main/java/retrofit/RequestBuilderAction.java
JakeWharton: Yeah I don't know why these work because I feel like they shouldn't. I'll break apart these behaviors into individual actions so validation happens at configuration time.

JakeWharton:
This is lazy. Going to do the whole iterable / array thing properly for all of the annotations.

PR #942, opened 2015-07-05
Use non-timing based methods to detect buffer and stream.

JakeWharton:


retrofit/src/test/java/retrofit/CallTest.java
```diff
@@ -424,18 +424,16 @@
...
-    assertThat(executeTook).isGreaterThan(MILLISECONDS.toNanos(1000));
-
-    long readStart = System.nanoTime();
-    String body = response.body().string();
-    long readTook = System.nanoTime() - readStart;
-    assertThat(readTook).isLessThan(MILLISECONDS.toNanos(500));
-    assertThat(body).isEqualTo("1234");
+    // When buffering we will detect all socket problems before returning the Response.
+    try {
+      example.getBody().execute();
```
swankjesse: fail fail

retrofit/src/test/java/retrofit/CallTest.java
```diff
@@ -445,18 +443,18 @@
...
-
-    long readStart = System.nanoTime();
-    String body = response.body().string();
-    long readTook = System.nanoTime() - readStart;
-    assertThat(readTook).isGreaterThan(MILLISECONDS.toNanos(1000));
-    assertThat(body).isEqualTo("1234");
+
+    // When streaming we only detect socket problems as the ResponseBody is read.
+    try {
+      response.body().string();
```
swankjesse: fail fail

retrofit/src/test/java/retrofit/CallTest.java
swankjesse: Maybe extract `getBody()` out of the try/catch to make it clear that this only happens on `execute()` ?

swankjesse:
LGTM

retrofit/src/test/java/retrofit/CallTest.java
JakeWharton: Damn. I need to make an IntelliJ template for this pattern so I stop forgetting.

swankjesse: At one point I checked out the IntelliJ source code with the hope of making `catch (Exception expected)` with the name _expected_ trigger some static analysis. Maybe I should just file a feature request.

swankjesse: https://youtrack.jetbrains.com/issue/IDEA-142295

PR #944, opened 2015-07-05
Restore mock module for creating delayed and erroring Calls.

### LINKED ISSUE #584: Build fails on windows at MockRestAdapterTest:syncHttpExceptionBecomesError, but only sometimes
On a fresh clone and ```mvn install``` of the project I got this failure, but only on the first two times that I tried to install. After that it installed fine. Is this a problem?
```
java.lang.AssertionError:
Expecting:
 <99L>
to be greater than or equal to:
 <100L>
```

[at this line](https://github.com/square/retrofit/blob/43b7ea14e5aca1d710deccb95b79484b03e99bb9/retrofit-mock/src/test/java/retrofit/MockRestAdapterTest.java#L391)

```
assertThat(tookMs).isGreaterThanOrEqualTo(100);
```

JakeWharton:
Closes #584.


retrofit-mock/src/main/java/retrofit/MockCall.java
```diff
@@ -0,0 +1,182 @@
...
+      executed = true;
+    }
+    task = backgroundExecutor.submit(new Runnable() {
+      private void callFailure(final Throwable throwable) {
+        callbackExecutor.execute(new Runnable() {
+          @Override public void run() {
+            try {
+              callback.failure(throwable);
+            } catch (Throwable t) {
+              t.printStackTrace();
```
swankjesse: Lame.

retrofit-mock/src/main/java/retrofit/MockCall.java
```diff
@@ -0,0 +1,182 @@
...
+        });
+      }
+
+      private void callSuccess(final Response<T> response) {
+        callbackExecutor.execute(new Runnable() {
+          @Override public void run() {
+            try {
+              callback.success(response);
+            } catch (Throwable t) {
+              t.printStackTrace();
```
swankjesse: Same.

retrofit-mock/src/main/java/retrofit/MockCall.java
```diff
@@ -0,0 +1,182 @@
...
+        interrupted = true;
+      }
+      duration = start + ms - uptimeMillis();
+    }
+
+    if (interrupted) {
+      // Important: we don't want to quietly eat an interrupt() event,
+      // so we make sure to re-interrupt the thread so that the next
+      // call to Thread.sleep() or Object.wait() will be interrupted.
+      Thread.currentThread().interrupt();
```
swankjesse: This is all sad.

retrofit-mock/src/main/java/retrofit/MockRetrofit.java
```diff
@@ -0,0 +1,182 @@
...
+ * public class MockUserService implements UserService {
+ *   private final MockRetrofit mock = // ...
+ *
+ *   &#64;Override public Call&lt;User> getUser(String userId) {
+ *     return mock.newSuccessCall(new User("Jake"));
+ *   }
+ * }
+ * </pre>
+ */
+public final class MockRetrofit {
```
swankjesse: Not strictly a mock, right? More like a Fake?

retrofit-mock/src/main/java/retrofit/MockCall.java
JakeWharton: What else should happen? OkHttp and Retrofit core do the same thing. Uncaught exception handler?

retrofit-mock/src/main/java/retrofit/MockRetrofit.java
JakeWharton: Yeah I want to rename all the things to "fake" or something. Mock stuck because it was originally yet incorrectly called "mock mode" in our apps.

retrofit-mock/src/main/java/retrofit/MockRetrofit.java
```diff
@@ -0,0 +1,182 @@
...
+  public void setErrorPercentage(int errorPct) {
+    if (errorPct < 0 || errorPct > 100) {
+      throw new IllegalArgumentException("Error percentage must be between 0 and 100.");
+    }
+    this.errorPct = errorPct;
+  }
+
+  /** The percentage of calls to {@link #calculateIsFailure()} that return {@code true}. */
+  public int errorPercentage() {
+    return errorPct;
```
swankjesse: nit: why `Pct` and not `Percentage` ? (I just don't like doing the mapping in my head)

retrofit-mock/src/main/java/retrofit/MockRetrofit.java
```diff
@@ -0,0 +1,182 @@
...
+
+  /**
+   * Randomly determine whether this call should result in a network failure.
+   * <p>
+   * This method is exposed for implementing other, non-Retrofit services which exhibit similar
+   * network behavior. Retrofit services automatically will exhibit network behavior when wrapped
+   * using {@link #newSuccessCall} or {@link #newFailureCall}.
+   */
+  public boolean calculateIsFailure() {
+    int randomValue = random.nextInt(100) + 1;
```
swankjesse: drop the `+ 1` here, and use `<` instead of `<=` ?

retrofit-mock/src/main/java/retrofit/MockRetrofit.java
swankjesse: (`Percent` is probably better than `Percentage` anyway, both for field name and method name)

retrofit-mock/src/main/java/retrofit/MockRetrofit.java
```diff
@@ -0,0 +1,182 @@
...
+    this.variancePct = variancePct;
+  }
+
+  /** The plus-or-minus variance percentage of the network round trip delay. */
+  public int variancePercentage() {
+    return variancePct;
+  }
+
+  /** Set the percentage of calls to {@link #calculateIsFailure()} that return {@code true}. */
+  public void setErrorPercentage(int errorPct) {
```
swankjesse: Using an int between 0 and 100 instead of a double between 0 and 1 is interesting. Are there other APIs we're consistent with?

retrofit-mock/src/test/java/retrofit/MockRetrofitTest.java
```diff
@@ -0,0 +1,327 @@
...
+      int delay = mockRetrofit.calculateDelayForCall();
+      if (delay > upperBound) {
+        upperBound = delay;
+      }
+      if (delay < lowerBound) {
+        lowerBound = delay;
+      }
+    }
+    assertThat(upperBound).isEqualTo(2799); // ~40% above 2000
+    assertThat(lowerBound).isEqualTo(1200); // ~40% below 2000
```
swankjesse: Nice.

retrofit-mock/src/test/java/retrofit/MockRetrofitTest.java
```diff
@@ -0,0 +1,327 @@
...
+    mockRetrofit.setErrorPercentage(0);
+
+    Call<String> call = mockRetrofit.newSuccessCall("Hi");
+
+    long startNanos = System.nanoTime();
+    Response<String> response = call.execute();
+    long tookMs = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - startNanos);
+
+    assertThat(response.body()).isEqualTo("Hi");
+    assertThat(tookMs).isGreaterThanOrEqualTo(100);
```
swankjesse: Nice tests.

retrofit-mock/src/test/java/retrofit/MockRetrofitTest.java
```diff
@@ -0,0 +1,327 @@
...
+      }
+
+      @Override public void failure(Throwable t) {
+        latch.countDown();
+      }
+    });
+
+    call.cancel();
+
+    assertFalse(latch.await(1, SECONDS));
```
swankjesse: In OkHttp cancel triggers the failure callback. I think we might want retrofit to do likewise.

retrofit-mock/src/main/java/retrofit/MockCall.java
swankjesse: Yeah. Uncaught exception handler, which will crash the app, which is I think what we want.

swankjesse: (Actually, OkHttp logs but only for IOExceptions. Other failures like NPEs crash the app.)

swankjesse:
LGTM

Very nice.

retrofit-mock/src/test/java/retrofit/MockRetrofitTest.java
JakeWharton: Ah, ok. Probably good to notify.

retrofit-mock/src/main/java/retrofit/MockRetrofit.java
JakeWharton: Not that I know of. I chose an integer for a few reasons:
 * You're unlikely to want higher granularity than whole percents.
 * Reading and writing to somewhere persisted is easy.
 * It's easier to present in UI, especially as whole numbers (which we do).

It's not a range of 100, but the individual channels of colors are usually presented as integers rather than floats. Only example I can think of...

swankjesse: Yeah. The only example I came up with after racking my brain was loadFactor on HashMap.

The only way I'd imagine a desire for more granularity is to fail 1 in 1000 requests. Say, for an app that makes a lot of requests.

retrofit-mock/src/main/java/retrofit/MockCall.java
JakeWharton: Blame Android. I copy/pasta.

retrofit-mock/src/main/java/retrofit/MockRetrofit.java
JakeWharton: DOne

retrofit-mock/src/main/java/retrofit/MockRetrofit.java
JakeWharton: Yep. Done

retrofit-mock/src/main/java/retrofit/MockCall.java
JakeWharton: Done

retrofit-mock/src/main/java/retrofit/MockCall.java
JakeWharton: Same

retrofit-mock/src/main/java/retrofit/MockRetrofit.java
hastebrot: Is the module `retrofit-mock` also going to be renamed to `retrofit-fake`?

hastebrot: Also: Snapshots for `retrofit-mock` and the samples are out-dated.

- https://oss.sonatype.org/content/repositories/snapshots/com/squareup/retrofit/retrofit-mock/2.0.0-SNAPSHOT/
- https://oss.sonatype.org/content/repositories/snapshots/com/squareup/retrofit/samples/

PR #948, opened 2015-07-09
Call.cancel at any time.

swankjesse:


swankjesse:
(this is consistent with OkHttp)

baoti:
I like this.

felipecsl:
good stuff :+1:

retrofit/src/main/java/retrofit/OkHttpCall.java
```diff
@@ -144,11 +155,11 @@ private void callSuccess(Response<T> response) {
     }
   }
 
-  public void cancel() {
+  public synchronized void cancel() {
```
JakeWharton: I hate that the use of this ends up in the docs.

retrofit/src/main/java/retrofit/OkHttpCall.java
```diff
@@ -30,6 +30,7 @@
 
   private volatile com.squareup.okhttp.Call rawCall;
   private boolean executed; // Guarded by this.
+  private boolean canceled; // Guarded by this.
```
JakeWharton: Can this just be a volatile? We never read and write in one place. One person just writes and another just reads.

retrofit/src/main/java/retrofit/OkHttpCall.java
```diff
@@ -30,6 +30,7 @@
 
   private volatile com.squareup.okhttp.Call rawCall;
   private boolean executed; // Guarded by this.
```
JakeWharton: Which makes me think this could use an AtomicBooleanFieldUpdater for lock-free CAS changes

JakeWharton:
LGTM. Nice tests.

retrofit/src/main/java/retrofit/OkHttpCall.java
swankjesse: You mean `synchronized`? It doesn't

retrofit/src/main/java/retrofit/OkHttpCall.java
swankjesse: yup, though I prefer `synchronized` 'cause it's easier to reason about

PR #950, opened 2015-07-09
Rename Callback.success() to onSuccess(), and failure() to onFailure().

swankjesse:


retrofit/src/main/java/retrofit/DefaultCallAdapter.java
```diff
@@ -0,0 +1,53 @@
...
+package retrofit;
+
+import java.lang.reflect.Type;
+
+/**
+ * A call adapter that uses the same thread for both I/O and application-level callbacks. For
+ * synchronous calls this is the application thread making the request; for asynchronous calls this
+ * is a thread provided by OkHttp's dispatcher.
+ */
+final class DefaultCallAdapter<T> implements CallAdapter<T> {
```
swankjesse: Extracted from `NothingCallAdapter`

retrofit/src/main/java/retrofit/Callback.java
```diff
@@ -30,8 +30,8 @@
  */
 public interface Callback<T> {
   /** Successful HTTP response. */
-  void success(Response<T> response);
+  void onSuccess(Response<T> response);
```
JakeWharton: onResponse to match OkHttp? Success might imply 2xx only and this gets called for all responses.

swankjesse: Can do.

JakeWharton:
LGTM but needs rebased off master? The PR build (post-merge) is failing.

PR #976, opened 2015-07-23
Shift mock module to allow applying behavior to any invocation type.

JakeWharton:
This supports both Call and the RxJava Observable.


retrofit-adapters/rxjava-mock/src/main/java/retrofit/mock/ObservableBehaviorAdapter.java
```diff
@@ -0,0 +1,44 @@
...
+
+import rx.Observable;
+import rx.functions.Func1;
+
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+
+public final class ObservableBehaviorAdapter implements BehaviorAdapter<Observable<?>> {
+  private static final ObservableBehaviorAdapter INSTANCE = new ObservableBehaviorAdapter();
+
+  public static ObservableBehaviorAdapter create() {
```
swankjesse: Is this an pattern? Weird for create to not create.

retrofit-adapters/rxjava-mock/src/main/java/retrofit/mock/ObservableBehaviorAdapter.java
```diff
@@ -0,0 +1,44 @@
...
+    return INSTANCE;
+  }
+
+  private ObservableBehaviorAdapter() {
+  }
+
+  @Override public Observable<?> applyBehavior(final Behavior behavior, final Observable<?> value) {
+    return Observable.timer(behavior.calculateDelay(MILLISECONDS), MILLISECONDS)
+        .flatMap(new Func1<Long, Observable<?>>() {
+          @Override public Observable<?> call(Long aLong) {
```
swankjesse: aLong ?

retrofit-mock/src/main/java/retrofit/mock/Behavior.java
```diff
@@ -0,0 +1,122 @@
...
+
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+
+public final class Behavior {
+  private static final int DEFAULT_DELAY_MS = 2000; // Network calls will take 2 seconds.
+  private static final int DEFAULT_VARIANCE_PERCENTT = 40; // Network delay varies by ±40%.
+  private static final int DEFAULT_FAILURE_PERCENTT = 3; // 3% of network calls will fail.
+
+  public static Behavior create() {
+    return new Behavior(new Random());
```
swankjesse: Maybe seed by default?

retrofit-mock/src/main/java/retrofit/mock/Behavior.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  public static Behavior create(Random random) {
+    return new Behavior(random);
+  }
+
+  private final Random random;
+
+  private volatile long delayMs = DEFAULT_DELAY_MS;
+  private volatile int variancePercent = DEFAULT_VARIANCE_PERCENTT;
```
swankjesse: PERCENTT ?

retrofit-mock/src/main/java/retrofit/mock/Behavior.java
```diff
@@ -0,0 +1,122 @@
...
+    this.delayMs = unit.toMillis(amount);
+  }
+
+  /** The network round trip delay. */
+  public long delay(TimeUnit unit) {
+    return MILLISECONDS.convert(delayMs, unit);
+  }
+
+  /** Set the plus-or-minus variance percentage of the network round trip delay. */
+  public void setVariancePercent(int variancePercentt) {
```
swankjesse: what's with the extra t?

retrofit-mock/src/main/java/retrofit/mock/Behavior.java
```diff
@@ -0,0 +1,122 @@
...
+  /**
+   * Get the delay that should be used for delaying a response in accordance with configured
+   * behavior.
+   */
+  public long calculateDelay(TimeUnit unit) {
+    float delta = variancePercent / 100f; // e.g., 20 / 100f == 0.2f
+    float lowerBound = 1f - delta; // 0.2f --> 0.8f
+    float upperBound = 1f + delta; // 0.2f --> 1.2f
+    float bound = upperBound - lowerBound; // 1.2f - 0.8f == 0.4f
+    float delayPercent = (random.nextFloat() * bound) + lowerBound; // 0.8 + (rnd * 0.4)
```
swankjesse: nit: your example comment has the + arguments flipped

retrofit-mock/src/main/java/retrofit/mock/BehaviorCall.java
```diff
@@ -0,0 +1,183 @@
...
+  @Override public void enqueue(final Callback<T> callback) {
+    synchronized (this) {
+      if (executed) throw new IllegalStateException("Already executed");
+      executed = true;
+    }
+    task = backgroundExecutor.submit(new Runnable() {
+      @Override public void run() {
+        if (behavior.calculateIsFailure()) {
+          try {
+            Thread.sleep(behavior.calculateDelay(MILLISECONDS));
```
swankjesse: If the delay is zero, bad things happen here

retrofit-mock/src/main/java/retrofit/mock/BehaviorCall.java
```diff
@@ -0,0 +1,183 @@
...
+              }
+              callbackExecutor.execute(new Runnable() {
+                @Override public void run() {
+                  callback.onResponse(response);
+                }
+              });
+            }
+
+            @Override public void onFailure(final Throwable t) {
+              long sleepMs = behavior.calculateDelay(MILLISECONDS);
```
swankjesse: maybe you want helper methods, `sleepAndCallOnFailure()`, `sleepAndCallOnResponse()` ?

retrofit-mock/src/main/java/retrofit/mock/BehaviorCall.java
```diff
@@ -0,0 +1,183 @@
...
+      executed = true;
+    }
+    if (canceled) {
+      throw new InterruptedIOException("canceled");
+    }
+
+    Future<Response<T>> task = backgroundExecutor.submit(new Callable<Response<T>>() {
+      @Override public Response<T> call() throws Exception {
+        try {
+          if (behavior.calculateIsFailure()) {
```
swankjesse: Aren't we missing a sleep in this case?

retrofit-mock/src/main/java/retrofit/mock/BehaviorCall.java
```diff
@@ -0,0 +1,183 @@
...
+            throw behavior.failureException();
+          }
+          Response<T> response = delegate.execute();
+          Thread.sleep(behavior.calculateDelay(MILLISECONDS));
+          return response;
+        } catch (IOException | RuntimeException e) {
+          Thread.sleep(behavior.calculateDelay(MILLISECONDS));
```
swankjesse: .... ahh, here it is

retrofit-mock/src/main/java/retrofit/mock/BehaviorCall.java
```diff
@@ -0,0 +1,183 @@
...
+  @Override public Response<T> execute() throws IOException {
+    synchronized (this) {
+      if (executed) throw new IllegalStateException("Already executed");
+      executed = true;
+    }
+    if (canceled) {
+      throw new InterruptedIOException("canceled");
+    }
+
+    Future<Response<T>> task = backgroundExecutor.submit(new Callable<Response<T>>() {
```
swankjesse: why split declaration and assignment?

swankjesse: ...avoid a race?

swankjesse: What's the utility of this background executor? Maybe add a comment?

swankjesse: If you're going to burn a thread anyway, I'd be tempted to have this call through the async path instead.

retrofit-mock/src/main/java/retrofit/mock/BehaviorCall.java
```diff
@@ -0,0 +1,183 @@
...
+      throw new InterruptedIOException("canceled");
+    } catch (ExecutionException e) {
+      Throwable cause = e.getCause();
+      if (cause instanceof RuntimeException) {
+        throw (RuntimeException) cause;
+      }
+      if (cause instanceof IOException) {
+        throw (IOException) cause;
+      }
+      throw new IllegalStateException(cause);
```
swankjesse: nit: not completely semantic; maybe you want a `RuntimeException`

retrofit-mock/src/main/java/retrofit/mock/BehaviorCall.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+  }
+
+  @Override public void cancel() {
+    canceled = true;
+    Future<?> task = this.task;
+    if (task != null) {
+      task.cancel(true);
```
swankjesse: ahhh, here's the motivation for the background task. I think there's less repetition if sync calls through async

retrofit-mock/src/main/java/retrofit/mock/Calls.java
```diff
@@ -0,0 +1,72 @@
...
+
+import java.io.IOException;
+import retrofit.Call;
+import retrofit.Callback;
+import retrofit.Response;
+
+/** Factory methods for creating {@link Call} instances which immediately respond or fail. */
+public final class Calls {
+  public static <T> Call<T> response(T successValue) {
+    return response(Response.fakeSuccess(successValue));
```
swankjesse: Do you really need the word `fake` here? I'm imagining other cases where it could be useful?

retrofit-mock/src/main/java/retrofit/mock/Calls.java
```diff
@@ -0,0 +1,72 @@
...
+      @Override public void enqueue(Callback<T> callback) {
+        callback.onResponse(response);
+      }
+
+      @Override public void cancel() {
+      }
+
+      @SuppressWarnings("CloneDoesntCallSuperClone") // Immutable object.
+      @Override public Call<T> clone() {
+        return response(response);
```
swankjesse: return this ?

retrofit-mock/src/main/java/retrofit/mock/MockRetrofit.java
```diff
@@ -0,0 +1,46 @@
...
+  @SuppressWarnings("unchecked") // Single-interface proxy creation guarded by parameter safety.
+  public <T> T create(Class<T> service, final T instance) {
+    return (T) Proxy.newProxyInstance(service.getClassLoader(), new Class[] { service },
+        new InvocationHandler() {
+          @Override public Object invoke(Object proxy, Method method, Object[] args)
+              throws Throwable {
+            if (method.getDeclaringClass() == Object.class) {
+              return method.invoke(this, args);
+            }
+            method.setAccessible(true); // Just In Case™
```
swankjesse: What a great API.

swankjesse:
LGTM, though I'm happy to take another look if you'd like another round of review.

One question: why can't the Rx behavior adapter delegate to the Call behavior adapter? Seems like all of the behavior stuff should be able to live at the Call layer.

JakeWharton:
> why can't the Rx behavior adapter delegate to the Call behavior adapter?

Because there is no `Call` here. The mock class methods return an `Observable` on which we must apply the delay before returning it as the same type through the service interface to the caller.

swankjesse:
Ahhh.... because there's no bytes conversion! Makes sense.

LGTM.

retrofit-mock/src/main/java/retrofit/mock/Behavior.java
JakeWharton: I'd prefer to keep it truly random as we want it to exercise as much unpredictability as the real world by default. You cannot rely on mobile networks to behave just like you cannot rely on getting the same behavior over and over.

Unless you want to, of course, which is why there's the overload which lets you seed (like for tests).

hastebrot:
This is great! :+1:

PR #979, opened 2015-07-24
Use rx.Single instead of rx.Observable

runningcode:
This simplifies the API in my opinion. All of the network calls have
either a single success or an error so a Single works perfecly.
It is also easy to convert back to an Observable by calling toObservable.


JakeWharton:
Single is not stable API and therefore its inclusion will be 100% optional
and in a separate class that's only loaded when detected as the return type
to prevent classloading issues should its API change.

PR #981, opened 2015-07-25
Add optional support for RxJava's experimental Single type.

### LINKED ISSUE #969: Support rx.Single from RxJava?
With `rx.Single` released as experimented type of `Observable` do you see the value in supporting it as a return type for network requests?
https://github.com/ReactiveX/RxJava/blob/1.x/CHANGES.md#rxsingle
https://github.com/ReactiveX/RxJava/issues/1594#issuecomment-101300655

Quote from there:
```
A Single will always behave in one of 3 ways:

1) respond with an error
2) never respond
3) respond with a success

An Observable on the other hand may:

1) response with an error
2) never respond
3) respond successfully with no data and terminate
4) respond successfully with a single value and terminate
5) respond successfully with multiple values and terminate
7) respond successfully with one or more values and never terminate (waiting for more data)
```

JakeWharton:
Because this API is not stable the interactions are isolated to a separate class which is only loaded when the Single type is detected as being used.

Closes #979. Closes #969.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,7 +53,7 @@
     <okhttp.version>2.5.0-SNAPSHOT</okhttp.version>
 
     <!-- Adapter Dependencies -->
-    <rxjava.version>1.0.10</rxjava.version>
+    <rxjava.version>1.0.13</rxjava.version>
 
     <!-- Converter Dependencies -->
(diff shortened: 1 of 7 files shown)


swankjesse:
LGTM

PR #988, opened 2015-07-31
LoganSquare Converter

aurae:
A converter implementation using [LoganSquare](https://github.com/bluelinelabs/LoganSquare) as its serialization backbone.


retrofit-converters/logansquare/src/main/java/retrofit/LoganSquareConverter.java
```diff
@@ -0,0 +1,46 @@
...
+import com.squareup.okhttp.ResponseBody;
+
+import java.io.IOException;
+
+/**
+ * A {@linkplain Converter converter} which uses LoganSquare as its conversion backbone.
+ *
+ * @author Marcel Schnelle (aurae)
+ * @see <a>https://github.com/bluelinelabs/LoganSquare</a>
+ */
```
JakeWharton: Remove this whole comment. This class isn't public.

retrofit-converters/logansquare/src/main/java/retrofit/LoganSquareConverter.java
```diff
@@ -0,0 +1,46 @@
...
+final class LoganSquareConverter<T> implements Converter<T> {
+
+	/** Media Type for serialized request bodies */
+	private static final MediaType MEDIA_TYPE = MediaType.parse("application/json; charset=UTF-8");
```
JakeWharton: Indent with two spaces

retrofit-converters/logansquare/src/main/java/retrofit/LoganSquareConverter.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * A {@linkplain Converter converter} which uses LoganSquare as its conversion backbone.
+ *
+ * @author Marcel Schnelle (aurae)
+ * @see <a>https://github.com/bluelinelabs/LoganSquare</a>
+ */
+final class LoganSquareConverter<T> implements Converter<T> {
+
+	/** Media Type for serialized request bodies */
```
JakeWharton: Delete comment

retrofit-converters/logansquare/src/main/java/retrofit/LoganSquareConverter.java
```diff
@@ -0,0 +1,46 @@
...
+	private static final MediaType MEDIA_TYPE = MediaType.parse("application/json; charset=UTF-8");
+
+	/** LoganSquare JSON mapper for parsing & serialization */
```
JakeWharton: Delete comment

retrofit-converters/logansquare/src/main/java/retrofit/LoganSquareConverter.java
```diff
@@ -0,0 +1,46 @@
...
+	private final JsonMapper<T> mapper;
+
+	/**
+	 * Constructor
+	 *
+	 * @param mapper LoganSquare JSON mapper used to parse & serialize objects
+	 */
```
JakeWharton: Delete comment

retrofit-converters/logansquare/src/main/java/retrofit/LoganSquareConverterFactory.java
```diff
@@ -0,0 +1,21 @@
+package retrofit;
+
+import com.bluelinelabs.logansquare.JsonMapper;
+import com.bluelinelabs.logansquare.LoganSquare;
+
+import java.lang.reflect.Type;
+
+/**
+ * A {@linkplain Converter.Factory converter} which uses LoganSquare as its conversion backbone.
```
JakeWharton: change "as its conversion backbone" to "for JSON" to match others.

retrofit-converters/logansquare/src/main/java/retrofit/LoganSquareConverterFactory.java
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * @author Marcel Schnelle (aurae)
```
JakeWharton: Remove author tag

retrofit-converters/logansquare/src/main/java/retrofit/LoganSquareConverterFactory.java
```diff
@@ -0,0 +1,21 @@
...
+ * @see <a>https://github.com/bluelinelabs/LoganSquare</a>
+ */
+public final class LoganSquareConverterFactory implements Converter.Factory {
+
```
JakeWharton: Add private constructor and static `create()` method which calls it

retrofit-converters/logansquare/src/main/java/retrofit/LoganSquareConverterFactory.java
```diff
@@ -0,0 +1,21 @@
...
+	@Override public Converter<?> get(Type type) {
+		// Obtain the JsonMapper for the type and serve a Converter
+		JsonMapper<?> mapper = LoganSquare.mapperFor((Class<?>) type);
```
JakeWharton: This needs to check if the `type` is not a `Class` and throw an exception with a meaningful message. The implicit `ClassCastException` that's thrown here is unsatisfactory.

Also, does this mean I can't use LoganSquare on a `List` of objects? That's really strange...

retrofit-converters/logansquare/src/test/java/retrofit/LoganSquareConverterTest.java
```diff
@@ -0,0 +1,152 @@
...
+import retrofit.http.POST;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests for the {@linkplain LoganSquareConverter}.
+ *
+ * @author Marcel Schnelle (aurae)
+ * @see <a>https://github.com/bluelinelabs/LoganSquare</a>
+ */
```
JakeWharton: Delete whole comment

retrofit-converters/logansquare/src/test/java/retrofit/LoganSquareConverterTest.java
```diff
@@ -0,0 +1,152 @@
...
+public class LoganSquareConverterTest {
```
JakeWharton: Add `final`

retrofit-converters/logansquare/src/test/java/retrofit/LoganSquareConverterTest.java
```diff
@@ -0,0 +1,152 @@
...
+		assertThat(responseBody.getCustomType()).isEqualTo(CustomType.VAL_2);
+		// 4) Excluded field
+		assertThat(responseBody.getNotSerialized()).isNull();
+
+		// Check request body and the received header
+		RecordedRequest request = mockWebServer.takeRequest();
+		assertThat(request.getBody().readUtf8()).isEqualTo("{\"customType\":2,\"name\":\"LOGAN SQUARE IS COOL\",\"list\":[\"value1\",\"value2\"]}");
+		assertThat(request.getHeader("Content-Type")).isEqualTo("application/json; charset=UTF-8");
+	}
+}
```
JakeWharton: Needs a test verifying that deserializing a `List<Foo>` throws an appropriate error message.

retrofit-converters/logansquare/src/main/java/retrofit/LoganSquareConverter.java
JakeWharton: (that applies to all these files)

retrofit-converters/logansquare/README.md
```diff
@@ -0,0 +1,8 @@
+LoganSquare Converter
+=================
+
+A `Converter` which uses [LoganSquare][1] for serialization to and from JSON.
+
+Each type is assigned its corresponding `JsonMapper`, then parsed and serialized.
```
JakeWharton: The second paragraph of other converters focuses on details you need to know for configuring. Since this library has no runtime configuration, you end up describing implementation details of the converter here.

Maybe change it to "The `JsonMapper` instances generated by the compiler are used for conversion."

retrofit-converters/logansquare/src/main/java/retrofit/LoganSquareConverterFactory.java
aurae: You absolutely can use it on Lists (and String-keyed Maps, for that matter). I'm not sure what I was thinking here. I will update the PR with List and Map support asap.

ffelini:
Guys any luck to get thsi pull requested merged soon? :)

bryli:
+1 would love to have this merged

JakeWharton:
Sorry for the delay here. After some thought, we aren't going to add new first-party converters without a compelling reason. Copying a response I put on another PR:

Retrofit already offers 3 first-party JSON converters. Gson and Jackson are the two leading serializers for which we offer converters, and we include Moshi just because it's a Square library. Similarly, we have two first-party protocol buffer converters, one using the official library and one with Wire because it's a Square library. We have a single XML converter because... well no one really knows.

There are many more eclectic and specialized JSON converters just as there are for other serialization formats. Maintaining a bunch of these is burdensome as we have to track the APIs and changes in 6 external libraries which have nothing to do with Retrofit's core. I can think of at least 5 other JSON libraries that have popped up recently almost always just being built on top of Jackson or Gson.

Retrofit was modified (by me, actually) to be agnostic to converters well before it hit 1.0 a few years ago. The point was that as libraries and formats change in popularity, no changes in Retrofit would be needed to support them. This is very much the case with the forthcoming version 2.0 as well.

As such, we won't be adding any more first-party converters unless a compelling reason can be made as to why we should. Since we already have 3 JSON converters, there is little argument to be made for more.

However, there's no reason that these cannot be turned into third-party libraries that exist on their own. I have [made a wiki page][1] for listing third-party converters so that they are able to be discovered in one place.

Thanks for taking the time to implement this, despite that fact that we aren't able to accept it into the main repo.

 [1]: https://github.com/square/retrofit/wiki/Converters

aurae:
Thanks for your reply, Jake. I understand your point about the abundance of JSON converters, which is why I'm going to convert (heh get it) this into a third-party artifact. I will drop another message in this thread once it's deployed, hoping that it might get appended to the wiki page.

JakeWharton:
You should be able to edit the Wiki page directly. And please do put it there!

aurae:
@ffelini @bryli Version 1.0.0 of the LoganSquare converter has been released [and added to the wiki](https://github.com/square/retrofit/wiki/Converters)! Further maintenance will happen in its dedicated repository from now on.

jfragosoperez:
I understand your thoughts, Jake. Thanks for your reply guys and also for the addition of the link (to the converter as third-party artifact) into the wiki.

PR #1001, opened 2015-08-07
HTTP-specific exception for the simple Observable case.

JakeWharton:
When you specify a return type of Observable<ResponseType>, we need a way to propagate non-2xx failures through the onError pipeline which accepts any Throwable. A custom exception subclassing IOException allows downstream consumers to filter and extract the code, message, and full response.

### DIFF
--- a/retrofit-adapters/rxjava/src/main/java/retrofit/HttpException.java
+++ b/retrofit-adapters/rxjava/src/main/java/retrofit/HttpException.java
@@ -1,0 +1,34 @@
+package retrofit;
+
+import java.io.IOException;
+
+/** Exception for an unexpected, non-2xx HTTP response. */
+public final class HttpException extends IOException {
+  private final int code;
+  private final String message;
+  private final transient Response<?> response;
+
+  public HttpException(Response<?> response) {
+    super("HTTP " + response.code() + " " + response.message());
+    this.code = response.code();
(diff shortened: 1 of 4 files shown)


JakeWharton:
Kicking build. OkHttp's master build broke on a flaky test thus a snapshot wasn't published which corrected the error which failed this build.

/me climbs out of rabbit hole

JakeWharton:
Travis refuses to see the new snapshots.
```
$ mcv -U 2> /dev/null | tail -7
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 19.116 s
[INFO] Finished at: 2015-08-07T15:42:43-04:00
[INFO] Final Memory: 43M/539M
[INFO] ------------------------------------------------------------------------
```

retrofit-adapters/rxjava/src/main/java/retrofit/HttpException.java
```diff
@@ -0,0 +1,34 @@
+package retrofit;
```
swankjesse: nit: copyright

JakeWharton: I need to set this up in checkstyle...

PR #1003, opened 2015-08-10
Switch HttpException to not be an IOException.

JakeWharton:
This allows for easier 'instanceof' checks in an onError to differentiate whether the problem was in the transport or the received response.

### DIFF
--- a/retrofit-adapters/rxjava/src/main/java/retrofit/HttpException.java
+++ b/retrofit-adapters/rxjava/src/main/java/retrofit/HttpException.java
@@ -1,9 +1,7 @@
 package retrofit;
 
-import java.io.IOException;
-
 /** Exception for an unexpected, non-2xx HTTP response. */
-public final class HttpException extends IOException {
+public final class HttpException extends Exception {
   private final int code;
   private final String message;
   private final transient Response<?> response;


JakeWharton:
Saves you from `e instanceof IOException && !(e instanceof HttpException)` in common error handling.

retrofit-adapters/rxjava/src/main/java/retrofit/HttpException.java
```diff
@@ -1,9 +1,7 @@
 package retrofit;
 
-import java.io.IOException;
-
 /** Exception for an unexpected, non-2xx HTTP response. */
```
swankjesse: unexpected –> RuntimeException?

Should we promote this to OkHttp ?

JakeWharton: I don't think it has a reason to. It always delivers non-200s to response callback or returned from execute. The need for this arises only out of a user wanting 200s sent to 'onNext' and everything else needing to go to 'onError'.

swankjesse: Got it.

PR #1006, opened 2015-08-12
Do not proxy default methods on service interfaces.

### LINKED ISSUE #998: Do not modify default interface methods (just ignore them)
Currently the following exception is thrown when default interface methods a used:

    java.lang.IllegalArgumentException: MyService.load: HTTP method annotation is required (e.g., @GET, @POST, etc.).

Found while trying to workaround  #997 

    @GET(..)
    Response load( 
         @Query("a") Integer a,
         @Query("b") Integer b
    );

    default Response simpleLoad(Request r 
         return load(
                  r.getA(),
                  r.getB()
         );
    );

JakeWharton:
Closes #998.

This is not easily testable, and I'm not in the mood to craft Java 8 integration tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: java
 
 jdk:
-  - oraclejdk7
   - oraclejdk8
 
 after_success:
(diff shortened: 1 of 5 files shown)


JakeWharton:
Can't compile with JDK 7 with this change. Not sure how I feel about that.

swankjesse:
LGTM. If you want to require Java 8 for building, we'll want to add animal sniffer.

swankjesse:
(eventually!)

JakeWharton:
Added animal sniffer.

PR #1021, opened 2015-08-22
Retrofit.Builder.addConverter().

swankjesse:
Seems like it would be convenient.

### DIFF
--- a/retrofit/src/main/java/retrofit/Retrofit.java
+++ b/retrofit/src/main/java/retrofit/Retrofit.java
@@ -22,6 +22,7 @@ import com.squareup.okhttp.ResponseBody;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
+import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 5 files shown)


retrofit/src/main/java/retrofit/Retrofit.java
```diff
@@ -205,6 +206,21 @@ public Builder baseUrl(BaseUrl baseUrl) {
...
 
+    /** Add converter for serialization and deserialization of {@code type}. */
+    public <T> Builder addConverter(final Type type, final Converter<T> converter) {
+      checkNotNull(type, "type == null");
+      checkNotNull(converter, "converter == null");
+      converterFactories.add(new Converter.Factory() {
+        @Override public Converter<?> get(Type candidate) {
+          return candidate.equals(type) ? converter : null;
+        }
+        @Override public String toString() {
```
swankjesse: (We don't do a great job of overriding `toString()` throughout Retrofit)

retrofit/src/main/java/retrofit/http/Part.java
```diff
@@ -18,6 +18,7 @@
 import java.lang.annotation.Documented;
 import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
+import retrofit.Converter;
 import retrofit.Retrofit;
```
JakeWharton: Unused import

retrofit/src/main/java/retrofit/http/PartMap.java
```diff
@@ -18,6 +18,7 @@
 import java.lang.annotation.Documented;
 import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
+import retrofit.Converter;
 import retrofit.Retrofit;
```
JakeWharton: Unused import

JakeWharton:
LGTM

PR #1024, opened 2015-08-22
Prefer the word contentType.

swankjesse:
The type is MediaType, but our application of that type is the body
content type.

### DIFF
--- a/retrofit/src/main/java/retrofit/Utils.java
+++ b/retrofit/src/main/java/retrofit/Utils.java
@@ -94,22 +94,22 @@ final class Utils {
     }
 
     BufferedSource source = body.source();
-    final Buffer buffer = new Buffer();
+    Buffer buffer = new Buffer();
     buffer.writeAll(source);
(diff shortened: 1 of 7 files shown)


retrofit/src/main/java/retrofit/Utils.java
```diff
@@ -94,22 +94,22 @@ static ResponseBody readBodyToBytesIfNecessary(final ResponseBody body) throws I
...
     source.close();
 
     return ResponseBody.create(body.contentType(), body.contentLength(), buffer);
   }
 
-  static <T> void validateServiceClass(Class<T> service) {
+  static <T> void validateServiceInterface(Class<T> service) {
     if (!service.isInterface()) {
-      throw new IllegalArgumentException("Only interface baseUrl definitions are supported.");
+      throw new IllegalArgumentException("API declarations must be interfaces.");
```
JakeWharton: Heh, clearly there's no test for this check

PR #1039, opened 2015-08-30
Give method or parameter annotations to CallAdapter and Converter factories.

JakeWharton:
Giant win: we no longer have to special-case the streaming ResponseBody behavior.

### DIFF
--- a/retrofit/src/main/java/retrofit/OkHttpBodyConverterFactory.java
+++ b/retrofit/src/main/java/retrofit/OkHttpBodyConverterFactory.java
@@ -17,10 +17,12 @@ package retrofit;
 
 import com.squareup.okhttp.RequestBody;
 import com.squareup.okhttp.ResponseBody;
+import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
+import retrofit.http.Streaming;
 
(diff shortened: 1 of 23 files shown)


retrofit/src/main/java/retrofit/OkHttpBodyConverterFactory.java
```diff
@@ -28,8 +30,9 @@
     if (RequestBody.class.isAssignableFrom(cls)) {
       return new OkHttpRequestBodyConverter();
     }
-    if (ResponseBody.class.isAssignableFrom(cls)) {
-      return new OkHttpResponseBodyConverter(false);
+    if (cls == ResponseBody.class) {
```
JakeWharton: Slight behavior change here for the better. If you used `MyResponseBody` we would incorrectly match it as valid and then blow up trying to cast a different subclass.

swankjesse:
Array instead of Set is a reasonable optimization, though I don't love the fact that the array is mutable.

I think you want a `NO_ANNOTATIONS` constant instead of null. Nulls lame.

swankjesse:
LGTM.

JakeWharton:
I started with `NO_ANNOTATIONS` and reverted because it was late (and I was lazy). And I agree mutable is sad.

PR #1052, opened 2015-09-03
Switch to a generic to/from converter, explicit factory.

### LINKED ISSUE #1044: Potential new Converter API
Copy/paste of conversation:

> i'm a bit concerned about the Converter/Factory interfaces now that there's multiple

> if you want to add a thing that lets you encode enum's to text/plain strings by calling .name() you also have to implement the deserialize method

> tempted to make:
> ```java
> interface Converter<F, T> {
>   T convert(F value);
> 
>   interface Factory {
>     Converter<ResponseBody, ?> fromBody(Type type, Annotation[] annotations);
>     Converter<?, RequestBody> toBody(Type type, Annotation[] annotations);
>   }
> }
> ```
> 
> actually would probably do an abstract class for `Factory`
> 
> what's interesting is that there's a potential for using that same interface for parameters then
> 
> ```java
> Converter<String, ?> forParam(Type type, Annotation[] annotations);
> ```
> or something...
> 
> to handle the date formatting, enum .name() calling, etc. path/query/header handling

JakeWharton:
This allows for specialization based on for what purpose the converter is being created. For the built-in converter factory, for example, types like `ResponseBody` and `Void` only apply to the response type and types like `RequestBody` only apply to a body/part type.

Closes #1044.


retrofit-converters/gson/src/main/java/retrofit/GsonResponseBodyConverter.java
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+package retrofit;
+
+import com.google.gson.TypeAdapter;
+import com.squareup.okhttp.ResponseBody;
+import java.io.IOException;
+import java.io.Reader;
+
+final class GsonResponseBodyConverter<T> implements Converter<ResponseBody, T> {
```
swankjesse: ultranit: should these just be nested classes in the Factory?

retrofit-converters/jackson/src/main/java/retrofit/JacksonRequestBodyConverter.java
```diff
@@ -0,0 +1,36 @@
...
+  private static final MediaType MEDIA_TYPE = MediaType.parse("application/json; charset=UTF-8");
+
+  private final ObjectWriter adapter;
+
+  JacksonRequestBodyConverter(ObjectWriter adapter) {
+    this.adapter = adapter;
+  }
+
+  @Override public RequestBody convert(T value) throws IOException {
+    byte[] bytes = adapter.writeValueAsBytes(value);
```
swankjesse: non-streaming, not that it matters much

retrofit-converters/moshi/src/main/java/retrofit/MoshiRequestBodyConverter.java
```diff
@@ -18,39 +18,25 @@
...
-      try {
-        source.close();
-      } catch (IOException ignored) {
-      }
-    }
-  }
-
-  @Override public RequestBody toBody(T value) {
+  @Override public RequestBody convert(T value) throws IOException {
     Buffer buffer = new Buffer();
```
swankjesse: Also non-streaming. If I recall correctly this is because `value` might be mutable, right?

swankjesse: (I wonder if that should be called out somewhere)

swankjesse: ... like in docs, about how converters shouldn't retain a reference to the values they've converted

retrofit-converters/gson/src/main/java/retrofit/GsonResponseBodyConverter.java
JakeWharton: They could be. I actually broke them out for the purposes of the stacktrace!

PR #1096, opened 2015-09-19
Use form-data subtype for multipart requests.

### LINKED ISSUE #1092: Multipart request subtype
As discussed in #1063 multipart upload upload still has some issues. For one, it defaults to multipart/mixed, when in [Retrofit 1.9 it used multipart/form-data](https://github.com/square/retrofit/blob/2f4caa6f5df8e7f47a881266bbfc723f537b0b47/retrofit/src/main/java/retrofit/mime/MultipartTypedOutput.java).

I suggest we use the form-data subtype as it seems to be the standard for HTTP file upload. To use it, the following needs to be changed:
- Set Content-Type header of request to `multipart/form-data; boundary=...`
- Set Content-Type header of parts to `multipart/form-data`
- Add `form-data` value to Content-Disposition header and optionally also add the filename parameter

lukaciko:
Use [form-data](http://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.2) content type for multipart requests. Closes #1092.

I don't especially like how this stretches over multiple classes. Perhaps RequestBuilder should be responsible for creating the right part headers.

### DIFF
--- a/retrofit/src/main/java/retrofit/RequestBuilder.java
+++ b/retrofit/src/main/java/retrofit/RequestBuilder.java
@@ -61,6 +61,7 @@ final class RequestBuilder {
     } else if (isMultipart) {
       // Will be set to 'body' in 'build'.
       multipartBuilder = new MultipartBuilder();
+      multipartBuilder.type(MultipartBuilder.FORM);
     }
   }
 
--- a/retrofit/src/main/java/retrofit/RequestBuilderAction.java
+++ b/retrofit/src/main/java/retrofit/RequestBuilderAction.java
@@ -241,7 +241,7 @@ abstract class RequestBuilderAction {
         }
 
         Headers headers = Headers.of(
-            "Content-Disposition", "name=\"" + entryKey + "\"",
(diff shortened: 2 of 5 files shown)


JakeWharton:
Thanks. I'll look soon.

TheGhoul21:
I think that the header for the Part of a file is missing the filename as well. 

```java
Part part = (Part) methodParameterAnnotation;
            com.squareup.okhttp.Headers headers = com.squareup.okhttp.Headers.of(
                "Content-Disposition", "form-data; name=\"" + part.value() + "\";filename=\"file.jpg\"",
                "Content-Transfer-Encoding", part.encoding());```

This fixed at least for me

wKovacs64:
I think e8431de only solved most of #1092. The tail end of the 3rd bullet point was left out:
> and optionally also add the filename parameter

lukaciko:
The filename is required by some backend configurations. In Retrofit 1.9 you could set the filename by creating a `TypedFile` class, but that class doesn't exist anymore. I think it would be best if you open a new issue.

Issue #378, opened 2014-01-14
Request and response logging of the unserialized Java types


lsuski:
Hi,
I'd like to log only request and response body along with http status code and execution time (no headers, urls and methods names). Is there possibility to make log level mechanism more flexible? For example use some interface or simple class with methods like isLogStatus(), isLogBody() etc. and ovveride it for own needs.
Thanks

hosamaly:
It looks like 2.0.0-beta1 removed logging support altogether in commit 40bfa340bb back in January. Any idea when it might return?

JakeWharton:
Nope. Maybe never. Use OkHttp interceptors in the mean time.

hosamaly:
Thanks. FWIW, here is an example for a logging interceptor for OkHttp:
https://github.com/square/okhttp/wiki/Interceptors

vitaliyistomov:
Any chances to have default supported by Square logging interceptor implementation for OkHttp with logging of the bodies of the request and response like it was in Retrofit 1.x? I believe every second person would be interested in having that instead of tweaking the snippet from https://github.com/square/okhttp/wiki/Interceptors to log full request/response for debugging every time.

cemo:
Is this already solved by https://github.com/square/okhttp/pull/1874?
```
OkHttpClient client = new OkHttpClient();
HttpLoggingInterceptor logging = new HttpLoggingInterceptor();
logging.setLevel(Level.BASIC);
client.interceptors().add(logging);
```

JakeWharton:
No. That's logging at the HTTP layer which I mostly don't care about.
Retrofit needs logging at the application layer using your types, not HTTP,
for the representation.

JakeWharton:
Might be time for an event listener

https://publicobject.com/2022/05/01/eventlisteners-are-good/

PR #1098, opened 2015-09-19
Add Logging Mechanism

f2prateek:
This is mostly a copy paste from the v1. I didn't want to pass write code to update `LogLevel` everywhere or make it final, so it's folded into `Logger` for now.

### DIFF
--- a/retrofit/src/main/java/retrofit/OkHttpCall.java
+++ b/retrofit/src/main/java/retrofit/OkHttpCall.java
@@ -15,10 +15,16 @@
  */
 package retrofit;
 
+import com.squareup.okhttp.MediaType;
 import com.squareup.okhttp.OkHttpClient;
 import com.squareup.okhttp.Request;
+import com.squareup.okhttp.RequestBody;
 import com.squareup.okhttp.ResponseBody;
 import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+import okio.Buffer;
+import okio.BufferedSink;
+import okio.BufferedSource;
 
 import static retrofit.Utils.closeQuietly;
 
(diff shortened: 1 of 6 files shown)


retrofit/src/main/java/retrofit/OkHttpCall.java
```diff
@@ -85,9 +95,10 @@ private void callSuccess(Response<T> response) {
       }
 
       @Override public void onResponse(com.squareup.okhttp.Response rawResponse) {
+        long elapsedTime = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start);
```
artem-zinnatullin: `final`?

retrofit/src/main/java/retrofit/Retrofit.java
```diff
@@ -261,6 +265,12 @@ public Builder validateEagerly() {
       return this;
     }
 
+    /** Configure debug logging mechanism. */
+    public Builder setLogger(Logger logger) {
```
artem-zinnatullin: Other methods of the `Builder` don't have `set` prefix, I'd remove it for consistency :)

retrofit/src/main/java/retrofit/OkHttpCall.java
```diff
@@ -109,14 +120,93 @@ private void callSuccess(Response<T> response) {
...
+                    return contentType;
+                  }
+
+                  @Override public void writeTo(BufferedSink sink) throws IOException {
+                    sink.write(buffer, buffer.size());
+                  }
+                })
+                .build();
+          } catch (IOException e) {
+            e.printStackTrace();
```
artem-zinnatullin: Looks suspicious, if we can not read request's body due to `IOException` (bad connection, etc), this exception should be thrown up to the caller, shouldn't it?

artem-zinnatullin:
No tests :crying_cat_face:

retrofit/src/main/java/retrofit/OkHttpCall.java
JakeWharton: never for locals

JakeWharton:
This is not the direction in which we want logging to go.

HTTP-level logging belongs in an interceptor. Whether we provide one in a sibling module with semantics similar to V1 remains to be seen.

As to argument and response value `toString` logging, that's probably something we should provide, but I haven't figured out the interface to support it yet. That can be discussed in #378 which is what is tracking logging in general.

f2prateek:
I considered using interceptors, but the problem I ran into was that it would have to be added first to the list of interceptors, which might not always be possible when people set their own client.

Maybe we could ship a logging interceptor that people have to manually set on the client?

JakeWharton:
Why does it have to be first? If anything I would think you'd want it last to see the actual HTTP data that's going across the wire.

f2prateek:
Interesting, I was thinking that Retrofit should only log the request as it transformed it for OkHttp, and the response as it received from OkHttp before it ran it through the converter, instead of worrying about the work that other interceptors might have done.

By running last, other interceptors will have transformed the original request made by Retrofit into something different by the time it was sent across the wire, or rewritten a response into something else by the time Retrofit receives it.

JakeWharton:
One of the complaints that came up frequently from V1 was the fact that the
logging was a fake representation of the HTTP request and response. It was
a best guess, but often did not represent what was actually sent along the
wire. Providing an interceptor, though, would allow the person to choose.

PR #1101, opened 2015-09-20
Added OPTIONS to the built-in methods

avbk:
Hey guys,

first of all, thanks for the great work. 
We are currently working on a project, which heavily uses the OPTIONS method. I did not read about the @HTTP annotation (just saw it when adding the docs and the test; darn). So I forked the project and added the @OPTIONS annotation. 
Since there is the @HTTP annotation, this might not be that useful, but maybe you still like to add it to your codebase.

Best regards
Alex

### DIFF
--- a/retrofit/src/main/java/retrofit/RequestFactoryParser.java
+++ b/retrofit/src/main/java/retrofit/RequestFactoryParser.java
@@ -37,6 +37,7 @@ import retrofit.http.HTTP;
 import retrofit.http.Header;
 import retrofit.http.Headers;
 import retrofit.http.Multipart;
+import retrofit.http.OPTIONS;
 import retrofit.http.PATCH;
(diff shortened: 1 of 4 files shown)


JakeWharton:
Does your project return a body in the OPTIONS response?

jkwiecien:
We could use @OPTIONS aswell :)

RoyMontoya:
i think it could work fine as optional annotation, but it could also be confusing for the users because its pretty much the same as @HTTP

damianflannery:
+1 for OPTIONS

avbk:
@JakeWharton sorry for the late answer. Yes, we do have a body specifying allowed values for the API for example.

JakeWharton:
Fixed, squashed, and merged. Thanks!

PR #1108, opened 2015-09-23
Added extra caution before calling onError/onCompleted

dlew:
The chances are low (but not zero) of an unsubscription between onNext() and
the terminal event.

### DIFF
--- a/retrofit-adapters/rxjava/src/main/java/retrofit/RxJavaCallAdapterFactory.java
+++ b/retrofit-adapters/rxjava/src/main/java/retrofit/RxJavaCallAdapterFactory.java
@@ -109,10 +109,14 @@ public final class RxJavaCallAdapterFactory implements CallAdapter.Factory {
           try {
             subscriber.onNext(response);
           } catch (Throwable t) {
-            subscriber.onError(t);
+            if (!subscriber.isUnsubscribed()) {
+              subscriber.onError(t);
+            }
             return;
           }
-          subscriber.onCompleted();
+          if (!subscriber.isUnsubscribed()) {
+            subscriber.onCompleted();
+          }
         }
 
         @Override public void onFailure(Throwable t) {


artem-zinnatullin:
It's not required because if user uses any overload of `subscribe()` — his `Subscriber` will be wrapped into [`SafeSubscriber`](https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/observers/SafeSubscriber.java).

The only way for the user to create possible problem for himself — use `unsafeSubscribe()`, and I think there is no much sense in giving protection from this :)

dlew:
@artem-zinnatullin `SafeSubscriber` will not protect you here. Try out this code:

```java
Observable<Long> observable = Observable.create(subscriber -> {
  for (long a = 0; a < 10; a++) {
    sleep(10);
    subscriber.onNext(a);
  }
  subscriber.onCompleted();
});

Subscription sub = observable
    .subscribeOn(Schedulers.newThread())
    .subscribe(
        l -> System.out.println("Received: " + l),
        e -> System.out.println("Error: " + e),
        () -> System.out.println("Completed")
    );
sleep(50);
sub.unsubscribe();
System.out.println("Unsubscribed!");
```

You'll find that the output is this:

```
Received: 0
Received: 1
Received: 2
Received: 3
Received: 4
Unsubscribed!
Received: 5
Received: 6
Received: 7
Received: 8
Received: 9
Completed
```

[Here are more things `SafeSubscriber` does not actually guarantee](https://github.com/ReactiveX/RxJava/issues/3059). At this point, I'm not entirely sure *what* `SafeSubscriber` does. :P

artem-zinnatullin:
@dlew Nooooo, I don't want to believe you! 

Hope when I'll wake up and check your code it will become just a bad dream!

akarnokd:
Unsubscription handling is best effort and events may slip through. As a consumer, one must be prepared for such occasion. Checking `isUnsubscribed()` could reduce the likelihood of a race, but it isn't 100% guarantee and isn't even mandatory. The problem is that it is impossible to serialize the events with the `unsubscribe` call because that is exposed to everyone in 1.x and is final thus not overridable. 

Where did this behavior cause you trouble?

dlew:
This is more of a preventative measure; I know that RxBinding ran into problems by not being aggressive enough with its checks.

PR #1109, opened 2015-09-23
Add delegation support to CallAdapter.Factory.

### LINKED ISSUE #946: Custom call adapter and reuse of default implementation classes
Hi Retrofit team,

I am trying retrofit 2.0, and the embryo of API is nice, for now it solves pretty well the signature de-duplication problem (having to provide a sync / async service signature).

In our app, we would like to generalize an error management process. I mean that most of our endpoints can return an http error code and some JSON to provide details about the failure. 

Actually, I would think that a pretty nice place to do it would be to get a custom `Call`object whose execute method could throw exceptions, the exceptions wrapping the returned error JSON.

However, to achieve this, it would be nice to be able to extend default implementations of `Call` like `ExecutorCallbackCall`. I could delegate most cases to it with a decorator pattern, and handle special error cases in it. 

Is that the suggested best practice to create a common error management entity ? If so, could you make the default implementations classes public ?

Keep up the good work,
 S.

JakeWharton:
Closes #946.

### DIFF
--- a/retrofit/src/main/java/retrofit/Retrofit.java
+++ b/retrofit/src/main/java/retrofit/Retrofit.java
@@ -19,9 +19,11 @@ import com.squareup.okhttp.HttpUrl;
 import com.squareup.okhttp.OkHttpClient;
 import com.squareup.okhttp.RequestBody;
 import com.squareup.okhttp.ResponseBody;
+import java.lang.annotation.Annotation;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
+import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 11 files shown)


retrofit/src/main/java/retrofit/Retrofit.java
```diff
@@ -147,6 +148,51 @@ public BaseUrl baseUrl() {
...
+   * #callAdapterFactories() factories} except {@code skipPast}.
+   */
+  public CallAdapter<?> nextCallAdapter(CallAdapter.Factory skipPast, Type returnType,
+      Annotation[] annotations) {
+    checkNotNull(returnType, "returnType == null");
+    checkNotNull(annotations, "annotations == null");
+
+    for (int i = 0, count = adapterFactories.size(); i < count; i++) {
+      CallAdapter.Factory factory = adapterFactories.get(i);
+      if (factory == skipPast) continue;
```
swankjesse: For recursive delegation you probably need to skip past both this and all prior!

swankjesse:
LGTM, though I think you need a code change for recursive delegation

retrofit/src/main/java/retrofit/Retrofit.java
JakeWharton: Never thought of this. Nice. Fixed.

PR #1114, opened 2015-09-24
Adding an HTTP logging module which is similar in output to V1's logging.

JakeWharton:
Refs #378.

### DIFF
--- a/retrofit-http-logging/pom.xml
+++ b/retrofit-http-logging/pom.xml
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>com.squareup.retrofit</groupId>
+    <artifactId>parent</artifactId>
+    <version>2.0.0-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+
+  <artifactId>retrofit-http-logging</artifactId>
+  <name>HTTP Logging Interceptor</name>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.squareup.retrofit</groupId>
+      <artifactId>retrofit</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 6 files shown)


retrofit-http-logging/pom.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
swankjesse: Should this project go in OkHttp?

JakeWharton: Sure, if you want it. I'm just trying to stop the torrent of SO, issues, and questions on how to log (since there's some tricky things and some not-so-tricky things).

retrofit-http-logging/src/main/java/retrofit/HttpLoggingInterceptor.java
```diff
@@ -0,0 +1,234 @@
...
+
+  public enum Level {
+    /** No logs. */
+    NONE,
+    /**
+     * Logs request and response lines.
+     * <p>
+     * Example:
+     * <pre>{@code
+     * --> POST /greeting HTTP/1.1 (3-body)
```
swankjesse: 3-byte body

retrofit-http-logging/pom.xml
swankjesse: Let's move it!

JakeWharton: :+1:

retrofit-http-logging/src/main/java/retrofit/HttpLoggingInterceptor.java
```diff
@@ -0,0 +1,234 @@
...
+  public HttpLoggingInterceptor(Logger logger) {
+    this.logger = logger;
+  }
+
+  private final Logger logger;
+
+  private volatile Level level = Level.NONE;
+
+  /** Change the level at which this interceptor logs. */
+  public void setLevel(Level level) {
```
swankjesse: Should this be just a factory method? `LoggingInterceptor.basic()` ? Or does it need to be mutable?

retrofit-http-logging/src/main/java/retrofit/HttpLoggingInterceptor.java
```diff
@@ -0,0 +1,234 @@
...
+          charset = contentType.charset(UTF8);
+        }
+
+        if (responseBody.contentLength() > 0) {
+          logger.log("");
+          logger.log(buffer.clone().readString(charset));
+        }
+
+        // Since we consumed the original, replace the one-shot body in the response with a new one.
+        response = response.newBuilder()
```
swankjesse: If you wanna save some work, you could load the full body into a buffer and call clone().

retrofit-http-logging/src/main/java/retrofit/HttpLoggingInterceptor.java
JakeWharton: We allow and frequently change the level at runtime in debug builds. Now that can be done with immutable interceptors and replacing, but you won't affect any `.clone()`'d instances.

retrofit-http-logging/src/main/java/retrofit/HttpLoggingInterceptor.java
JakeWharton: See three lines up!

swankjesse:
LGTM, on other project!

PR #1117, opened 2015-09-24
Use execute() instead of enqueue() in Rx adapter

dlew:
Retrofit should leave threading up to the Scheduler used to invoke call().

Otherwise, there can be two types of unwanted behavior:

1. It can execute on another Thread when sync behavior is desired.

2. It uses unnecessary extra Threads even when already called on a Scheduler.

### DIFF
--- a/retrofit-adapters/rxjava/src/main/java/retrofit/RxJavaCallAdapterFactory.java
+++ b/retrofit-adapters/rxjava/src/main/java/retrofit/RxJavaCallAdapterFactory.java
@@ -102,31 +102,25 @@ public final class RxJavaCallAdapterFactory implements CallAdapter.Factory {
         }
       }));
 
-      call.enqueue(new Callback<T>() {
-        @Override public void onResponse(Response<T> response, Retrofit retrofit) {
-          if (subscriber.isUnsubscribed()) {
-            return;
-          }
-          try {
-            subscriber.onNext(response);
-          } catch (Throwable t) {
-            if (!subscriber.isUnsubscribed()) {
-              subscriber.onError(t);
-            }
-            return;
-          }
-          if (!subscriber.isUnsubscribed()) {
-            subscriber.onCompleted();
-          }
-        }
+      if (subscriber.isUnsubscribed()) {
+        return;
+      }
 
-        @Override public void onFailure(Throwable t) {
-          if (subscriber.isUnsubscribed()) {
-            return;
-          }
+      try {
+        Response<T> response = call.execute();
+        if (!subscriber.isUnsubscribed()) {
+          subscriber.onNext(response);
+        }
+      } catch (Throwable t) {
+        if (!subscriber.isUnsubscribed()) {
           subscriber.onError(t);
         }
-      });
+        return;
+      }
+
+      if (!subscriber.isUnsubscribed()) {
+        subscriber.onCompleted();
+      }
     }
   }
 


artem-zinnatullin:
+1 to this change.

tomkoptel:
+1

JakeWharton:
Stop "+1"ing things. I realize GitHub doesn't have a voting system, but
you're making me hate Dan Lew for submitting this change (and every other
issue or PR this happens on).

nsk-mironov:
I'm against this PR. This is a breaking change and I bet there are a lot of applications that relies on retrofit doing its job in a background thread by default. I know 2.0 isn't supposed to be backward compatible, but this pull request will make the transition to 2.0 even more hard (and more error-prone, since every retrofit call will REQUIRE calling `subscribeOn` as well).

Any who cares enough about scheduling can always implement a custom `CallAdapter.Factory`.

dlew:
There are plenty of other breaking changes; I'm not sure this one is more severe than others. For example, now that `RetrofitError` doesn't exist, any code that used to use it has to move a lot of its logic into `onResponse()` now.

> Anyone who cares enough about scheduling can always implement a custom `CallAdapter.Factory`.

The same be said of people who want to keep the old behavior.

artem-zinnatullin:
The simplicity that gives current implementation is great, and I guess, most of the users are okay with `Schedulers.io()` by default.

But when you need to use RxJava, let's say, more carefully, you'll want not to have `subscribeOn()` set by default.

What about two separate artifacts with `RxJavaCallAdapterFactory` shipped as part of Retrofit? 1st one will be current implementation, and the 2nd one will be impl without default scheduling. And section in the readme/website about the difference between them.

JakeWharton:
No. That sounds awful as you're just pushing decision paralysis onto
consumers.

artem-zinnatullin:
>decision paralysis onto consumers.

Lol. I could say same about call-adapters and converters with different implementations :)

Okay, I'm leaving this discussion. Anyway, I'll use Dan's implementation. @dlew if this PR won't be merged, pls :shipit: as a separate artifact (or ping me so I'll do it).

JakeWharton:
Yes. _You_ could say that because you lack the context of the 5 years that Retrofit has been open source and dealing with people who use it.

artem-zinnatullin:
Yeah, about a year I am subscribed to Retrofit repo, tons of emails every day. Valid argument.

// leaving the discussion again

JakeWharton:
FYI the existing behavior will be restored in Retrofit 3 which will depend on OkHttp 3 as the threading model will be changed such that synchronous execution will actually be the one wasting threads.

adi1133:
Why don't you guys adopt the builder pattern for the factory and let the user choose if they want to use execute() or enqueue() internally.

I currently have to use a shiv to get the expected functionality I want

https://gist.github.com/adi1133/0942364293ddd13ef5f3

JakeWharton:
The number of users who understand the difference between the two is not
high enough to justify that. Also, there's no reason to not use execute()
with OkHttp 2.x as the number of threads is exactly the same.

adi1133:
**execute()** causes an **NetworkOnMainThreadException** on android and blocks the UI thread when subscribing

JakeWharton:
So subscribe on a worker thread, like you would any other blocking IO task.

nsk-mironov:
Just a thought: how about introducing a `Transformation` interface which will allow to transform an `Observable` returned from `adapt` call?

```java
public interface Transformation {
	public <T> Observable<T> transform(final Observable<T> observable);
}

public final class RxJavaCallAdapterFactory implements CallAdapter.Factory {
  public static RxJavaCallAdapterFactory create() {
    return new RxJavaCallAdapterFactory(null);
  }

  public static RxJavaCallAdapterFactory create(final Transformation transformation) {
    return new RxJavaCallAdapterFactory(transformation);
  }
}

RxJavaCallAdapterFactory.create(new Transformation() {
	@Override
	public <T> Observable<T> transform(final Observable<T> observable) {
		return observable.subscribeOn(Schedulers.io());
	}
});
```

JakeWharton:
You are describing what a CallAdapter.Factory already is!

PR #1118, opened 2015-09-24
Provide Retrofit to response callback. Add error converter sample.

### LINKED ISSUE #1053: Convert errorBody just like body
I have a JSON API that returns a non-20x Status Code.The response body looks similar to this:
```
{ "error": "Device limit exceeded" }
```
The behaviour of Retrofit currently seems to be to always return a `ResponseBody` for the `errorBody` ([see here](https://github.com/square/retrofit/blob/9cd5dfa0e0c66704ab035835259fde2721511237/retrofit/src/main/java/retrofit/Response.java#L69)).
Am I missing something? Should the errorBody not be parsed like the regular body?

JakeWharton:
Closes #1053.

### DIFF
--- a/retrofit-adapters/rxjava/src/main/java/retrofit/RxJavaCallAdapterFactory.java
+++ b/retrofit-adapters/rxjava/src/main/java/retrofit/RxJavaCallAdapterFactory.java
@@ -103,7 +103,7 @@ public final class RxJavaCallAdapterFactory implements CallAdapter.Factory {
       }));
 
       call.enqueue(new Callback<T>() {
-        @Override public void onResponse(Response<T> response) {
+        @Override public void onResponse(Response<T> response, Retrofit retrofit) {
           if (subscriber.isUnsubscribed()) {
             return;
           }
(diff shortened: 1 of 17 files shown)


swankjesse:
We're converting some code over to beta2 and this is tripping us up. I think we might be better off with the previous API, asking the caller to send their Retrofit instances through some other channel.

The problem is that `Calls.response()` is particularly awkward.

JakeWharton:
Isn't that the same problem but now in the production code instead of test
code? I'm not against it, because I like the value nature without the
Retrofit instance, but it seems like it's effectively the same thing just
who has to do the work.

swankjesse:
The difference is that previously the tests didn't need to create a `Retrofit` instance if they didn't need one. That means no bogus `example.com` base URL, and no anxiety about adapters being used in unexpected ways.

With this change tests now need to fake out a bigger environment. Rather than calling the callback they need to set stuff up, that may or may not be used by the subject code.

JakeWharton:
PRs welcome :trollface: 

I think this will also obviate #1142. cc @dlew

dlew:
I am not a fan of #1142, but the reason I did it was because it was necessary for us to convert to Retrofit (while still using RxJava). We need *some* way to convert the errors.

PR #1119, opened 2015-09-25
Improve error handling in RxJavaCallAdapterFactory

artem-zinnatullin:
1. `throwIfFatal()`, pretty important thing.
2. `addValueAsLastCause()` + `toString()` for `Call` -> much easier to debug errors!

### DIFF
--- a/retrofit-adapters/rxjava/src/main/java/retrofit/RxJavaCallAdapterFactory.java
+++ b/retrofit-adapters/rxjava/src/main/java/retrofit/RxJavaCallAdapterFactory.java
@@ -20,6 +20,7 @@ import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import rx.Observable;
 import rx.Subscriber;
+import rx.exceptions.Exceptions;
 import rx.functions.Action0;
 import rx.functions.Func1;
 import rx.subscriptions.Subscriptions;
@@ -112,6 +113,7 @@ public final class RxJavaCallAdapterFactory implements CallAdapter.Factory {
           subscriber.onNext(response);
         }
       } catch (Throwable t) {
+        Exceptions.throwIfFatal(t);
         if (!subscriber.isUnsubscribed()) {
           subscriber.onError(t);
         }


retrofit-adapters/rxjava/src/main/java/retrofit/RxJavaCallAdapterFactory.java
```diff
@@ -110,8 +112,9 @@ private CallOnSubscribe(Call<T> originalCall) {
           try {
             subscriber.onNext(response);
           } catch (Throwable t) {
+            Exceptions.throwIfFatal(t);
```
JakeWharton: Anything thrown from this will just get swallowed by OkHttp

retrofit-adapters/rxjava/src/main/java/retrofit/RxJavaCallAdapterFactory.java
```diff
@@ -121,10 +124,11 @@ private CallOnSubscribe(Call<T> originalCall) {
         }
 
         @Override public void onFailure(Throwable t) {
+          Exceptions.throwIfFatal(t);
```
JakeWharton: Similarly

retrofit-adapters/rxjava/src/main/java/retrofit/RxJavaCallAdapterFactory.java
JakeWharton: Actually I lied. It only eats IOExceptions. These will just kill the executor thread.

retrofit/src/main/java/retrofit/ExecutorCallAdapterFactory.java
```diff
@@ -69,6 +69,16 @@
     @Override public Call<T> clone() {
       return new ExecutorCallbackCall<>(callbackExecutor, delegate.clone());
     }
+
+    @Override
+    public String toString() {
+      return "ExecutorCallbackCall{"
+          + "callbackExecutor="
+          + callbackExecutor
```
JakeWharton: This is not useful information. Just call the delegate's toString.

retrofit/src/main/java/retrofit/OkHttpCall.java
```diff
@@ -157,4 +158,24 @@ public void cancel() {
...
+  public String toString() {
+    return "OkHttpCall{"
+        + "retrofit="
+        + retrofit
+        + ", requestFactory="
+        + requestFactory
+        + ", responseConverter="
+        + responseConverter
+        + ", args="
+        + Arrays.toString(args)
```
JakeWharton: This is troublesome as a password might be an argument to a service method. By moving its string value into an exception it could be inadvertently logged or passed to untrusted code.

artem-zinnatullin: Yeah… Similar to the WebSockets threads naming issue in the OkHttp.

>By moving its string value into an exception it could be inadvertently logged
I've removed `toString()` and `addValueAsLastCause()` for now.

But it won't help with untrusted code since it could use reflection to retrieve fields! (though reflection can be disabled).

artem-zinnatullin:
Updated PR, with changes from #1117 we can use `Exceptions.throwIfFatal()`.

JakeWharton:
Thanks

PR #1123, opened 2015-09-25
Use a correct implementation of path encoding.

### LINKED ISSUE #1045: Improper encoding of path segment
Here's a test case:

https://android.api.simple.com/atm-api/nearest/45.529801,-122.683633 (works)

Retrofit encodes as:

https://android.api.simple.com/atm-api/nearest/45.529801%2C-122.683633 (doesn't work)

According to [this guide](http://blog.lunatech.com/2009/02/03/what-every-web-developer-must-know-about-url-encoding) the `,` (comma) character should be allowed unencoded in the path.

Easily worked around with `encoded=true`.

JakeWharton:
At least until we figure out if we can/want to do replacement support in `HttpUrl`...

Closes #1045.


JakeWharton:
This is only slightly optimized for the use case. I'm sure there's more that could be made in the future due to the narrow scope for which the encoding is done.

PR #1125, opened 2015-09-25
HEAD requests must use Void response type.

JakeWharton:


retrofit/src/main/java/retrofit/RequestFactoryParser.java
```diff
@@ -100,6 +100,9 @@ private void parseMethodAnnotations() {
         parseHttpMethodAndPath("GET", ((GET) annotation).value(), false);
       } else if (annotation instanceof HEAD) {
         parseHttpMethodAndPath("HEAD", ((HEAD) annotation).value(), false);
+        if (!Void.class.equals(responseType)) {
+          throw methodError(method, "HEAD method must use Void as response type.");
```
swankjesse: what about `void` ?

swankjesse:
LGTM

retrofit/src/main/java/retrofit/RequestFactoryParser.java
JakeWharton: This is response type, not return type, so it'll be `Call<Void>`. I suppose someone could write a `CallAdapter` which took the raw return type as the response type, but for that unlikely case I'm going to say they can just map `void.class` to `Void.class`.

PR #1136, opened 2015-09-29
Check for cancellation on callback executor run.

### LINKED ISSUE #1135: The callback Executor should check for cancellation before delivering the result
Currently if the callback `Executor` executes the callback on a different thread (the default on Android), then there is no check performed for `Call` cancellation, and the result may be delivered even after it has been cancelled. This is not a good design, and may confound expectations from the client code (as e.g. `AsyncTask` handles these cases properly).

1zaman:
Fixes #1135.

### DIFF
--- a/retrofit/src/main/java/retrofit/ExecutorCallAdapterFactory.java
+++ b/retrofit/src/main/java/retrofit/ExecutorCallAdapterFactory.java
@@ -47,14 +47,36 @@ final class ExecutorCallAdapterFactory implements CallAdapter.Factory {
   static final class ExecutorCallbackCall<T> implements Call<T> {
     private final Executor callbackExecutor;
     private final Call<T> delegate;
+    private volatile boolean canceled;
 
     ExecutorCallbackCall(Executor callbackExecutor, Call<T> delegate) {
       this.callbackExecutor = callbackExecutor;
       this.delegate = delegate;
     }
 
-    @Override public void enqueue(Callback<T> callback) {
-      delegate.enqueue(new ExecutorCallback<>(callbackExecutor, callback));
+    @Override public void enqueue(final Callback<T> callback) {
+      delegate.enqueue(new Callback<T>() {
+        @Override public void onResponse(final Response<T> response, final Retrofit retrofit) {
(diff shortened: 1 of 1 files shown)


1zaman:
Hmm, it seems that the OkHttp `Call` implementation actually throws/delivers an `IOException` upon cancellation. Is that the desired behavior for asynchronous calls in OkHttp and Retrofit? Should I change my logic to also do the same?

JakeWharton:
Cancel should deliver an InterrupedIOException to allow the application
code to tear down any state it's keeping (like a progress dialog, for
example).

1zaman:
Yes, I had updated my comment and changes that day. Sorry you missed it. I should have posted another comment.

JakeWharton:
Oops! I was triaging issues over email. Not the first time someone has
updated a comment and I've missed it. I'll look tomorrow.

retrofit/src/main/java/retrofit/ExecutorCallAdapterFactory.java
```diff
@@ -47,14 +47,36 @@
...
 
-    @Override public void enqueue(Callback<T> callback) {
-      delegate.enqueue(new ExecutorCallback<>(callbackExecutor, callback));
+    @Override public void enqueue(final Callback<T> callback) {
+      delegate.enqueue(new Callback<T>() {
+        @Override public void onResponse(final Response<T> response, final Retrofit retrofit) {
+          callbackExecutor.execute(new Runnable() {
+            @Override public void run() {
+              if (canceled) {
+                // Emulate OkHttp's behavior of throwing/delivering an IOException on cancelation
```
JakeWharton: Do we need this, though? If OkHttp throws an exception when canceled it will be automatically propagated to the callback. I think there is even a test for this already.

1zaman: This change is for handling the case where the `Call` is cancelled after the response has been delivered by OkHttp, while still being delivered by the callback `Executor`. As mentioned in #1135, this would cause the result to be delivered to the callback even after cancellation.

PR #1138, opened 2015-09-29
Common prefix for all methods in an interface

larsgrefer:
Allow a common prefix for all methods in an interface by annotating the interface:
```java
@PathPrefix("/foo")
public interface FooService {
    @GET("/bar)
    Call<Object> bar(...);
}
```
The bar() method would produce the relative url ```/foo/bar```

### DIFF
--- a/retrofit/src/main/java/retrofit/RequestFactoryParser.java
+++ b/retrofit/src/main/java/retrofit/RequestFactoryParser.java
@@ -26,6 +26,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+
 import retrofit.http.Body;
 import retrofit.http.DELETE;
 import retrofit.http.Field;
@@ -44,6 +45,7 @@ import retrofit.http.PUT;
 import retrofit.http.Part;
 import retrofit.http.PartMap;
 import retrofit.http.Path;
+import retrofit.http.PathPrefix;
 import retrofit.http.Query;
 import retrofit.http.QueryMap;
 import retrofit.http.Url;
@@ -69,6 +71,7 @@ final class RequestFactoryParser {
   private boolean hasBody;
   private boolean isFormEncoded;
   private boolean isMultipart;
+  private String pathPrefix;
   private String relativeUrl;
   private com.squareup.okhttp.Headers headers;
   private MediaType contentType;
@@ -92,6 +95,18 @@ final class RequestFactoryParser {
 
   private RuntimeException parameterError(int index, String message, Object... args) {
     return methodError(method, message + " (parameter #" + (index + 1) + ")", args);
+  }
+
+  private String getPathPrefix() {
+    if (pathPrefix == null) {
+      if (method.getDeclaringClass().isAnnotationPresent(PathPrefix.class)) {
+        PathPrefix pathAnnotation = method.getDeclaringClass().getAnnotation(PathPrefix.class);
+        pathPrefix = pathAnnotation.value();
+      } else {
+        pathPrefix = "";
+      }
+    }
+    return pathPrefix;
   }
 
   private void parseMethodAnnotations(Type responseType) {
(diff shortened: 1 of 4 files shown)


sohayb:
:+1:

Frisch12:
:+1:

RoyMontoya:
awesome(Y)

larsgrefer:
@JakeWharton Whats your opinion?

retrofit/src/main/java/retrofit/RequestFactoryParser.java
```diff
@@ -155,15 +170,17 @@ private void parseHttpMethodAndPath(String httpMethod, String value, boolean has
...
     }
 
     // Get the relative URL path and existing query string, if present.
-    int question = value.indexOf('?');
-    if (question != -1 && question < value.length() - 1) {
+    int question = combinedRelativeUrl.indexOf('?');
+    if (question != -1 && question < combinedRelativeUrl.length() - 1) {
       // Ensure the query string does not have any named parameters.
-      String queryParams = value.substring(question + 1);
+      String queryParams = combinedRelativeUrl.substring(question + 1);
```
RoyMontoya: is there any other queryParams validation happening after this? since the if only checks that the question mark exists and that is not the last character. someone could pass an invalid param like "name="

RoyMontoya: is there any queryParams validation happening? because as i see it someone could pass an invalid param string like "name="

artem-zinnatullin:
Why not add a prefix (postfix) to the `baseUrl` during the creation of `Retrofit` instance?

larsgrefer:
@artem-zinnatullin If you`ll add the prefix as postfix to the base URL all interfaces would have this prefix.
I intend to allow diffrenet prefixes for different service interfaces.

Look at this exampe:
```java
    .baseUrl("https://api.example.com/rest/")
    .build();
```
First service interface:
```java
@PathPrefix("users")
public interface UserService{
    @GET
    Call<List<User>> getAll();

    @GET("/{id}")
    Call<User> getById(@Path("id) int id);
}
```
Second service interface
```java
@PathPrefix("products")
public interface ProductService{
    @GET
    Call<List<Product>> getAll();

    @GET("/{id}")
    Call<Product> getById(@Path("id) int id);
}
```
Resulting urls
```java
UserService.getAll() ->     "https://api.example.com/rest/users"
UserService.getById() ->    "https://api.example.com/rest/users/{id}"
ProductService.getAll() >   "https://api.example.com/rest/products"
ProductService.getById() -> "https://api.example.com/rest/products/{id}"
```

retrofit/src/main/java/retrofit/RequestFactoryParser.java
larsgrefer: @RoyMontoya I'm not sure if this PR is the right place to discuss this. 
Would the problem you are describing here be introduced by this PR, or does it already exist in master?

larsgrefer:
Just an idea:

The interface `@PathPrefix` could be renamed to `@RestService`(or something) and argument could be made optional.

Doing this, we`d have an interface for denoting all Retrofit Rest Services, this might be useful for finding them over reflection.

retrofit/src/main/java/retrofit/RequestFactoryParser.java
RoyMontoya: @larsgrefer you are right!, i didnt though of that, let me check it out! there should be a validation on master

larsgrefer:
I've rebased the branch on the current master in order to rerun the tests

JakeWharton:
We've decided not to move forward with this, at least for 2.0. First of all, if we did add something like this I'm not sure concatenation is the best route. We would probably want to use `HttpUrl`'s `resolve` method to allow using things like absolute paths or even different hosts. But most of all, promoting something to method-level opens more doors than we want right now. If we allow paths to be defined at the interface level then what about headers? And in that case what's the precedence of headers? Are they concatenated with method-level headers or do method-level headers replace? There is a lot to be considered if we were to allow this and we're just not ready to do that. For now, putting the prefix on every endpoint or using a constant is the recommended solution.

PR #1142, opened 2015-09-29
Added Retrofit to CallAdapter and Result

dlew:
`Callback.onResponse()` provides `Retrofit` now, so it makes sense that the call adapters would also need to provide it as well.

Two changes:

- Added `Retrofit` to `CallAdapter.adapt()` as a convenience (otherwise you have to pass `Retrofit` around quite a bit).
- Added `Retrofit` to `Result` so that the RxJava adapter can parse errors using `Retrofit`.

### DIFF
--- a/retrofit-adapters/rxjava/src/main/java/retrofit/Result.java
+++ b/retrofit-adapters/rxjava/src/main/java/retrofit/Result.java
@@ -22,18 +22,21 @@ import static retrofit.Utils.checkNotNull;
 /** The result of executing an HTTP request. */
 public final class Result<T> {
   public static <T> Result<T> error(Throwable error) {
-    return new Result<>(null, checkNotNull(error, "error == null"));
+    return new Result<>(null, null, checkNotNull(error, "error == null"));
   }
 
-  public static <T> Result<T> response(Response<T> response) {
-    return new Result<>(checkNotNull(response, "response == null"), null);
+  public static <T> Result<T> response(Response<T> response, Retrofit retrofit) {
+    return new Result<>(checkNotNull(response, "response == null"),
+      checkNotNull(retrofit, "retrofit == null"), null);
   }
 
   private final Response<T> response;
+  private final Retrofit retrofit;
   private final Throwable error;
 
-  Result(Response<T> response, Throwable error) {
+  Result(Response<T> response, Retrofit retrofit, Throwable error) {
     this.response = response;
+    this.retrofit = retrofit;
     this.error = error;
   }
 
@@ -43,6 +46,14 @@ public final class Result<T> {
    */
   public Response<T> response() {
     return response;
+  }
+
+  /**
+   * The Retrofit used to make this request. Only present when {@link #isError()} is
+   * false, null otherwise.
+   */
+  public Retrofit retrofit() {
+    return retrofit;
   }
 
   /**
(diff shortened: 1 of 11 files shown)


retrofit-adapters/rxjava/src/main/java/retrofit/RxJavaCallAdapterFactory.java
```diff
@@ -173,13 +182,8 @@ private CallOnSubscribe(Call<T> originalCall) {
...
-    @Override public <R> Observable<Result<R>> adapt(Call<R> call) {
-      return Observable.create(new CallOnSubscribe<>(call)) //
-          .map(new Func1<Response<R>, Result<R>>() {
-            @Override public Result<R> call(Response<R> response) {
-              return Result.response(response);
-            }
-          })
+    @Override public <R> Observable<Result<R>> adapt(Call<R> call, Retrofit retrofit) {
+      return Observable.create(new CallOnSubscribe<>(call, retrofit)) //
           .onErrorReturn(new Func1<Throwable, Result<R>>() {
```
JakeWharton: This can be removed now.

dlew: I thought that initially, but that would require us calling `onNext(Result.error())` in `CallOnSubscribe`, which means that for the other adapters we'd have to purposefully throw that error later.

JakeWharton: Oh, no it can't. But it _could_. I find it a bit strange that `CallOnSubscribe` emits `Result` instances for responses but not for errors. I think we probably want it to conform to always emitting a `Result` instance and then have these functions decompose its instances.

retrofit-adapters/rxjava/src/main/java/retrofit/Result.java
```diff
@@ -46,6 +49,14 @@
   }
 
   /**
+   * The Retrofit used to make this request. Only present when {@link #isError()} is
+   * false, null otherwise.
```
JakeWharton: Is this conditional useful? Or should we always just include it?

retrofit-adapters/rxjava/src/main/java/retrofit/RxJavaCallAdapterFactory.java
dlew: I'll make a version like that.

dlew:
I've added a version which always returns `Result` and just flatMaps the errors.

retrofit-adapters/rxjava/src/main/java/retrofit/Result.java
dlew: I'm not sure. It's just mimicking the behavior of `Callback` right now.

dlew:
Is there anything more that needs to be done? Or stewed upon?

JakeWharton:
I'm stewing. But I think we'll probably move forward with this.

My big worry is that it feels a lot like RetrofitError in 1.x which had to
have a Converter supplied. Result feels like it should be a simple value
object, and not something that carries or requires the "giant" Retrofit
instance to create.

dlew:
I agree with you, but having `Retrofit` is better for decoding errors.

It seems like whatever solution is decided on here should also apply to the non-Rx version.

JakeWharton:
Closing because of what was discussed in #1118 and #1276. BYOR (Bring your own `Retrofit`).

Issue #1156, opened 2015-10-02
Path encoding


mPetrus:
I have problem with path encoding. 

I have request like this:
```
@GET("{url}")
Observable<Response<Void>> getItem(@Path(value = "url", encoded = true) String url);
```

and I need use url like 
2015-10-01/560d3447c28eb5853953101b/560d3448fca6b4191105633a.jpg
 but retrofit change it to 
/2015-10-01%2F560d3447c28eb5853953101b%2F560d3448fca6b4191105633a.jpg.
It works in beta1 but it doesn't work in beta2.

juliendn:
I got the same issue passing from 2.0.0-beta1 to 2.0.0-beta2, '/' characters are encoded even if I set encoded = true in path annotation.

venilnoronha:
Hi all,

As Jake has mentioned on my PR, it would be better to model the Service as:

@GET
Observable<Response<Void>> getItem(@Url String url);

This should solve your issue.

JakeWharton:
If you need a full path you should use as @venilnoronha said. Partial path replacements is when you should use `@Path` and it can either do encoding or not based on the rules that a path part must adhere to.

juliendn:
How to solve case with prefix containing more than one segment (ie containing '/' characters) ?

```java
@GET("{prefix}/segment1/segment2")
Call<MyObject> getMyObject(@Path(value = "prefix", encoded = true) String prefix);

/////

getMyObject("/api/2.0");

// result
// GET http://host/%2Fapi%2F2.0/segment1/segment2 http 1.1
```

JakeWharton:
That should be put on the base URL passed to `Retrofit.Builder`, not on the individual paths.

PR #1159, opened 2015-10-02
Fixes - Path encoding #1156

venilnoronha:
Hi all,

This commit should fix the issue #1156 - Path encoding.

Earlier, the alreadyEncoded flag was ignored while encoding the / character in the canonicalize method of RequestBuilder.java. I've now explicitly handled encoding for this particular character to resolve the issue.

I've also created a JUnit test to verify the fix. You can find the test under RequestBuilderTest.java with the name getWithEncodedRelativeUrl.

Please review and pull.

Thank you.


swankjesse:
I don't like the model here. Either the path segment is encoded, or it isn't. This is partial.

JakeWharton:
The test case should be using an `@Url` parameter with no path specified in the `@GET` annotation. This allows resolving a relative URL which contains slashes.

venilnoronha:
Okay. Thanks Jake.

PR #1169, opened 2015-10-06
Build with Gradle

larsgrefer:
I've just tried if its possible to move from maven to gradle and it works.

Here are the results


JakeWharton:
What's the value proposition of switching? Having used both build system extensively, I cannot think of any.

larsgrefer:
I can't think of any, too (except that gradle is much cooler :stuck_out_tongue_closed_eyes:), but if someone wants to move to gradle someday: here is how.

JakeWharton:
At this time, Maven is a much easier build system to use for managing open source projects. The release process, CI process, and IDE integration are all superior to Gradle. Without some additional value to offset the fact that Gradle is worse at these things, I don't see any reason to switch at this time.

JakeWharton:
I appreciate you putting in the work, but for something so drastic please create an issue ahead of time that we can potentially save you some effort or direct it towards something we know will be integrated.

PR #1176, opened 2015-10-07
Update index.html

danielchristopher1:
It's misleading if it says to "Create a class that implements the Converter.Factory interface" because I tried to implement it and then realized that it's an abstract class.


JakeWharton:
Pull requests for the website should be done on the master branch, as it's
versioned with the code.

JakeWharton:
The change looks good though! So please resubmit

PR #1184, opened 2015-10-08
Support Iterable values by QueryMap

nsmolenskii:
Adding an ability to use ```<T extends Iterable<?>>``` as values of ```@QueryMap``` to handle cases when you have to provide more that one value into query parameter by using ```@QueryMap```

```java 

interface OrderApi {
    @GET(/orders)
    List<Order> findOrders(@QueryMap Map<String, Object> filter);
}

class OrderService {
    private OrderApi orderApi;

    public List<Order> findPendingOrders(Date createdBefore, OrderState... states){
        Map<String, Object> filter = new HashMap<>();
        filter.put("created_before", createdBefore);
        filter.put("state_in", Arrays.asList(OrderState.NEW, OrderState.PENDING));
        return orderApi.findOrders(filter);
    }
}

### DIFF
--- a/retrofit/src/main/java/retrofit/RequestBuilder.java
+++ b/retrofit/src/main/java/retrofit/RequestBuilder.java
@@ -217,7 +217,7 @@ final class RequestBuilder implements RequestInterceptor.RequestFacade {
       }
       Object entryValue = entry.getValue();
       if (entryValue != null) { // Skip null values.
-        addQueryParam(entryKey.toString(), entryValue.toString(), encodeNames, encodeValues);
+        addQueryParam(entryKey.toString(), entryValue, encodeNames, encodeValues);
       }
     }
   }
--- a/retrofit/src/test/java/retrofit/RequestBuilderTest.java
+++ b/retrofit/src/test/java/retrofit/RequestBuilderTest.java
@@ -1149,11 +1149,12 @@ public class RequestBuilderTest {
     params.put("kit", "kat");
     params.put("foo", null);
     params.put("ping", "pong");
+    params.put("key", Arrays.<Object>asList(1, 2, null, "three"));
 
     Request request = buildRequest(Example.class, params);
     assertThat(request.getMethod()).isEqualTo("GET");
     assertThat(request.getHeaders()).isEmpty();
-    assertThat(request.getUrl()).isEqualTo("http://example.com/foo/bar/?kit=kat&ping=pong");
+    assertThat(request.getUrl()).isEqualTo("http://example.com/foo/bar/?kit=kat&ping=pong&key=1&key=2&key=three");
     assertThat(request.getBody()).isNull();
   }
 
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by JakeWharton** [comment]
Been busy. Will look soon. There's an issue for this, but also one for
supporting custom types (like Guava's Multimap). We'd also want to add this
to PartMap and FieldMap for party, but that could be in a follow-up.

**Comment by RoyMontoya** [comment]
@nsmolenskii hey awesome job, i was just wondering, do you know why if the implementation of addQueryParam with an object was there, it wasn't being used before?

**Comment by nsmolenskii** [comment]
@RoyMontoya actually i don't know the reason. But in case that it's could be a problem I could produce back compatible solution like:
```java 
    @QueryMap(iterableValuesSupport = true) Map<String, Object> filter
```

RoyMontoya:
@nsmolenskii dont get me wrong, the fix is cool but its the first time i see a fix like that, were all the implementation was ready but no one deliverate it xD!

ming13:
Just curious—is there any way to achieve such behaviour without this PR applied? I mean, I have multiple values for a parameter `foo` and the desired output is `foo=42&foo=24&foo=21&foo=12`.

nsmolenskii:
Of course, it is. You are able to write:

```java
interface FooApi {
    @GET(/orders)
    List<Foo> query(@Query List<?> foo);
}
```
and than use it:
```java
List<Foo> foos = fooApi.query(Arrays.asList(42, 24, 21, 21));
```

Unfortunately, this approach is require from You have to mention every possible parameter in method signature. Unfortunately this approach is not acceptable if You have a lot of optional params or You try to work with filters dynamically.

ming13:
@nsmolenskii Yep, I’m aware of this approach, but thanks anyway. Unfortunately I’m accessing API where some query parameters are available dynamically only, i. e. names and values are determined at runtime. Let’s hope your change will be available at Retrofit 2.0 this way or another! :angel:

pawel-kaminski-krk:
me like it!

JakeWharton:
There's an issue tracking this feature, #1324. We'll add it in some way in the future. For now I'm going to close the PR though.

dyguests:
wait for this.

jsroyal:
public interface Service {
    @GET("/search/users")
    Call<ItemResponse> getItems(
            @Query("location") String location
        //@Query("language") String language
    );
}

getting null
Not working this method

nsmolenskii:
Hi, @jsroyal. It seems to be that you comment to a wrong issue, because:
* This issue relates to version `1.x` and your example looks like `2.x` code because of using `Call`
* This issue was rejected cause it should be already implemented in `2.x`

zsaltys:
@swankjesse @nsmolenskii Would you know what is the status with this thing? Is it actually supported in version 2? Are there any plans to support it? It seems like all tickets for this have been closed down?

nsmolenskii:
Hi @zsaltys. That particular PR was rejected. Required functionality supported in version 2.
When have used patched 1.x version , until switch to openfeign

Issue #1140, opened 2015-09-29
Add a way to set filename on multipart requests


lukaciko:
This has already been discussed in #1063 #1092 #1096 and #1130, but I thought I'd open a separate issue to track this specifically.

[RFC2388](https://www.ietf.org/rfc/rfc2388.txt) describes the `filename` header parameter of `multipart/form-data` requests that Retrofit now uses. This parameter is required on some backend configurations. More details and examples can be found [here](http://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.2).

We need a way of figuring out what the filename is. The problem with `RequestBody` (which is a part of OkHttp) is that it doesn't expose the filename when created with a `File`.

erikdcch:
Hi @lukaciko I had the same problem but finally I found the solution of this bug. I just had to add the name of the field in @Part() RequestBody.

This is an example: 
Call<Object> postMethod(@Part("picture\"; filename=\"1.jpg\" ") RequestBody file);

picture - is the name of the parameter and after put the "filename" parameter with any name.

Note: Do not forget scape quotation marks  :)

wKovacs64:
That may work, but it seems a bit hacky. Inherent support would seem more natural.

Should this be an OkHttp issue rather than a Retrofit one? Maybe the `File` overload of the `RequestBody.create()` could automatically include the `filename` header parameter obtained via `File#getName()` or something.

lukaciko:
Injecting the filename can be used as a workaround, but it's not exactly pretty. It'd be better if it could be defined with an optional annotation parameter (i.e. `@Part(value="picture", filename="cool_cat.jpg")` but this still has the bigger issue - form field name is known in advanced and is thus a part of the interface, but filenames are dynamic.

@JakeWharton you've advocated using `RequestBody` directly when creating requests. Any ideas how we could change the API to accommodate this?

JakeWharton:
Sorry I haven't been too involved here. There is a lot going on in this
library, others, and internally, so I can only focus on so many things at
once.

Most of what you describe has nothing to do with Retrofit, but OkHttp.
RequestBody is an OkHttp type and internally we use OkHttp's multipart
builder. The filename property of TypedInput always felt out-of-place, so I
don't think we're likely to add it to RequestBody (which is the analogous
to TypedInput now).

I would encourage you to experiment with it and see possible ways forward
if you can. Otherwise I will take a look before v2.0, but I can't promise
when that will be just yet.

[...]

Issue #1063, opened 2015-09-07
How to Upload a File using Retrofit 2.0


vivekkiran:
How to upload a file using multipart/form-data upload in Retrofit 2.0, I am unable to upload and there is no proper documentation anywhere on uploading file using retrofit 2.0, I followed the docs in retrofit website but I was not able to make it work. 

All the examples I see used TypedFile. It is not clear how to use RequestBody. Please post sample code.

JakeWharton:
Same as v1. Use the Multipart annotation on the method and Part annotation
on parameters.

JakeWharton:
Oh for the file itself, use RequestBody for the type.

VadimWelldone:
Hi,

I am getting following exception while executing:

    @Headers("User-Agent: " + Constants.CLIENT_NAME)
    @Multipart
    @POST(Constants.Urls.SYNC_URL)
    Call<BaseResponseDto> uploadFile(@Part("file") RequestBody song, @Part("file_path") String fileName, @Part("method") String method, @Part("user_id") String userId, @Part("token") String token);

09-07 17:49:38.072  13562-13582/? W/EGL_emulation﹕ eglSurfaceAttrib not implemented


java.lang.IllegalStateException: JSON must start with an array or an object.
            at com.google.gson.stream.JsonWriter.beforeValue(JsonWriter.java:609)
            at com.google.gson.stream.JsonWriter.value(JsonWriter.java:418)
            at com.google.gson.internal.bind.TypeAdapters$13.write(TypeAdapters.java:362)
            at com.google.gson.internal.bind.TypeAdapters$13.write(TypeAdapters.java:346)
            at com.google.gson.TypeAdapter.toJson(TypeAdapter.java:142)
            at retrofit.GsonConverter.toBody(GsonConverter.java:55)
            at retrofit.RequestBuilderAction$Part.perform(RequestBuilderAction.java:204)
            at retrofit.RequestFactory.create(RequestFactory.java:62)
            at retrofit.OkHttpCall.createRawCall(OkHttpCall.java:116)
            at retrofit.OkHttpCall.execute(OkHttpCall.java:106)
            at retrofit.ExecutorCallAdapterFactory$ExecutorCallbackCall.execute(ExecutorCallAdapterFactory.java:59)
            at com.ibroadcast.mediasynclite.api.upload.UploadRunnable.run(UploadRunnable.java:49)
[...]

PR #1188, opened 2015-10-09
Introduce @PartFile for including TypedFile parts in a multi-part request

wKovacs64:
Since the removal of `TypedFile` in Retrofit 2.x, it has been somewhat of a challenge trying to figure out how to include a `File` part in a multi-part request. @JakeWharton has recommended using `RequestBody` but it doesn't include the `filename` property when created using a `File`, which some backends require. See #1140 for more information and discussion.

I've personally been using `@PartMap`, similar to @ayon115's [comment](https://github.com/square/retrofit/issues/1063#issuecomment-145920568) in #1063, but @BugsBunnyBR suggested a `@PartFile` annotation which sounded like a nice convenience. I took a stab at it here.

It [may need some help](http://www.quickmeme.com/img/d6/d6a1143f571184db25f94613edd43b40af6d3a629221aba00d9efdcfef5efd84.jpg).

Example usage (added 2015-12-01):

```java
public interface SomeService {
    @Multipart
    @POST("images/new")
    Call<UploadResponse> upload(@PartFile("image") TypedFile myImage);
}

// ... (build your service, etc.)

File file = getSomeFile("picture.png");
TypedFile image = new TypedFile("image/png", file);
Call<UploadResponse> response = someService.upload(image);
```

Resulting request headers:

```
Content-Disposition: form-data; name="image"; filename="picture.png"
Content-Type: image/png
Content-Transfer-Encoding: binary
```

Note: the `filename` value defaults to the result of `File#getName()` but you can override it when creating the `TypedFile` if desired:

```java
TypedFile foo = new TypedFile("image/png", file, "customFilename.png")
```

### DIFF
--- a/retrofit/src/main/java/retrofit2/RequestAction.java
+++ b/retrofit/src/main/java/retrofit2/RequestAction.java
@@ -229,6 +229,26 @@ abstract class RequestAction<T> {
     }
   }
 
+  static final class PartFile extends RequestAction {
+    private final String partName;
+
+    PartFile(String partName) {
+      this.partName = partName;
+    }
+
+    @Override void perform(RequestBuilder builder, Object value) {
+      if (value == null) return; // Skip null values.
+
+      TypedFile typedFile = (TypedFile) value;
+      Headers headers = Headers.of(
+              "Content-Disposition", "form-data; name=\"" + partName + "\"; filename=\""
+                      + typedFile.fileName() + "\"",
+              "Content-Transfer-Encoding", "binary");
+      RequestBody body = RequestBody.create(typedFile.mediaType(), typedFile.file());
+      builder.addPart(headers, body);
+    }
+  }
+
   static final class Body<T> extends RequestAction<T> {
     private final Converter<T, RequestBody> converter;
 
--- a/retrofit/src/main/java/retrofit2/RequestFactoryParser.java
+++ b/retrofit/src/main/java/retrofit2/RequestFactoryParser.java
@@ -42,6 +42,7 @@ import retrofit2.http.PATCH;
 import retrofit2.http.POST;
 import retrofit2.http.PUT;
 import retrofit2.http.Part;
+import retrofit2.http.PartFile;
 import retrofit2.http.PartMap;
 import retrofit2.http.Path;
 import retrofit2.http.Query;
@@ -454,6 +455,18 @@ final class RequestFactoryParser {
 
             PartMap partMap = (PartMap) methodParameterAnnotation;
             action = new RequestAction.PartMap<>(valueConverter, partMap.encoding());
+            gotPart = true;
+
+          } else if (methodParameterAnnotation instanceof PartFile) {
+            if (!isMultipart) {
+              throw parameterError(i,
+                  "@PartFile parameters can only be used with multipart encoding.");
+            }
+            if (methodParameterType != TypedFile.class) {
+              throw parameterError(i, "@PartFile parameter type must be TypedFile.");
+            }
+            PartFile partFile = (PartFile) methodParameterAnnotation;
+            action = new RequestAction.PartFile(partFile.value());
             gotPart = true;
 
           } else if (methodParameterAnnotation instanceof Body) {
--- a/retrofit/src/main/java/retrofit2/TypedFile.java
+++ b/retrofit/src/main/java/retrofit2/TypedFile.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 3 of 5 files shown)


BugsBunnyBR:
@wKovacs64 , very nice!
I have just one suggestion, the contentType should be optional and the api could try to detect it automatically, something like http://stackoverflow.com/questions/8589645/how-to-determine-mime-type-of-file-in-android

JakeWharton:
Retrofit is not an Android library and cannot depend on its APIs. But
besides, mime detection is a losing game that would require constant
attention and updating.

BugsBunnyBR:
yet, it may be a good thing to be a dynamic detection.

JakeWharton:
If the type is not known explicitly, the caller can dynamically detect
using whatever platform APIs and libraries are available to them.

wKovacs64:
I tried using `MimetypesFileTypeMap` but it always returned `application/octet-stream` (some quick searching reveals why). I agree with Jake, it's a losing battle.

BugsBunnyBR:
As far as I could check in the PR, I would need an interface  method to each contentType I would want to support, right?

wKovacs64:
As it stands currently, yes. Open to suggestions on an alternative approach, though.

BugsBunnyBR:
I miss TypedFIle.
One of the nice things of retrofit 2.0 is that we don't need to replicate interface methods to execute sync and async calls. With the current PR it would need to replicate interface methods to archive multiple file types upload.
If retrofit can't detect detect the content type by itself, something like TypeFile where you could give the File and content type String dynamically would be needed... and why not the filename.

wKovacs64:
Are you suggesting replacing `File` with a wrapper (say, a new `TypedFile`)? If so, can we still rely on getting the filename from `File#getName()` or do you think we'd need to provide a field in the wrapper to explicitly set/override it?

BugsBunnyBR:
I think the new TypedFile could have at least 2 constructor methods.. 
1) just a file and the content type.. use file.getName() for filename
2) file, content type and filename.

wKovacs64:
Any reason to include the encoding, or can we force `binary`?

BugsBunnyBR:
don't know doc, @JakeWharton ?

BugsBunnyBR:
How about just use the okhttp MultipartRequestBuilder ?

wKovacs64:
Something like this, if we pursue the wrapper idea?
```java
public final class TypedFile {
  private final MediaType mediaType;
  private final File file;
  private final String fileName;

  public TypedFile(MediaType mediaType, File file) {
    this.mediaType = checkNotNull(mediaType, "MediaType cannot be null.");
    this.file = checkNotNull(file, "File cannot be null.");
    this.fileName = checkNotNull(file.getName(), "Filename cannot be null.");
  }

  public TypedFile(MediaType mediaType, File file, String fileName) {
    this.mediaType = checkNotNull(mediaType, "MediaType cannot be null.");
    this.file = checkNotNull(file, "File cannot be null.");
    this.fileName = checkNotNull(fileName, "Filename cannot be null.");
  }

  public MediaType mediaType() {
    return mediaType;
  }

  public File file() {
    return file;
  }

  public String fileName() {
    return fileName;
  }
}
```
If we go this route, I guess we could just use `@Part` and a `Converter` for `TypedFile`? This is starting to look like Retrofit 1.x.

BugsBunnyBR:
Yes, try to reuse the second constructor method to set the values in the first constructor method. Something like this(type,file, file.getname());

wKovacs64:
Yeah, sure. Mostly wondering about the `@Part` and `Converter` part.

wKovacs64:
@BugsBunnyBR What about this? https://github.com/wKovacs64/retrofit/commit/76433edb2f9cdca446f1000dd013f2aad3c20157

I'm still wrestling with the feeling we could/should use `@Part` with a `Converter` but might need someone more intimate with the project to comment on that.

BugsBunnyBR:
To me it is fine in this way, but someone of square could say more what is the best strategy here. Let's wait.

BugsBunnyBR:
@JakeWharton , could you share your thoughts about this pull request ideas?

px-amaac:
@wKovacs64 How were you thinking of using @Part with converter for TypedFile?  I tried doing this but ran into problems when it came to actually editing the Header like you did in @PartFile RequestBuilderAction.  Perhaps I am missing something or do not know enough about the project but I could not figure out where to edit the header. Which means I was still left with the header injection in the @Part parameter.

wKovacs64:
@px-amaac I tried it briefly and couldn't get it either. If it came down to that being the more correct approach than the current state of the PR, I would need help from Jake/Jesse/whoever.

wKovacs64:
Rebased and squashed.

JakeWharton:
OkHttp has a `Multipart.Part` type now which is what we're going to use for this with the normal `@Part` annotation. It can't be integrated until we bump to the latest OkHttp version which I want to do after releasing another beta.

Thanks for taking the time to push this forward though.

wKovacs64:
Sounds good, will wait for that then. Thanks.

PR #1191, opened 2015-10-12
Moar useful info for exceptions!

### LINKED ISSUE #1162: Include debug info into exceptions
What do you think about a flag for including debug info (such as `request.toString()`) into exceptions?

Usage:
```java
    .baseUrl(API_URL)
    .addCallAdapterFactory(RxJavaCallAdapterFactory.create())
    .includeDebugInfoIntoExceptions(BuildConfig.DEBUG) // only for internal builds!
    .build();
```

Example of a stacktrace of incorrect call via RxJava **without** debug info:
screen shot 2015-10-03 at 23 22 48

Example of a stacktrace of incorrect call via RxJava **with** included debug info:
screen shot 2015-10-03 at 23 23 14

Async request (RxJava, etc) loses information about its start point (jumps to the other Thread and booms there), stacktraces become useless :crying_cat_face:. 

Logging is a great helper too! But sometimes you just don't want to log all the requests to keep the log readable, but you also want to see as much info as possible about the problems. Once I've spent several hours on trying to fix a request that was correct but because of a massive amount of async requests in the app — the last thing before the crash that I saw was a log of this correct request when the actual problem was in a request that fired few seconds before this.

At the moment, I have a draft of this feature where I add `request.toString()` to the exception message. And it works even if you fail before the request execution (converters/calladapters detection problems), also we can include more info in case of response parsing fail.

artem-zinnatullin:
Closes #1162.

Now stacktraces make sense!

**Before:**
screen shot 2015-10-03 at 23 22 48

**Now:**
screen shot 2015-10-12 at 06 07 22

Though I don't like at least two things:
1. Name of the  `infoForException`
2. Need in wrapping exceptions into `RuntimeException` just to set message.

### DIFF
--- a/retrofit/src/main/java/retrofit/Call.java
+++ b/retrofit/src/main/java/retrofit/Call.java
@@ -27,6 +27,9 @@ import java.io.IOException;
  * #enqueue}. In either case the call can be canceled at any time with {@link #cancel}. A call that
  * is busy writing its request or reading its response may receive a {@link IOException}; this is
  * working as designed.
+ *
+ * <p>Implementation of {@code toString()} for this interface should never include sensitive data
+ * such as query params, headers and so on.
  */
 public interface Call<T> extends Cloneable {
   Response<T> execute() throws IOException;
--- a/retrofit/src/main/java/retrofit/OkHttpCall.java
+++ b/retrofit/src/main/java/retrofit/OkHttpCall.java
@@ -51,7 +51,7 @@ final class OkHttpCall<T> implements Call<T> {
 
   @Override public void enqueue(final Callback<T> callback) {
     synchronized (this) {
-      if (executed) throw new IllegalStateException("Already executed");
+      if (executed) throw new IllegalStateException("Already executed, " + toString());
       executed = true;
     }
 
@@ -59,7 +59,7 @@ final class OkHttpCall<T> implements Call<T> {
     try {
       rawCall = createRawCall();
     } catch (Throwable t) {
-      callback.onFailure(t);
+      callback.onFailure(new RuntimeException("Can not create call for " + toString(), t));
       return;
     }
     if (canceled) {
@@ -103,7 +103,7 @@ final class OkHttpCall<T> implements Call<T> {
 
   public Response<T> execute() throws IOException {
     synchronized (this) {
-      if (executed) throw new IllegalStateException("Already executed");
+      if (executed) throw new IllegalStateException("Already executed, " + toString());
       executed = true;
     }
 
@@ -151,7 +151,8 @@ final class OkHttpCall<T> implements Call<T> {
       // If the underlying source threw an exception, propagate that rather than indicating it was
       // a runtime exception.
       catchingBody.throwIfCaught();
-      throw e;
+      throw new RuntimeException("Problem has occurred during conversion of response for "
+          + toString(), e);
     }
   }
 
@@ -161,6 +162,12 @@ final class OkHttpCall<T> implements Call<T> {
     if (rawCall != null) {
       rawCall.cancel();
     }
+  }
+
+  @Override
+  public String toString() {
+    // Should never include sensitive data such as query params, headers and so on.
+    return requestFactory.toString();
   }
 
   static final class NoContentResponseBody extends ResponseBody {
--- a/retrofit/src/main/java/retrofit/RequestFactoryParser.java
+++ b/retrofit/src/main/java/retrofit/RequestFactoryParser.java
@@ -79,8 +79,8 @@ final class RequestFactoryParser {
   }
 
   private RequestFactory toRequestFactory(BaseUrl baseUrl) {
-    return new RequestFactory(httpMethod, baseUrl, relativeUrl, headers, contentType, hasBody,
-        isFormEncoded, isMultipart, requestBuilderActions);
+    return new RequestFactory(method, httpMethod, baseUrl, relativeUrl, headers, contentType,
+        hasBody, isFormEncoded, isMultipart, requestBuilderActions);
   }
 
   private RuntimeException parameterError(Throwable cause, int index, String message,
--- a/retrofit/src/test/java/retrofit/CallTest.java
+++ b/retrofit/src/test/java/retrofit/CallTest.java
@@ -245,8 +245,11 @@ public final class CallTest {
     });
     assertTrue(latch.await(2, SECONDS));
 
-    assertThat(failureRef.get()).isInstanceOf(UnsupportedOperationException.class)
-        .hasMessage("I am broken!");
+    Throwable failure = failureRef.get();
+
+    assertThat(failure).hasMessage("Can not create call for Service.postString(String), HTTP method = POST, relative path template = /");
+    assertThat(failure.getCause()).hasMessage("I am broken!")
+        .isInstanceOf(UnsupportedOperationException.class);
   }
 
   @Test public void conversionProblemIncomingSync() throws IOException {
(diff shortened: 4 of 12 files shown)


artem-zinnatullin:
Now it also collects method's parameters types — useful for methods with same names, but different parameters. Plus tests for `Util.parseHttpMethodAndRelativePathTemplate()` and `MethodHandler.collectInfoForException()`.

cemo:
:heart: @artem-zinnatullin

retrofit/src/main/java/retrofit/Call.java
```diff
@@ -33,4 +33,5 @@
   void enqueue(Callback<T> callback);
   void cancel();
   Call<T> clone();
+  String infoForException();
```
JakeWharton: I really don't like this name. Is there any reason this shouldn't just be `toString()`? Otherwise I would call this `describe()`.

retrofit/src/main/java/retrofit/MethodHandler.java
```diff
@@ -28,7 +29,9 @@
     Converter<ResponseBody, Object> responseConverter =
         (Converter<ResponseBody, Object>) createResponseConverter(method, retrofit, responseType);
     RequestFactory requestFactory = RequestFactoryParser.parse(method, responseType, retrofit);
-    return new MethodHandler<>(retrofit, requestFactory, callAdapter, responseConverter);
+    String infoForException = collectInfoForException(method);
```
JakeWharton: Yikes. Is there any way to make this lazily computed when needed? I realize we're caching it so the cost is only paid once, but since it's in an exceptional case I don't mind paying the cost of computation every time.

retrofit/src/main/java/retrofit/OkHttpCall.java
```diff
@@ -151,7 +158,8 @@ private void callSuccess(Response<T> response) {
       // If the underlying source threw an exception, propagate that rather than indicating it was
       // a runtime exception.
       catchingBody.throwIfCaught();
-      throw e;
+      throw new RuntimeException("Problem has occurred during conversion of response for "
```
JakeWharton: Not a fan of wrapping here. It hides the underlying error and creates RetrofitError-like semantics all over again.

retrofit/src/main/java/retrofit/Call.java
artem-zinnatullin: Yeah, me too. 

`toString()` will play really nice with IntelliJ's 14 inline debugger comments feature! And it looks like a good choice for cases when user accidentally includes `call` into exception message.

Will rename.

retrofit/src/main/java/retrofit/RequestFactoryParser.java
```diff
@@ -94,25 +92,27 @@ private RuntimeException parameterError(int index, String message, Object... arg
...
-        parseHttpMethodAndPath("PUT", ((PUT) annotation).value(), true);
-      } else if (annotation instanceof HTTP) {
-        HTTP http = (HTTP) annotation;
-        parseHttpMethodAndPath(http.method(), http.path(), http.hasBody());
-      } else if (annotation instanceof Headers) {
+
+        boolean hasBody = annotation instanceof PATCH
+            || annotation instanceof POST
+            || annotation instanceof PUT
+            || (annotation instanceof HTTP && ((HTTP) annotation).hasBody());
```
JakeWharton: Where did HEAD and DELETE go?

retrofit/src/main/java/retrofit/MethodHandler.java
artem-zinnatullin: Will try to make it lazy. But it may require sharing some data: `method` or `MethodHandler` with the `Call`, that's why I didn't go that road in the initial impl.

retrofit/src/main/java/retrofit/RequestFactoryParser.java
JakeWharton: Nevermind. Diff played tricks on me

retrofit/src/main/java/retrofit/MethodHandler.java
JakeWharton: If it's too difficult, I at least think this belongs inside `RequestFactoryParser` and becomes `.describe()` or `.toString()` on the `RequestFactory` object.

JakeWharton: Actually doing that will allow you to pass it down into the `OkHttpCall` easily.

retrofit/src/test/java/retrofit/CallTest.java
```diff
@@ -245,8 +245,11 @@
     });
     assertTrue(latch.await(2, SECONDS));
 
-    assertThat(failureRef.get()).isInstanceOf(UnsupportedOperationException.class)
-        .hasMessage("I am broken!");
+    Throwable failure = failureRef.get();
+
+    assertThat(failure).hasMessage("Can not create call for Service.postString(String), HTTP method = POST, relative path template = /");
+    assertThat(failure).hasCauseExactlyInstanceOf(UnsupportedOperationException.class);
```
JakeWharton: You're already asserting on `getCause()` below, just put a normal `.isInstanceOf` in that chain.

Also, this is a damn ridiculously specific method for AssertJ to have...

retrofit/src/test/java/retrofit/CallTest.java
```diff
@@ -271,8 +274,10 @@
     try {
       call.execute();
       fail();
-    } catch (UnsupportedOperationException e) {
-      assertThat(e).hasMessage("I am broken!");
+    } catch (RuntimeException e) {
+      assertThat(e).hasMessage("Problem has occurred during conversion of response for Service.postString(String), HTTP method = POST, relative path template = /");
+      assertThat(e).hasCauseExactlyInstanceOf(UnsupportedOperationException.class);
```
JakeWharton: Ditto

retrofit/src/test/java/retrofit/CallTest.java
```diff
@@ -357,8 +362,11 @@
     });
     assertTrue(latch.await(2, SECONDS));
 
-    assertThat(failureRef.get()).isInstanceOf(UnsupportedOperationException.class)
-        .hasMessage("I am broken!");
+    Throwable failure = failureRef.get();
+
+    assertThat(failure).hasMessage("Problem has occurred during conversion of response for Service.postString(String), HTTP method = POST, relative path template = /");
+    assertThat(failure).hasCauseExactlyInstanceOf(UnsupportedOperationException.class);
```
JakeWharton: Ditto

retrofit/src/main/java/retrofit/OkHttpCall.java
artem-zinnatullin: Yeah, that's the second thing that I've mentioned in the PR description.

But but, otherwise we won't be able to include useful info into the exception. 

Though we can *try* to modify `exception`'s message via Reflection…

retrofit/src/main/java/retrofit/MethodHandler.java
artem-zinnatullin: Nice, ok.

JakeWharton:
Mostly looks good. My biggest concern is the wrapping of issues. It's probably not the worst thing in the world, but I'll stew on it a night or two.

artem-zinnatullin:
Sure. 

Please think about patching exception's message via reflection (not joking): it saves the type of exception and keeps stacktrace as short as possible. Implementation may be fail-safe: cannot patch — no info will be included.

If we will go with wrapping — looks like it should be one common type of exception and all exceptions should be wrapped into it. Otherwise, it will be hard for the users to understand when they will need to look at the cause and when not. I mean a case when you want to handle a special type of exception programmatically.

artem-zinnatullin:
PR now much cleaner. Moved `infoForException` to `RequestFactory.toString()` and it's now lazy! Reverted changes in other files.

artem-zinnatullin:
Any new thoughts on this?

vincentnock:
This is **really** missing in Retrofit
Any news about that ? (or other workaround to stop having crash logged without any single information about where it comes from)

swankjesse:
This PR has some nice ideas but it’s obsolete.

artem-zinnatullin:
@vincentnock you can achieve similar functionality through custom `CallAdapter.Factory` where you'll receive all interesting info via annotations and pointers to the service method and will be able to customize exception

// whoa, that was 2+ years ago

PR #1198, opened 2015-10-14
small refactor for queryParams

RoyMontoya:
asdfas


JakeWharton:
What is this trying to solve? Query parameters do not require `=` to be present.

PR #1200, opened 2015-10-14
add converter for fastjson

avenwu:
Retrofit has implemented some useful converters as listed on:
https://github.com/square/retrofit/tree/master/retrofit-converters

However the converter for [fastjson](https://github.com/alibaba/fastjson) is not implemented by default, personally I'v used [fastjson](https://github.com/alibaba/fastjson) to deal with json data somtimes. 

I'd like to contribute  the converter-fastjson implemention according to the [guidelines for contributing](https://github.com/square/retrofit/blob/master/CONTRIBUTING.md).

**Reference**
* [https://github.com/alibaba/fastjson](https://github.com/alibaba/fastjson)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,6 +63,7 @@
     <wire.version>1.7.0</wire.version>
     <simplexml.version>2.7.1</simplexml.version>
     <moshi.version>1.0.0</moshi.version>
+    <fastjson.version>1.1.45.android</fastjson.version>
 
     <!-- Test Dependencies -->
     <junit.version>4.12</junit.version>
@@ -147,6 +148,11 @@
         <groupId>com.squareup.moshi</groupId>
         <artifactId>moshi</artifactId>
         <version>${moshi.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>com.alibaba</groupId>
+        <artifactId>fastjson</artifactId>
+        <version>${fastjson.version}</version>
       </dependency>
 
       <dependency>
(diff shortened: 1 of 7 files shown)


JakeWharton:
Retrofit already offers 3 first-party JSON converters. Gson and Jackson are the two leading serializers for which we offer converters, and we include Moshi just because it's a Square library. Similarly, we have two first-party protocol buffer converters, one using the official library and one with Wire because it's a Square library. We have a single XML converter because... well no one really knows.

There are many more eclectic and specialized JSON converters just as there are for other serialization formats. Maintaining a bunch of these is burdensome as we have to track the APIs and changes in 6 external libraries which have nothing to do with Retrofit's core. I can think of at least 5 other JSON libraries that have popped up recently almost always just being built on top of Jackson or Gson.

Retrofit was modified (by me, actually) to be agnostic to converters well before it hit 1.0 a few years ago. The point was that as libraries and formats change in popularity, no changes in Retrofit would be needed to support them. This is very much the case with the forthcoming version 2.0 as well.

As such, we won't be adding any more first-party converters unless a compelling reason can be made as to why we should. Since we already have 3 JSON converters, there is little argument to be made for more.

However, there's no reason that these cannot be turned into third-party libraries that exist on their own. I have made a wiki page for listing third-party converters so that they are able to be discovered in one place.

Thanks for taking the time to implement this, despite that fact that we aren't able to accept it into the main repo.

avenwu:
@JakeWharton Thanks for your detailed explanation.
It's my pleasure to implement this though it'll not be accept into the main repo.
I agree with that maintaining too much converters a bunch of these is burdensome.
Anyway Retrofit is really cool, both 1.x and 2.x :+1: .

PR #1227, opened 2015-10-23
Add option to choose to use execute() or enqueue() in Rx adapter.

adi1133:
The default behaviour remains to use **execute()** with the option to **enqueue()** instead.
My use case for using enqueue is for android apps that have all of their requests initiated from activity lifecycle events.

Related to #1117


JakeWharton:
As I said in the issue, we will not be adding this option. You can use RxJava's subscribers to schedule on what thread the work happens.

PR #1252, opened 2015-10-31
RequestBuilderAction: pass exception from converter as cause of thrown n exception

### LINKED ISSUE #1251: Exception thrown by converter is ignored during request build phase
Relates to #1250.

Whenever the converter throws an `IOException` while converting a parameter (during building a okhttp request), that exception is ignored.

```java
  static final class Body<T> extends RequestBuilderAction {
    private final Converter<T, RequestBody> converter;

    Body(Converter<T, RequestBody> converter) {
      this.converter = converter;
    }

    @Override void perform(RequestBuilder builder, Object value) {
      if (value == null) {
        throw new IllegalArgumentException("Body parameter value must not be null.");
      }
      RequestBody body;
      try {
        //noinspection unchecked
        body = converter.convert((T) value);
      } catch (IOException e) {
        throw new RuntimeException("Unable to convert " + value + " to RequestBody");
      }
      builder.setBody(body);
    }
  }
}
```
A new exception is thrown without wrapping the `cause`. It will be very helpful when debugging converters if we have the full stacktrace.

artem-zinnatullin:
Closes #1251.

I've checked all `catch`es in the Retrofit for the same pattern of not passing real cause to the thrown exception — found only `RequestBuilderAction`. 

@JakeWharton, I also wanted to add tests for  `RequestBuilderAction` but when I started I've found that `RequestBuilder` is not mockable and it requires a lot of parameters for the constructor, and since it's final it's not possible (easily) to verify invocations on it. So, no tests from me today :(

### DIFF
--- a/retrofit/src/main/java/retrofit/RequestBuilderAction.java
+++ b/retrofit/src/main/java/retrofit/RequestBuilderAction.java
@@ -207,7 +207,7 @@ abstract class RequestBuilderAction {
         //noinspection unchecked
         body = converter.convert((T) value);
       } catch (IOException e) {
-        throw new RuntimeException("Unable to convert " + value + " to RequestBody");
+        throw new RuntimeException("Unable to convert " + value + " to RequestBody", e);
       }
       builder.addPart(headers, body);
     }
@@ -250,7 +250,7 @@ abstract class RequestBuilderAction {
         try {
           body = converter.convert(entryValue);
         } catch (IOException e) {
-          throw new RuntimeException("Unable to convert " + entryValue + " to RequestBody");
+          throw new RuntimeException("Unable to convert " + entryValue + " to RequestBody", e);
         }
         builder.addPart(headers, body);
       }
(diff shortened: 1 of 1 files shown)


raniejade:
Why not extract a method from it? They do exactly the same thing.

artem-zinnatullin:
Not sure that one-liner without conditional logic  deserves a method, also
if exceptions messages will be different in future — method will be
redundant :)

On Sat, Oct 31, 2015, 19:39 Ranie Jade Ramiso <notifications@github.com>
wrote:

> Why not extract a method from it? They do exactly the same thing.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/square/retrofit/pull/1252#issuecomment-152747630>.
>
-- 

@artem_zin

JakeWharton:
There are already tests for this behavior that you could add to. This isn't a public API so you shouldn't be writing tests against its API anyway, but at a much higher level.

PR #1263, opened 2015-11-06
Add support for iOS with RoboVM

BlueRiverInteractive:
This PR adds support to have all callbacks run on the iOS main thread when using Retrofit with RoboVM.


JakeWharton:
This looks fine. There just appears to be trailing spaces on some line which is failing the build.

BlueRiverInteractive:
Fixed all issues. All checks pass now.

pom.xml
```diff
@@ -52,6 +52,7 @@
     <android.platform>16</android.platform>
     <okhttp.version>2.5.0</okhttp.version>
     <animal.sniffer.version>1.14</animal.sniffer.version>
+    <robovm.version>1.9.0</robovm.version>
```
JakeWharton: Delete

retrofit/src/main/java/retrofit/Platform.java
```diff
@@ -94,4 +99,37 @@ Object invokeDefaultMethod(Method method, Class<?> declaringClass, Object object
...
+      private static Method addOperation;
+
+      static {
+        try {
+          // queue = NSOperationQueue.getMainQueue();
+          Class<?> operationQueue = Class.forName("org.robovm.apple.foundation.NSOperationQueue");
+          queue = operationQueue.getDeclaredMethod("getMainQueue").invoke(null);
+          addOperation = operationQueue.getDeclaredMethod("addOperation", Runnable.class);
+        } catch (Exception e) {
+          e.printStackTrace();
```
JakeWharton: This should be
```java
throw new AssertionError(e);
```
so that an error is not silently swallowed.

retrofit/src/main/java/retrofit/Platform.java
```diff
@@ -94,4 +99,37 @@ Object invokeDefaultMethod(Method method, Class<?> declaringClass, Object object
...
+        }
+      }
+
+      @Override public void execute(Runnable r) {
+        try {
+          // queue.addOperation(r);
+          addOperation.invoke(queue, r);
+        } catch (Exception ignored) {
```
JakeWharton: Catching `Exception` here means `RuntimeException`s will be swallowed. We should only need to catch `IllegalArgumentException` which can be rethrown wrapped in `AssertionError` and `InvocationTargetException` in which case we can re-throw its `getCause()` wrapped in a `RuntimeException`.

BlueRiverInteractive:
All issues fixed.

retrofit/src/main/java/retrofit/Platform.java
```diff
@@ -39,6 +40,11 @@ private static Platform findPlatform() {
     } catch (ClassNotFoundException ignored) {
     }
     try {
+      Class.forName("org.robovm.apple.foundation.NSObject");
+      return new IOS();
+    } catch (ClassNotFoundException | UnsatisfiedLinkError ignored) {
```
baoti: Could it be compiled on Java 7?

larsgrefer: @baoti Why do you think that it can not be compiled on Java 7?

baoti: Sorry, it's OK!

BlueRiverInteractive:
Merge conflicts have been resolved.

JakeWharton:
Rebased, squashed, and merged as bc7226b0ccc751fc1ee99dd50a389751f4128ced. Sorry for the delay. Thanks for the contribution.

PR #1266, opened 2015-11-09
Make Retrofit ready for usage in OSGi

SimonScholz:
Hello I just created a project, which is supposed to make use of Retrofit in an Eclipse RCP Application.

I did this by converting retrofit and it's dependencies to OSGi artifacts. See https://github.com/SimonScholz/retrofit-osgi

Due to different classloaders for every jar artifact in an OSGi runtime the split packages between retrofit and it's converter projects cause problems in OSGi.
The package private retrofit.Utils class cannot be used properly by the converter classes in an OSGi environment right now. Therefore this patch makes the  closeQuietly method and the Utils class iself public.

In case you apply this patch I can make proper use of the converted OSGi bundles of Retrofit, so the Eclipse community can also make use of Retrofit libraries, which would be awesome.

I also just signed the CLA.

retrofit eclipse 4 sample application _103

### DIFF
--- a/retrofit-converters/gson/src/main/java/retrofit/GsonResponseBodyConverter.java
+++ b/retrofit-converters/gson/src/main/java/retrofit/GsonResponseBodyConverter.java
@@ -32,7 +32,12 @@ final class GsonResponseBodyConverter<T> implements Converter<ResponseBody, T> {
     try {
       return adapter.fromJson(reader);
     } finally {
-      Utils.closeQuietly(reader);
+      if (reader != null) {
+        try {
+          reader.close();
+        } catch (IOException ignored) {
+        }
+      }
     }
   }
 }
--- a/retrofit-converters/jackson/src/main/java/retrofit/JacksonResponseBodyConverter.java
+++ b/retrofit-converters/jackson/src/main/java/retrofit/JacksonResponseBodyConverter.java
@@ -32,7 +32,12 @@ final class JacksonResponseBodyConverter<T> implements Converter<ResponseBody, T
     try {
       return adapter.readValue(reader);
     } finally {
-      Utils.closeQuietly(reader);
+      if (reader != null) {
+        try {
+          reader.close();
+        } catch (IOException ignored) {
+        }
+      }
     }
   }
 }
--- a/retrofit-converters/moshi/src/main/java/retrofit/MoshiResponseBodyConverter.java
+++ b/retrofit-converters/moshi/src/main/java/retrofit/MoshiResponseBodyConverter.java
@@ -32,7 +32,12 @@ final class MoshiResponseBodyConverter<T> implements Converter<ResponseBody, T> 
     try {
       return adapter.fromJson(source);
     } finally {
-      Utils.closeQuietly(source);
+      if (source != null) {
+        try {
+          source.close();
+        } catch (IOException ignored) {
+        }
+      }
     }
   }
 }
--- a/retrofit-converters/protobuf/src/main/java/retrofit/ProtoResponseBodyConverter.java
+++ b/retrofit-converters/protobuf/src/main/java/retrofit/ProtoResponseBodyConverter.java
@@ -37,7 +37,12 @@ final class ProtoResponseBodyConverter<T extends MessageLite>
     } catch (InvalidProtocolBufferException e) {
       throw new RuntimeException(e); // Despite extending IOException, this is data mismatch.
     } finally {
-      Utils.closeQuietly(is);
+      if (is != null) {
+        try {
+          is.close();
+        } catch (IOException ignored) {
+        }
+      }
     }
   }
 }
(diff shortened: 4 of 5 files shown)


JakeWharton:
There's no reason for this to be in the public API. This is an internal class for a reason. If this is all that prevents OSGi, just inline the implementation to consumers outside the main artifact.

SimonScholz:
In case the Utils class is package private I get an
java.lang.IllegalAccessError: tried to access class retrofit.Utils from class retrofit.GsonResponseBodyConverter
	at retrofit.GsonResponseBodyConverter.convert(GsonResponseBodyConverter.java:38)
	at retrofit.GsonResponseBodyConverter.convert(GsonResponseBodyConverter.java:24)
	at retrofit.OkHttpCall.parseResponse(OkHttpCall.java:148)
	at retrofit.OkHttpCall.execute(OkHttpCall.java:116)
	at de.simonscholz.retrofit.app.job.JsonContributorsJob.run(JsonContributorsJob.java:49)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:55)

when running the converted OSGi bundles in an Eclipse RCP application.

The  closeQuietly method does not do a lot of magic. So you mean I should place the implementation in the converters and provide another pull request or what do you mean by "consumers outside the main artifact."?

JakeWharton:
Yes. The converters who use this method can just do the try/catch
themselves instead of calling through.

SimonScholz:
I just updated the PR, so that the converters themselves handle the closeQuietly logic.
And I also tested this locally on my maschine and it simply works.
Once this PR is merged and available on maven central I can convert retrofit properly to OSGi and create p2 repositories for Eclipse, so that it can be used easily within an Eclipse RCP Application.

retrofit-converters/gson/src/main/java/retrofit/GsonResponseBodyConverter.java
```diff
@@ -32,7 +32,12 @@
     try {
       return adapter.fromJson(reader);
     } finally {
-      Utils.closeQuietly(reader);
+       if (reader != null) {
+            try {
```
JakeWharton: two space indent

retrofit-converters/jackson/src/main/java/retrofit/JacksonResponseBodyConverter.java
```diff
@@ -32,7 +32,12 @@
     try {
       return adapter.readValue(reader);
     } finally {
-      Utils.closeQuietly(reader);
+      if (reader != null) {
+          try {
```
JakeWharton: two space indent

retrofit-converters/moshi/src/main/java/retrofit/MoshiResponseBodyConverter.java
```diff
@@ -32,7 +32,12 @@
     try {
       return adapter.fromJson(source);
     } finally {
-      Utils.closeQuietly(source);
+        if (source != null) {
```
JakeWharton: two space indent

retrofit-converters/protobuf/src/main/java/retrofit/ProtoResponseBodyConverter.java
```diff
@@ -37,7 +37,12 @@
     } catch (InvalidProtocolBufferException e) {
       throw new RuntimeException(e); // Despite extending IOException, this is data mismatch.
     } finally {
-      Utils.closeQuietly(is);
+        if (is != null) {
```
JakeWharton: two space indent

retrofit-converters/wire/src/main/java/retrofit/WireResponseBodyConverter.java
```diff
@@ -30,11 +30,16 @@
   }
 
   @Override public T convert(ResponseBody value) throws IOException {
+    BufferedSource source = value.source();
```
JakeWharton: calling `value.source()` can throw, so this will need split into two lines, variable initialization and retrieving the source–the latter of which is inside the `try` block

retrofit-converters/wire/src/main/java/retrofit/WireResponseBodyConverter.java
```diff
@@ -30,11 +30,16 @@
...
     try {
-      BufferedSource source = value.source();
       return adapter.decode(source);
     } finally {
-      Utils.closeQuietly(value);
+        if (source != null) {
```
JakeWharton: two space indent

SimonScholz:
Thank you for the feedback. I just adopted your suggestions.

JakeWharton:
Awesome. Thanks for updating.

SimonScholz:
Thanks a bunch for merging it. I'd really love to use retrofit in my Eclipse RCP applications.

JakeWharton:
Work should be picking back up on Retrofit now so there will be a new
release with these changes in the coming weeks.

Issue #1272, opened 2015-11-10
support etcd restful style, path containing charactor '/'


yuankui:
When I use retrofit to implement a interface for etcd
https://coreos.com/etcd/docs/latest/api.html#setting-the-value-of-a-key

`curl http://127.0.0.1:2379/v2/keys/message -XPUT -d value="Hello world"`

where I declaire a method like
```
@PUT("/v2/keys/{key}")
    @FormUrlEncoded
    Call<Result> set(@Path(value = "key", encoded = true) String key, @Field("value")String value, @Field("ttl") int ttl);
```

when i invoke `etcdService.set("/test/key", "testValue")`, It still urlencoded my key although I declaired with `@Path(value = "key", encoded = true)`

I checked the source code and found It has a `PATH_SEGMENT_ENCODE_SET = " \"<>^`{}|/\\?#"` in which chars are always encoded.

But In the case I cannot do it with the `/` char encode.

So I propose to remove `/` from the PATH_SEGMENT_ENCODE_SET.

yuankui:
https://github.com/square/retrofit/pull/1273

swankjesse:
Would this work?

```
@PUT("/v2/keys/{key1}/{key2}")
    @FormUrlEncoded
    Call<Result> set(
        @Path(value = "key1") String key1,
        @Path(value = "key2") String key2,
        @Field("value")String value, @Field("ttl") int ttl);

```
etcdService.set("test", "key", "testValue")
```

The `@Path` annotation matches a single path segment.

**Comment by JakeWharton** [comment]
There's also @Url which lets you specify an entire relative (or absolute)
URL as a parameter to the method.

**Comment by JakeWharton** [comment]
```java
@PUT
@FormUrlEncoded
Call<Result> set(
    @Url String url,
    @Field("value") String value,
    @Field("ttl") int ttl);
```
```java
String value = "/test/key";
service.set("keys/" + value, "some", "value");
```

yuankui:
@swankjesse that will not cover /test/it/that, and /test/it/and/this, ie, the path deep is not always the same.

@JakeWharton that may be a solution. But I have two concat the prefix& key like `/v2/keys/$key` in every  when ever I invoke this service, That's not elegant, I think.

for me, I would prefer applying the the patch above, and will wait for your reconstruction.

JakeWharton:
[...]

Issue #1129, opened 2015-09-26
Use more sophisticated path replacement leveraging HttpUrl's native encoding


JakeWharton:
And delete all the custom path encoding introduced in #1123.

As we discussed. Just creating this as a tracking issue.

NightlyNexus:
May I ask what thoughts were here?
Copy HttpUrl's `canonicalize` implementation?
Refactor the usage of `relativeUrl`?

JakeWharton:
The implementation is already copied, that's what I'm trying to delete. The idea is that we `resolve()` the relative URL immediately on `RequestBuilder` creation and then when doing path replacement we read the segment string, replace the value, and set that segment back on the builder. This pushes all encoding responsibility into `HttpUrl`.

JakeWharton:
We have determined that, for the time being, there's no feasible way to work this into OkHttp's API. Some details are at https://github.com/square/okhttp/issues/2171. The encoding rules will remain duplicated into Retrofit.

PR #1273, opened 2015-11-10
remove '/' in PATH_SEGMENT_ENCODE_SET for etcd

yuankui:


yuankui:
https://github.com/square/retrofit/issues/1272

JakeWharton:
This code is eventually going to be deleted in favor of OkHttp's encoding. See #1129 for tracking that work. As such, we cannot change the logic here.

yuankui:
hi, @JakeWharton, I am more care about when It will release.
as for me, I think I must change this logic myself,orelse we cannot use it for query etcd.

PR #1274, opened 2015-11-10
Fix for when no 'Content-Length' header is returned on 204 response

marius-bardan:
According to http://tools.ietf.org/html/rfc7230#section-3.3.2 :

> A server MUST NOT send a Content-Length header field in any response with a status code of 1xx (Informational) or 204 (No Content).

However, having the server return a 204 response with no 'Content-Length' header yields in a retrofit conversion error, even if 'Void' type is being used as the expected body type.

### DIFF
--- a/retrofit/src/main/java/retrofit/client/OkClient.java
+++ b/retrofit/src/main/java/retrofit/client/OkClient.java
@@ -95,7 +95,7 @@ public class OkClient implements Client {
   }
 
   private static TypedInput createResponseBody(final ResponseBody body) {
-    if (body.contentLength() == 0) {
+    if (body.contentLength() <= 0) {
       return null;
     }
     return new TypedInput() {


swankjesse:
This won't work. We use `-1` to indicate that the length is unknown, so this will likely break many applications. Can you fix your server to follow the spec?

marius-bardan:
@swankjesse at which spec are you referring to? From what I posted from the RFC, 'Content-Length' should not be sent by the server if 204 status code is returned. 
My server follows the RFC and does not return 'Content-Length', breaking retrofit with a conversion error, because of that '==' check.

JakeWharton:
This is already fixed on master for v2.0. 204s and 205s will skip the
converter and return null or throw an exception if a non-empty body is
returned.

marius-bardan:
@JakeWharton so there will be no more bug-fixes merged into 1.9? Upgrading to 2.0 means a lot of work right now.

JakeWharton:
No. There's not enough resources to work on 2.0 let alone continue 1.x.

PR #1277, opened 2015-11-12
Add CA.F sample showing an all-custom type and its callback.

### LINKED ISSUE #1102: How do we create ErrorHandler in retrofit 2.0?
Hi
It seems ErrorHandler is removed in 2.0 and we need to use CallAdapter instead. Is there any examples to look at?

Cheers,

JakeWharton:
This also remedies some code and configuration problems which prevented the previous one for ListenableFuture from working.

For lack of response from the original poster, closes #1102.

### DIFF
--- a/samples/src/main/java/com/example/retrofit/ErrorHandlingCallAdapter.java
+++ b/samples/src/main/java/com/example/retrofit/ErrorHandlingCallAdapter.java
@@ -1,0 +1,173 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.example.retrofit;
+
+import com.google.common.reflect.TypeToken;
+import java.io.IOException;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import retrofit.Call;
+import retrofit.CallAdapter;
+import retrofit.Callback;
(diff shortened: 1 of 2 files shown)


samples/src/main/java/com/example/retrofit/ErrorHandlingCallAdapter.java
```diff
@@ -0,0 +1,172 @@
...
+import retrofit.Retrofit;
+import retrofit.http.GET;
+
+/**
+ * A sample showing a custom {@link CallAdapter} which adapts the built-in {@link Call} to a custom
+ * version whose callback has more granular methods.
+ */
+public final class ErrorHandlingCallAdapter {
+  /** A callback which offers granular callbacks for various conditions. */
+  interface MyCallback<T> {
```
JakeWharton: Fun fact: this was the built-in Callback API when I first started working on the project in 2012: https://github.com/square/retrofit/blob/d8fab22ab289f24caf73e2af9d486b92afdad490/core/src/main/java/retrofit/core/Callback.java

JakeWharton:
This is based on #1276 which has an API change. Will rebase once that's merged as I'm too lazy to deal with the conflict churn.

samples/src/main/java/com/example/retrofit/ErrorHandlingCallAdapter.java
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        @Override public <R> MyCall<R> adapt(Call<R> call) {
+          return new MyCallWrapper<>(call);
+        }
+      };
+    }
+  }
+
+  static class MyCallWrapper<T> implements MyCall<T> {
```
swankjesse: Wrapper –> Adapter?

samples/src/main/java/com/example/retrofit/ErrorHandlingCallAdapter.java
```diff
@@ -0,0 +1,172 @@
...
+    /** Called for 401 responses. */
+    void unauthenticated(Response<?> response);
+    /** Called for [400, 500) responses, except 401. */
+    void clientError(Response<?> response);
+    /** Called for [500, 600) response. */
+    void serverError(Response<?> response);
+    /** Called for network errors while making the call. */
+    void networkError(IOException e);
+    /** Called for unexpected errors while making the call. */
+    void unexpectedError(Throwable t);
```
swankjesse: Bravo on this interface, BTW

swankjesse:
LGTM

PR #1283, opened 2015-11-15
Fill out some documentation on public types.

JakeWharton:


retrofit/src/main/java/retrofit/Call.java
```diff
@@ -27,10 +27,31 @@
  * #enqueue}. In either case the call can be canceled at any time with {@link #cancel}. A call that
  * is busy writing its request or reading its response may receive a {@link IOException}; this is
  * working as designed.
+ *
+ * @param <T> Response body type.
```
swankjesse: _successful_ response body type ?

retrofit/src/main/java/retrofit/Call.java
```diff
@@ -27,10 +27,31 @@
...
+ *
+ * @param <T> Successful response body type.
  */
 public interface Call<T> extends Cloneable {
+  /**
+   * Synchronously send the request and return its response.
+   *
+   * @throws IOException if a problem occurred talking to the server.
+   * @throws RuntimeException (and subclasses) if an unexpected error occurs creating the request
+   * or processing the response.
```
swankjesse: _processing_ . . . should this just be _decoding_ ? The word _processing_ sounds like business logic, but I think we just do something more mechanical?

retrofit/src/main/java/retrofit/Call.java
```diff
@@ -27,10 +27,31 @@
...
+   */
   void enqueue(Callback<T> callback);
+
+  /**
+   * Cancel this call. An attempt will be made to cancel in-flight calls, and if the call has not
+   * yet been executed it never will be.
+   */
   void cancel();
+
+  /** Create a new, identical call to this one. */
```
swankjesse: ... the cloned call can be enqueued or executed even if this call has already been...

swankjesse:
Nice.

PR #1284, opened 2015-11-15
Documentation for endpoint resolving on a base URL.

### LINKED ISSUE #1049: Force trailing slash on base URL
You never want http://example.com/foo but rather http://example.com/foo/ to ensure the relative URL is resolved as a child rather than as a sibling of that path.

JakeWharton:
Closes #1049.

### DIFF
--- a/retrofit/src/test/java/retrofit/RetrofitTest.java
+++ b/retrofit/src/test/java/retrofit/RetrofitTest.java
@@ -479,6 +479,22 @@ public final class RetrofitTest {
     }
   }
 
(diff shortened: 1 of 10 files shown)


retrofit/src/test/java/retrofit/RetrofitTest.java
```diff
@@ -479,6 +479,22 @@
     }
   }
 
+  @Test public void baseUrlNoTrailingSlashThrows() {
```
swankjesse: Nice.

PR #1287, opened 2015-11-16
Cache stateless built-in converter instances.

JakeWharton:


retrofit/src/main/java/retrofit/BuiltInConverters.java
```diff
@@ -25,21 +25,40 @@
 import static retrofit.Utils.closeQuietly;
 
 final class BuiltInConverters extends Converter.Factory {
+  private volatile OkHttpResponseBodyConverter streamingResponseBodyConverter;
```
swankjesse: Why volatile ? it doesn't guarantee you single-instances, and it harms your performance

swankjesse:
LGTM

retrofit/src/main/java/retrofit/BuiltInConverters.java
JakeWharton: Sure. Worst case (which is practically impossible) would be every endpoint called on a different thread at the same time which would each create their own instance, but that's no worse than the current behavior which does that anyway!

PR #1294, opened 2015-11-18
Add string converter support for all non-body annotation types.

### LINKED ISSUE #816: Argument pre-processor, deserialized post-processor.
Consumers of the library want things that we cannot provide or simply will never provide. Examples are: validation of `@Nullable` or other arbitrary annotations for parameters, supporting non-body conversion of third-party library types (e.g., `Multimap`) or with custom rules (e.g., `@SerializedName` on enum constants), and custom handling of response data.

Supporting this isn't certain, and unlike to be in 2.0, but it would be a re-usable extension point which potentially would solve a lot of future cases. Perhaps this could obviate the half-assed `RequestInterceptor` that we have now.

JakeWharton:
This also changes body-based annotation types to go through the request body converter at parsing time.

Closes #816.

### DIFF
--- a/retrofit/src/main/java/retrofit/Converter.java
+++ b/retrofit/src/main/java/retrofit/Converter.java
@@ -20,6 +20,15 @@ import com.squareup.okhttp.ResponseBody;
 import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
+import retrofit.http.Body;
+import retrofit.http.Field;
+import retrofit.http.FieldMap;
+import retrofit.http.Header;
+import retrofit.http.Part;
+import retrofit.http.PartMap;
+import retrofit.http.Path;
+import retrofit.http.Query;
+import retrofit.http.QueryMap;
 
 /**
  * Convert objects to and from their representation in HTTP. Register a converter with Retrofit
@@ -32,17 +41,32 @@ public interface Converter<F, T> {
   abstract class Factory {
     /**
      * Returns a {@link Converter} for converting an HTTP response body to {@code type}, or null if
-     * {@code type} cannot be handled by this factory.
+     * {@code type} cannot be handled by this factory. This is used to create converters for
+     * response types such as {@code SimpleResponse} from a {@code Call<SimpleResponse>}
+     * declaration.
      */
-    public Converter<ResponseBody, ?> fromResponseBody(Type type, Annotation[] annotations) {
+    public Converter<ResponseBody, ?> responseBodyConverter(Type type, Annotation[] annotations) {
       return null;
     }
 
     /**
      * Returns a {@link Converter} for converting {@code type} to an HTTP request body, or null if
-     * {@code type} cannot be handled by this factory.
+     * {@code type} cannot be handled by this factory. This is used to create converters for types
+     * specified by {@link Body @Body}, {@link Part @Part}, and {@link PartMap @PartMap}
+     * values.
      */
-    public Converter<?, RequestBody> toRequestBody(Type type, Annotation[] annotations) {
+    public Converter<?, RequestBody> requestBodyConverter(Type type, Annotation[] annotations) {
+      return null;
+    }
+
+    /**
+     * Returns a {@link Converter} for converting {@code type} to a {@link String}, or null if
+     * {@code type} cannot be handled by this factory. This is used to create converters for types
+     * specified by {@link Field @Field}, {@link FieldMap @FieldMap} values,
+     * {@link Header @Header}, {@link Path @Path}, {@link Query @Query}, and
+     * {@link QueryMap @QueryMap} values.
+     */
+    public Converter<?, String> stringConverter(Type type, Annotation[] annotations) {
       return null;
     }
   }
--- a/retrofit/src/main/java/retrofit/RequestAction.java
+++ b/retrofit/src/main/java/retrofit/RequestAction.java
@@ -18,18 +18,17 @@ package retrofit;
 import com.squareup.okhttp.Headers;
 import com.squareup.okhttp.RequestBody;
 import java.io.IOException;
-import java.lang.annotation.Annotation;
 import java.lang.reflect.Array;
 import java.util.Map;
 
 import static retrofit.Utils.checkNotNull;
 
 abstract class RequestAction<T> {
-  abstract void perform(RequestBuilder builder, T value);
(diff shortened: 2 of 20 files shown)


JakeWharton:
The behavior here is good. There's some tests and de-duplication of parsing code that should happen in a follow-up.

retrofit/src/main/java/retrofit/Converter.java
```diff
@@ -40,10 +51,23 @@
...
     }
+
+    /**
+     * Returns a {@link Converter} for converting {@code type} to a {@link String}, or null if
+     * {@code type} cannot be handled by this factory. This is used to create converters for types
+     * specified by {@link Field @Field}, {@link FieldMap @FieldMap} values,
+     * {@link Header @Header}, {@link Path @Path}, {@link Query @Query}, and
+     * {@link QueryMap @QueryMap} values.
+     */
+    public Converter<?, String> toString(Type type, Annotation[] annotations) {
```
JakeWharton: This name sucks.

swankjesse: Yeah, pretty badly. Anything in `Object` is going to cause problems with nested classes, etc.

swankjesse: requestBodyConverter / responseBodyConverter / stringConverter
requestBody / responseBody / string

retrofit/src/main/java/retrofit/RequestAction.java
```diff
@@ -215,16 +224,7 @@
...
-        Converter<Object, RequestBody> converter =
-            retrofit.requestConverter(entryClass, annotations);
-        RequestBody body;
-        try {
-          body = converter.convert(entryValue);
-        } catch (IOException e) {
-          throw new RuntimeException("Unable to convert " + entryValue + " to RequestBody", e);
-        }
-        builder.addPart(headers, body);
+        builder.addPart(headers, valueConverter.convert(entryValue));
```
swankjesse: nice simplification!

retrofit/src/main/java/retrofit/RequestFactoryParser.java
```diff
@@ -251,57 +252,121 @@ private void parseParameters(Retrofit retrofit) {
...
           } else if (methodParameterAnnotation instanceof Header) {
             Header header = (Header) methodParameterAnnotation;
+            String name = header.value();
 
             Class<?> rawParameterType = Utils.getRawType(methodParameterType);
-            RequestAction<?> headerAction = new RequestAction.Header(header.value());
             if (Iterable.class.isAssignableFrom(rawParameterType)) {
-              headerAction = headerAction.iterable();
+              if (!(methodParameterType instanceof ParameterizedType)) {
+                throw parameterError(i, rawParameterType.getSimpleName()
```
swankjesse: nit: this might not actually be the element type.

```
class Foo<T> extends AbstractList<String> { ... }
```

swankjesse: https://github.com/square/moshi/blob/master/moshi/src/main/java/com/squareup/moshi/Types.java#L193

retrofit/src/main/java/retrofit/RequestFactoryParser.java
```diff
@@ -251,57 +252,121 @@ private void parseParameters(Retrofit retrofit) {
...
+                    + "<String>)");
+              }
+              ParameterizedType parameterizedType = (ParameterizedType) methodParameterType;
+              Type iterableType = Utils.getParameterUpperBound(0, parameterizedType);
+              Converter<?, String> valueConverter =
+                  retrofit.stringConverter(iterableType, methodParameterAnnotations);
+              action = new RequestAction.Header<>(name, valueConverter).iterable();
             } else if (rawParameterType.isArray()) {
-              headerAction = headerAction.array();
+              Class<?> arrayType = rawParameterType.getComponentType(); // TODO box
```
swankjesse: arrayType -> arrayComponentType ?

swankjesse: is boxing actually necessary?

retrofit/src/test/java/retrofit/RetrofitTest.java
```diff
@@ -291,12 +295,64 @@
...
+      @Override public Converter<?, String> toString(Type type, Annotation[] annotations) {
+        return null;
+      }
+    }
+    Retrofit retrofit = new Retrofit.Builder()
+        .baseUrl(server.url("/"))
+        .addConverterFactory(new MyConverterFactory())
+        .build();
+    CallMethod service = retrofit.create(CallMethod.class);
+    service.queryObject(null);
```
swankjesse: should this assert someting?

swankjesse:
LGTM.

retrofit/src/main/java/retrofit/RequestFactoryParser.java
JakeWharton: That would be defined as `Foo<String>` on the parameter though. We forbid wildcards and type variables (or at least we do on body and return type).

retrofit/src/main/java/retrofit/RequestFactoryParser.java
JakeWharton: I think so, yeah. We use `Array.get(Object, int)` to get items out which returns `Object` and thus boxes values from, say, `int[]`. So we don't want converters to be queried for `int.class` (which is impossible to return a generic for anyway) yet receive `Integer.class` instances.

retrofit/src/main/java/retrofit/RequestFactoryParser.java
JakeWharton: I added enforcement and tests for unresolvable parameter types.

retrofit/src/test/java/retrofit/RetrofitTest.java
JakeWharton: Yeah. I'll assert the `Call` is non-null and that the supplied factory was called.

retrofit/src/main/java/retrofit/RequestFactoryParser.java
JakeWharton: Actually, do we even want to let you override primitive behavior?

JakeWharton: Punting that choice & testing it: https://github.com/square/retrofit/issues/1301

PR #1302, opened 2015-11-19
Expose isCanceled on Call. Do not deliver canceled calls on callback executor.

### LINKED ISSUE #1135: The callback Executor should check for cancellation before delivering the result

JakeWharton:
Closes #1135. Closes #1136.

### DIFF
--- a/retrofit/src/main/java/retrofit/ExecutorCallAdapterFactory.java
+++ b/retrofit/src/main/java/retrofit/ExecutorCallAdapterFactory.java
@@ -47,7 +47,6 @@ final class ExecutorCallAdapterFactory implements CallAdapter.Factory {
   static final class ExecutorCallbackCall<T> implements Call<T> {
     private final Executor callbackExecutor;
     private final Call<T> delegate;
-    private volatile boolean canceled;
 
     ExecutorCallbackCall(Executor callbackExecutor, Call<T> delegate) {
       this.callbackExecutor = callbackExecutor;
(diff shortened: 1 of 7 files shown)


retrofit/src/main/java/retrofit/ExecutorCallAdapterFactory.java
```diff
@@ -53,8 +53,29 @@
...
     }
 
-    @Override public void enqueue(Callback<T> callback) {
-      delegate.enqueue(new ExecutorCallback<>(callbackExecutor, callback));
+    @Override public void enqueue(final Callback<T> callback) {
+      delegate.enqueue(new Callback<T>() {
+        @Override public void onResponse(final Response<T> response) {
+          callbackExecutor.execute(new Runnable() {
+            @Override public void run() {
+              if (delegate.isCanceled()) {
```
swankjesse: interestingly, this only happens if the cancel is received `delegate.enqueue()` and `run()`. I suspect it'll also work without this case.

1zaman: @swankjesse: Yes, that was the [issue](https://github.com/square/retrofit/issues/1135) that my commit was designed to handle in #1136 :wink:

PR #1316, opened 2015-11-24
为Call添加CacheControl参数,允许在请求的时候指定缓存策略.

zawn:
In some scenarios we need to refresh the data, such as Android's ListView, it should provide an entry for the operation


zawn:
Sorry, did not submit a change of unit testing.

PR #1318, opened 2015-11-24
为Call添加CacheControl参数,允许在请求的时候指定缓存策略.

zawn:
In some scenarios we need to refresh the data, such as Android's ListView, it should provide an entry for the operation

### DIFF
--- a/retrofit-mock/src/main/java/retrofit/mock/BehaviorCall.java
+++ b/retrofit-mock/src/main/java/retrofit/mock/BehaviorCall.java
@@ -22,6 +22,8 @@ import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicReference;
+
+import com.squareup.okhttp.CacheControl;
 import retrofit.Call;
 import retrofit.Callback;
 import retrofit.Response;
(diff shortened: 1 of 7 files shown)


retrofit-mock/src/main/java/retrofit/mock/BehaviorCall.java
```diff
@@ -58,7 +60,7 @@
     return new BehaviorCall<>(behavior, backgroundExecutor, callbackExecutor, delegate.clone());
   }
 
-  @Override public void enqueue(final Callback<T> callback) {
+  @Override public void enqueue(final Callback<T> callback,CacheControl cacheControl) {
```
baoti: Needs reformat

swankjesse:
I don't think this is the right API for this problem. Instead use the existing headers annotations for the `"Cache-Control"` request header. The `CacheControl` class is just a convenient way to prepare that header.

zawn:
We also would like to do but can not be changed dynamically annotation，As for the pull-down refresh this commonly used functions, we do not want to write the same two basic functions, just different ways annotation

JakeWharton:
You can change it dynamically with a `@Header("Cache-Control") String cacheControl` parameter on the method.

JakeWharton:
In fact, you can even use the `CacheControl` type as the parameter type since its `toString()` renders to the header value.

zawn:
Yes, this method is possible。

zawn:
A lot of time we just want to add CacheControl.FORCE NETWORK or CacheControl.FORCE CACHE, we feel that this should not be associated with the interface, after all this is just a different scenario your local use.

JakeWharton:
That same argument could be made for _any_ part of the HTTP request itself though. Whether it's cache control, another header, or a query parameter, inevitably someone can come up with a use for adding these after-the-fact.

The way we think about it is that the interface method's parameters are the contents of the HTTP request itself (of which Cache-Control is a member), and the `Call` is the execution mechanism. These two things have no idea about each other and they can't interact with each other.

zawn:
Understand, thank you

Issue #1315, opened 2015-11-23
Support for chunked transfer encoding


baev:
I guess there is no way to use chunked transfer encoding. My service looks like this:

```java
public interface UploadService {

    @PUT("/upload/{id}")
    Response upload(@Path("id") long id, @Body TypedFile file);

}
```

JakeWharton:
I think you can add `@Headers("Transfer-Encoding: chunked")` to the method and OkHttp will automatically honor the request for a chunked transfer.

baev:
@JakeWharton thanks for the answer. But it seems it doesn't work.

I simply replace the 

```java
try (CloseableHttpClient client = HttpClients.createDefault();
     InputStream in = Files.newInputStream(file)) {
    HttpPut put = new HttpPut(ENDPOINT + "/upload/" + id);
    InputStreamEntity reqEntity = new InputStreamEntity(in);
    reqEntity.setChunked(true);
    put.setEntity(reqEntity);

    try (CloseableHttpResponse response = client.execute(put)) {
        //process response
    }
} 
```

with

```java
public interface UploadService {

    @PUT("/upload/{id}")
    @Headers({"Transfer-Encoding: chunked"})
    Response upload(@Path("id") long id, @Body TypedFile file);

}

...

UploadService service = new RestAdapter.Builder()
                .setEndpoint(ENDPOINT)
                .build()
                .create(UploadService.class);
service.upload(id, new TypedFile("application/octet-stream", file.toFile()));
```

and its fail with the following error:

`retrofit.RetrofitError: java.io.IOException: Error writing request body to server`

JakeWharton:
What's the whole trace?

baev:
```
Caused by: java.io.IOException: Error writing request body to server
	at sun.net.www.protocol.http.HttpURLConnection$StreamingOutputStream.checkError(HttpURLConnection.java:3478)
	at sun.net.www.protocol.http.HttpURLConnection$StreamingOutputStream.write(HttpURLConnection.java:3461)
	at retrofit.mime.TypedFile.writeTo(TypedFile.java:78)
	at retrofit.client.UrlConnectionClient.prepareRequest(UrlConnectionClient.java:66)
	at retrofit.client.UrlConnectionClient.execute(UrlConnectionClient.java:37)
	at retrofit.RestAdapter$RestHandler.invokeRequest(RestAdapter.java:322)
	... 32 more
```

JakeWharton:
[...]

PR #1319, opened 2015-11-24
Allow converters to delegate to others.

JakeWharton:
Add a sample for this in the form of one that forces chunked uploads when `@Chunked` is added to `@Body`.

Refs #1315.

### DIFF
--- a/retrofit/src/main/java/retrofit/Retrofit.java
+++ b/retrofit/src/main/java/retrofit/Retrofit.java
@@ -225,13 +225,25 @@ public final class Retrofit {
    *
    * @throws IllegalArgumentException if no converter available for {@code type}.
    */
-  public <T> Converter<T, RequestBody> requestConverter(Type type, Annotation[] annotations) {
+  public <T> Converter<T, RequestBody> requestBodyConverter(Type type, Annotation[] annotations) {
+    return nextRequestBodyConverter(null, type, annotations);
+  }
+
+  /**
+   * Returns a {@link Converter} for {@code type} to {@link RequestBody} from the available
+   * {@linkplain #converterFactories() factories} except {@code skipPast}.
+   *
+   * @throws IllegalArgumentException if no converter available for {@code type}.
+   */
+  public <T> Converter<T, RequestBody> nextRequestBodyConverter(Converter.Factory skipPast,
+      Type type, Annotation[] annotations) {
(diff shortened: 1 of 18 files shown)


retrofit/src/main/java/retrofit/Retrofit.java
```diff
@@ -225,13 +225,25 @@ public BaseUrl baseUrl() {
...
+   *
+   * @throws IllegalArgumentException if no converter available for {@code type}.
+   */
+  public <T> Converter<T, RequestBody> nextRequestBodyConverter(Converter.Factory skipPast,
+      Type type, Annotation[] annotations) {
     checkNotNull(type, "type == null");
     checkNotNull(annotations, "annotations == null");
 
-    for (int i = 0, count = converterFactories.size(); i < count; i++) {
+    int start = converterFactories.indexOf(skipPast) + 1;
```
swankjesse: This Just Works with -1!

samples/src/main/java/com/example/retrofit/ChunkingConverter.java
```diff
@@ -0,0 +1,129 @@
...
+import retrofit.Converter;
+import retrofit.GsonConverterFactory;
+import retrofit.Retrofit;
+import retrofit.http.Body;
+import retrofit.http.POST;
+
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+public final class ChunkingConverter {
```
swankjesse: This is a craxy example.

swankjesse: (I think it's a bit distracting; normals shouldn't need to do this)

JakeWharton: Yeah I wrote it for myself to prove the implementation and figured I'd include it in the PR for kicks. Happy to remove, but any idea for other examples?

swankjesse: Luhny-bin? Rewrite a text file to replace all ascii characters and digits with X ?

swankjesse: (that'd be a String to String conversion)

PR #1326, opened 2015-11-25
Support iterable and array types for `@Part`.

### LINKED ISSUE #685: Passing array of values which have the same parameter name in Multipart request

JakeWharton:
Closes #685.

### DIFF
--- a/retrofit/src/main/java/retrofit/RequestFactoryParser.java
+++ b/retrofit/src/main/java/retrofit/RequestFactoryParser.java
@@ -405,15 +405,31 @@ final class RequestFactoryParser {
             com.squareup.okhttp.Headers headers = com.squareup.okhttp.Headers.of(
                 "Content-Disposition", "form-data; name=\"" + part.value() + "\"",
                 "Content-Transfer-Encoding", part.encoding());
-            Converter<?, RequestBody> converter;
-            try {
-              converter =
+
+            Class<?> rawParameterType = Utils.getRawType(methodParameterType);
+            if (Iterable.class.isAssignableFrom(rawParameterType)) {
+              if (!(methodParameterType instanceof ParameterizedType)) {
+                throw parameterError(i, rawParameterType.getSimpleName()
+                    + " must include generic type (e.g., "
+                    + rawParameterType.getSimpleName()
+                    + "<String>)");
+              }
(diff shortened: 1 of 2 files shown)


retrofit/src/main/java/retrofit/RequestFactoryParser.java
```diff
@@ -405,15 +405,31 @@ private void parseParameters(Retrofit retrofit) {
             com.squareup.okhttp.Headers headers = com.squareup.okhttp.Headers.of(
                 "Content-Disposition", "form-data; name=\"" + part.value() + "\"",
                 "Content-Transfer-Encoding", part.encoding());
-            Converter<?, RequestBody> converter;
-            try {
-              converter =
+
+            Class<?> rawParameterType = Utils.getRawType(methodParameterType);
+            if (Iterable.class.isAssignableFrom(rawParameterType)) {
+              if (!(methodParameterType instanceof ParameterizedType)) {
```
swankjesse: Should these 3 likes have a helper method like getIterableElementType, that throws appropriately?

JakeWharton: Yep. This code is duplicated 4 times and I plan to dedupe!

iwo: Deduped in https://github.com/iwo/retrofit/commit/a58dd5985a625569f10bcaec53581ad81c4b0212

PR #1342, opened 2015-12-05
New package name for version 2.

### LINKED ISSUE #1209: Existing libraries using retrofit1 crash when main project uses retrofit2
This happened while including Twitterkit in an Android project that uses retrofit2.

These is the relative line in gradle:

    compile('com.twitter.sdk.android:twitter:1.8.0@aar') {
        transitive = true
    }

Twitterkit uses an old version of retrofit1 and therefore as soon as one initializes the Twitter() object in it the whole app will crash due to the incompatible API across the versions.

I'm not sure how to work around this, but maybe changing retrofit2's full class path (because it indeeds breaks compatibility with v1) could allow having both versions of the library working.

Of course library developers should upgrade the retrofit version in their libraries but what if they don't? And what if a library uses retrofit2 and the main projects is still using retrofit1?

Thanks,
Bye,
Marco

JakeWharton:
Closes #1209.


swankjesse:
I'm in.

PR #1343, opened 2015-12-05
Introduce new mock pattern with a behavior delegate.

### LINKED ISSUE #1286: Add AdapterFactory functionality to MockRetrofit
I think what it needed to pass CallAdapterFactory to MockRetrofit or create ObservableBehaviorAdapter. If I use MockRetrofit, created with Retrofit with RxJavaCallAdapterFactory, then I get the ClassCastException:

```
java.lang.ClassCastException: rx.Observable cannot be cast to retrofit.Call
at retrofit.mock.CallBehaviorAdapter.applyBehavior(CallBehaviorAdapter.java:22)
at retrofit.mock.MockRetrofit$1.invoke(MockRetrofit.java:44)
at $Proxy1.setInviteAcceptionStatus(Native Method)
...
```

My code

```
(from dagger module)
    @Provides
    @Singleton
    Retrofit provideRetrofit() {
                .baseUrl("http://OLOLO")
                .addCallAdapterFactory(RxJavaCallAdapterFactory.create())
                .build();
        return retrofit;
    }

    @Provides
    @Singleton
    ServerApi provideServerApi(Retrofit retrofit) {
        NetworkBehavior behavior = NetworkBehavior.create();
        behavior.setFailurePercent(10);
        behavior.setVariancePercent(30);
        behavior.setDelay(3, TimeUnit.SECONDS);
        MockRetrofit mockRetrofit = new MockRetrofit(behavior, new CallBehaviorAdapter(retrofit, Executors.newSingleThreadExecutor()));
        return mockRetrofit.create(ServerApi.class, new MockServerApi());
    }

...
(my api interface)
public interface ServerApi {
    @GET("/test")
    Observable<Object>           test();
}
```

### LINKED ISSUE #1309: Mock Call not using callback executor
Dear square guys, 
sorry for posting this which is half a request, half an issue (depending on my errors or if it is an issue) but i have problems with async calls and threads.

I'm trying to run an async call (enqueue method) and in the "onResponse" method I save to Realm (the realm.io library) some data, and refresh a recyclerview. Last call is that I stop a "SwipeRefreshLayout" to show the loading circle.

I'm having weird behaviour since if I run realm directly in the onResponse method, I get an issue from realm telling me that I'm not on the UI Thread (but the documentation of callback says I should be!).
So it works if I put the realm code in runOnUIThread method. 
The problem arises if then I put the swiperefreshLayout out of this force to UI method. I get an exception form the framework telling me that I'm touching the view outside of the thread that generated it, but shouldn't be!
In addition, if i compare loopers it says false!

    public void onResponse(Response<Publication> response, Retrofit retrofit) {
                try {
                    //get the response object
                    dailies = response.body().getIssues();                  
                    Log.i(TAG, "Check loopers=" + (Looper.getMainLooper() == Looper.myLooper()));//Always false!
                    MainActivity.this.runOnUiThread(new Runnable() {
                        @Override
                        public void run() {
                            List<RealmIssue> realmIssues = new ArrayList<>();
                            RealmIssue realmIssue;
                            for (Issue issue : dailies) {
                                realmIssue = new RealmIssue(issue);
                                mRealmInstance.beginTransaction();
                                realmIssue = mRealmInstance.copyToRealmOrUpdate(realmIssue);
                                mRealmInstance.commitTransaction();
                                realmIssues.add(realmIssue);
                            }
                            Log.w("TEST", "Items copied to realm: " + realmIssues.size());
                           mRefreshLayout.setRefreshing(false);//Here works
                        }
                    });
                   //mRefreshLayout.setRefreshing(false);//Here does NOT work
            }

I'm not sure if it's an issue of the library or mine mistake, in case of the latter, please forgive me!

EDIT: I forgot to say, if it is important (and I think it is) that I'm running a mock retrofit for now: 

    MockRetrofit mockRetrofit = new MockRetrofit(NetworkBehavior.create(), new CallBehaviorAdapter(retrofit, Executors.newSingleThreadExecutor()));

JakeWharton:
This allows the mock implementation to only concern itself with the response type and the BehaviorDelegate will apply any delays from NetworkBehavior and also run through the Retrofit's CallAdapter.Factory list to turn the delayed Call into the appropriate type that matches the return type of the method.

Closes #1286. Closes #1309.

This needs more tests and more documentation, but those can come later.


retrofit-mock/src/main/java/retrofit2/mock/BehaviorDelegate.java
```diff
@@ -0,0 +1,64 @@
...
+  private final NetworkBehavior behavior;
+  private final Class<T> service;
+
+  BehaviorDelegate(Retrofit retrofit, NetworkBehavior behavior, Class<T> service) {
+    this.retrofit = retrofit;
+    this.behavior = behavior;
+    this.service = service;
+  }
+
+  public T returningResponse(Object response) {
```
swankjesse: Nice

PR #1354, opened 2015-12-08
Correct some Javadoc examples, always run Javadoc plugin.

### LINKED ISSUE #1351: No SNAPSHOT builds with new package name
The automatic SNAPSHOT deployment seems to be broken after #1342 . See https://travis-ci.org/square/retrofit/builds/95452176#L1793 for the Travis output/warnings.

JakeWharton:
Closes #1351.

### DIFF
--- a/retrofit/src/main/java/retrofit2/http/Field.java
+++ b/retrofit/src/main/java/retrofit2/http/Field.java
@@ -30,21 +30,22 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
  * field pair for each non-{@code null} item.
  * <p>
  * Simple Example:
- * <pre>
+ * <pre>{@code
  * &#64;FormUrlEncoded
  * &#64;POST("/")
- * void example(@Field("name") String name, @Field("occupation") String occupation);
- * }
- * </pre>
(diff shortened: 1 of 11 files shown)


retrofit/src/main/java/retrofit2/http/Field.java
```diff
@@ -30,21 +30,22 @@
  * field pair for each non-{@code null} item.
  * <p>
  * Simple Example:
- * <pre>
+ * <pre>{@code
```
swankjesse: not sure if you want to bother, but Javadoc might format better with three spaces before `{@code` and a newline after it.

yes, this advice is ridiculous

JakeWharton: Hah. Do you know why? I'm inclined to believe you since Javadoc is ridiculous, but I checked the IDEA-rendered and HTML-rendered Javadoc and what's in this PR rendered correctly.

swankjesse: Sounds good to me. That format was obtained by Kevin B at one point by trial & error; things might work better with the latest Javadoc.

PR #1361, opened 2015-12-10
Change sonatype directory to retrofit

vivekkiran:


JakeWharton:
This is a link to the root of the repository because this is the URL you need to add to Maven or Gradle when configuring snapshots.

e.g.,
```groovy
maven {
  url 'https://oss.sonatype.org/content/repositories/snapshots/'
}
```

PR #1370, opened 2015-12-10
[resolves #1192] Fixes the type passed to Jackson writer

### LINKED ISSUE #1192: Jackson converter breaks polymorphism
in https://github.com/square/retrofit/blob/master/retrofit/src/main/java/retrofit/RequestFactoryParser.java#L299: 
```java
converter = retrofit.requestConverter(methodParameterType, methodParameterAnnotations);
```
The method parameter type was directly passed into the creation of the `ObjectWriter` in Jackson converter.

In the case of
```java
public interface MyService {
    @PUT("/url")
    Call<ResponseBody> getMyClass(@Body BaseClass obj);
}
```
An object of `SubClass` will be written as `BaseClass` by the `ObjectWriter` with `_rootType=BaseClass`.

jianghaolu:
Addresses the issue: https://github.com/square/retrofit/issues/1192


JakeWharton:
Please see my comments on the issue. This is not the desired behavior.

PR #1376, opened 2015-12-12
Add a XStream converter

arkadiytehgraet:
Add XStream converter to handle xml-encoded requests and responses.


felipecsl:
See https://github.com/square/retrofit/pull/1200#issuecomment-148282153

JakeWharton:
What he said. Feel free to ship it as a standalone library when Retrofit 2 is final and add it to the wiki page though!

PR #1380, opened 2015-12-13
Add buildRequest() and proper implementation

### LINKED ISSUE #1367: Provide new method to get the original request from Retrofit Call proxies
Hi,

When using Retrofit Call proxies, it's not possible to be aware of the original request through those objects. When implementing a custom response cache (not depending on response cache of okhttp), generating a cache key with the request is necessary. However we cannot access to this original request, headers, params through Call objects (okhttpcall being final as well).

Would be nice to provide in the interface Call the method CreateRawCall already implemented in private.

dimitrovskif:
This PR fixes #1367, which is a basic issue:

* Once you instantiate a `Call<T>`, you may need to inspect the URL (which now contains set query params), as well as all body and header data that will be send in the future if you dispatch the call (by either `enqueue()` or `execute()`

### DIFF
--- a/retrofit/src/main/java/retrofit2/Call.java
+++ b/retrofit/src/main/java/retrofit2/Call.java
@@ -16,7 +16,6 @@
 package retrofit2;
 
 import com.squareup.okhttp.Request;
-
 import java.io.IOException;
 
 /**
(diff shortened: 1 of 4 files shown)


JakeWharton:
See #1484.

PR #1393, opened 2015-12-21
Switch to a proper builder for MockRetrofit.

JakeWharton:
This allows passing an explicit executor for test cases.

### DIFF
--- a/retrofit-mock/src/test/java/retrofit2/mock/BehaviorDelegateTest.java
+++ b/retrofit-mock/src/test/java/retrofit2/mock/BehaviorDelegateTest.java
@@ -1,0 +1,298 @@
+/*
+ * Copyright (C) 2013 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


retrofit-mock/src/test/java/retrofit2/mock/BehaviorDelegateTest.java
```diff
@@ -0,0 +1,298 @@
...
+import retrofit2.Response;
+import retrofit2.Retrofit;
+
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public final class BehaviorDelegateTest {
```
JakeWharton: This file is a rename of `MockRetrofitTest`, nothing changed exception the use of the `Builder` in `@Before`.

JakeWharton: (I wanted to separate the public API testing from the implementation testing as `RetrofitTest` vs `CallTest` in the main artifact does)

Issue #1149, opened 2015-09-30
Using Retrofit in Google App Engine


Udinic:
I used Retrofit 1.9 on a GAE app I have, in order to make network calls to other services. I used the default client (not UrlFetchClient) and everything worked fine.

After upgrading to Retrofit 2, I made all the necessary changes and now I get this error when trying to make a network call:

java.lang.NoClassDefFoundError: java.net.ProxySelector is a restricted class. Please see the Google  App Engine developer's guide for more details.
	at com.google.appengine.tools.development.agent.runtime.Runtime.reject(Runtime.java:52)
	at com.squareup.okhttp.OkHttpClient.copyWithDefaults(OkHttpClient.java:614)
	at com.squareup.okhttp.Call.<init>(Call.java:48)
	at com.squareup.okhttp.OkHttpClient.newCall(OkHttpClient.java:595)
	at retrofit.OkHttpCall.createRawCall(OkHttpCall.java:120)
	at retrofit.OkHttpCall.execute(OkHttpCall.java:110)

I assume that's because OkHttp is the default client (Similar problem in 1.9: http://stackoverflow.com/questions/30912029/using-okhttp-client-via-okclient-on-google-app-engine-throws-a-java-lang-noclas).

Since GAE was supported in older version, will it continue to be supported in Retrofit 2 ? Is there some configuration I need to do to get it to work?

JakeWharton:
There are no plans to support AppEngine in version 2. This isn't so much as an explicit thing, as it is implicit with our choice to couple the library to OkHttp.

Retrofit 1 existed in a time where there wasn't an HTTP client that met the needs of this library directly, and since Android support was of particular importance, we chose to make it pluggable with an abstraction. We paid a cost for this in the public API, the allocation overhead, and rigidity of execution mechanisms (clients were forced to be synchronous).

Over the years that Retrofit 1 existed, OkHttp grew up. We pushed what we learned from Retrofit and from clients and APIs all over into OkHttp and it became clean and highly robust. The extra abstraction of Retrofit 1 to support multiple clients seemed very old and outdated.

So with v2 we made hard choice to couple ourselves directly to OkHttp knowing that it would cause a small subset of people pain, but have large gains for the majority of its users. AppEngine seems like it's going to be one of those side-effect casualties.

Udinic:
Got it, thanks for the explanation.

On Wed, Sep 30, 2015, 9:49 PM Jake Wharton <notifications@github.com> wrote:

[...]

Issue #924, opened 2015-06-26
Allow access to OkHttp's WebSocket type


JakeWharton:
```java
interface Service {
  @GET("/chat/{id}")
  Call<WebSocket> chat(
      @Path("id") String id);
}
```

swankjesse:
Whoa.

JakeWharton:
There's problems here:

`WebSocket` is solely a writing type. Messages that are read come in as callbacks to the `WebSocketListener` which is also what has the open/failure callbacks. Additionally, we are re-using the failure callback for both setup failure and then protocol or transport failure down the line.

A potential solution would be to split the listener into two callback classes: one for the connection and one for messages. This could be done either here as a set of wrappers, or in OkHttp. You may remember that my initial API for WebSocket was done in this way before you commented that we could consolidate them into one.

However, the split is not entirely a solution because we have a threading problems. When we get the `WebSocket` instance in `onOpen` we are on the reader thread. Once that method returns, it sits on the loop reading messages from the wire. If the call adapter choses to move threads before handing the `WebSocket` to the caller, there's a timing issue where messages might be dropped because a listener has not been set.

One potential solution would be to do the callback split, change `onOpen` to `onConnected`, and add a method like `start(WebSocketListener)` to `WebSocket`. This would just be a `CountDownLatch` that holds the reader thread up for a period of time (like the connect timeout that we already have on the `OkHttpClient`) in case threads need to be moved.

I don't have other solutions in mind yet so I'm open to those and thoughts on what I proposed above.

swankjesse:
We'll want Retrofit's web socket to make `onMessage` take a `<T>`, and similarly for `sendMessage`.

I'm guessing Retrofit will encapsulate OkHttp's `WebSocket` and layer its own abstraction on top, to handle type adapters and threading. This is symmetric with `Call`, right?

JakeWharton:
Making it typed is very interesting. And that will of course require an abstraction. I'll experiment with it.

JakeWharton:
APIs to look at for inspiration/samples:

 * Slack: https://api.slack.com/rtm
 * IRC Cloud: https://github.com/irccloud/irccloud-tools/wiki/API-Stream-Message-Reference
 * Chrome DevTools: https://developer.chrome.com/devtools/docs/debugger-protocol
 * Trello: Not documented online beyond the WebSocket connect endpoint, but you can use Chrome DevTools to inspect.

JakeWharton:
[...]

PR #1394, opened 2015-12-21
Add Call.Factory for alternate HTTP client implementations.

### LINKED ISSUE #1377: Retrofit2 support for alternative HTTP clients?
I see that v2 couples very tightly to `OkHttp`. Specifically, `Retrofit.Builder.client(OkHttpClient)` only takes an implementation of `OkHttpClient` instead of an abstract interface.

I was looking to integrate v2 support in to Ratpack and I can make it work with the `CallAdapter` except that the implementation continues to use OkHttpClient behind the scenes. While functional, I was wondering if there is a way that I could replace the underlying HTTP client with Ratpack's Netty based Http client.

Is there a way to do this (even if hacky)? Is this a feature that you would be interested in supporting?

JakeWharton:
Closes #1377. Also solves #1149 indirectly!

For #924 we'll do a `WebSocketCall` and it will have its own `Factory`. Calling `client` on `Retrofit.Builder` will populate that instance automatically just like it does for `Call.Factory`.

### DIFF
--- a/retrofit/src/main/java/retrofit2/DeferredRequest.java
+++ b/retrofit/src/main/java/retrofit2/DeferredRequest.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package retrofit2;
+
+import com.squareup.okhttp.Request;
+import java.io.IOException;
+
+/**
+ * An un-built HTTP request. This class defers any work necessary to create an HTTP request until
+ * {@link #get()} is called.
+ */
+public interface DeferredRequest {
+  /** Perform the work necessary to create and then return the {@link Request}. */
+  Request get() throws IOException;
+}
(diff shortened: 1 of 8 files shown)


retrofit/src/main/java/retrofit2/DeferredRequest.java
```diff
@@ -0,0 +1,13 @@
+package retrofit2;
+
+import com.squareup.okhttp.Request;
+import java.io.IOException;
+
+/**
+ * An un-built HTTP request. This class defers any work necessary to create an HTTP request until
+ * {@link #get()} is called.
+ */
+public interface DeferredRequest {
```
JakeWharton: I originally used `Callable<Request>` for this, but the fact that it throws a full-on `Exception` was ugly and the naming clashes severely with out `Call` and `Callback` and friends. Really low-value class here. Java 8's `Supplier` doesn't fit either because of the need to throw `IOException` (not that we can use Java 8 types anyway).

retrofit/src/main/java/retrofit2/OkHttpCallFactory.java
```diff
@@ -0,0 +1,17 @@
+package retrofit2;
```
swankjesse: copyright

retrofit/src/main/java/retrofit2/Retrofit.java
```diff
@@ -357,9 +360,25 @@ public Builder() {
...
-    /** The HTTP client used for requests. */
+    /**
+     * The HTTP client used for requests.
+     * <p>
+     * This is a convenience method for calling {@link #callFactory} with an instance that uses
+     * {@code client} for execution.
+     * <p>
+     * Note: This method <b>does not</b> make a defensive copy of {@code client}. Changes to its
+     * settings will affect subsequent requests. Pass in a {@linkplain OkHttpClient#clone() cloned}
+     * instance to prevent this.
```
swankjesse: to prevent this if desired?

swankjesse:
LGTM

PR #1400, opened 2015-12-22
A random assortment of documentation additions.

JakeWharton:
I expect these to evolve even more... just had some time to start filling them out.

### DIFF
--- a/retrofit/src/main/java/retrofit2/Retrofit.java
+++ b/retrofit/src/main/java/retrofit2/Retrofit.java
@@ -33,58 +33,25 @@ import java.util.concurrent.Executor;
 import retrofit2.http.GET;
 import retrofit2.http.HTTP;
 import retrofit2.http.Header;
+import retrofit2.http.Url;
 
 import static retrofit2.Utils.checkNotNull;
 
 /**
- * Adapts a Java interface to a REST API.
+ * Retrofit adapts a Java interface to HTTP calls by using annotations on the declared methods to
+ * define how requests are made. Create instances using {@linkplain Builder
+ * the builder} and pass your interface to {@link #create} to generate an implementation.
  * <p>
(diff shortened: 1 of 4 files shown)


retrofit/src/main/java/retrofit2/Retrofit.java
```diff
@@ -110,7 +77,52 @@
...
+   * representation. {@link Void} can be used if you do not care about the body contents.
+   * <p>
+   * For example:
+   * <pre>
+   * public interface CategoryService {
+   *   &#64;POST("category/{cat}/")
+   *   Call&lt;List&lt;Item&gt;&gt; categoryList(@Path("cat") String a, @Query("page") int b);
+   * }
+   * </pre>
+   */
```
JakeWharton: I'd like to pare this down in the future...

retrofit/src/main/java/retrofit2/Retrofit.java
```diff
@@ -110,7 +77,52 @@
...
+   * </ul>
+   * <p>
+   * Additional static headers can be added for an endpoint using the
+   * {@link retrofit2.http.Headers @Headers} method annotation. For per-request control over a
+   * header annotate a parameter with {@link Header @Header}.
+   * <p>
+   * By default, methods return a {@link Call} which represents the HTTP request. The generic
+   * parameter of the call is the response body type and will be converted by one of the
+   * {@link Converter.Factory} instances. {@link ResponseBody} can also be used for a raw
+   * representation. {@link Void} can be used if you do not care about the body contents.
```
swankjesse: Is void synchronous?

JakeWharton: It's `Void` as in `Call<Void>` not `void` which leaves sync or async up to you.

PR #1423, opened 2016-01-01
Use Call.Factory in RequestBuilderTest.

swankjesse:


swankjesse:
Fix for build break: https://github.com/square/okhttp/pull/2183/files

PR #1426, opened 2016-01-02
Use new OkHttp's non-throwing close and a simpler body copy.

JakeWharton:


retrofit/src/main/java/retrofit2/OkHttpCall.java
```diff
@@ -140,10 +138,10 @@ private void callSuccess(Response<T> response) {
     if (code < 200 || code >= 300) {
       try {
         // Buffer the entire body to avoid future I/O.
-        ResponseBody bufferedBody = Utils.readBodyToBytesIfNecessary(rawBody);
+        ResponseBody bufferedBody = Utils.buffer(rawBody);
```
JakeWharton: Can't use `rawResponse.peekBody` here because we replace its body with a no-content one.

retrofit/src/main/java/retrofit2/Utils.java
```diff
@@ -59,21 +48,10 @@ static boolean isAnnotationPresent(Annotation[] annotations,
...
-
+  static ResponseBody buffer(final ResponseBody body) throws IOException {
     BufferedSource source = body.source();
-    Buffer buffer = new Buffer();
-    buffer.writeAll(source);
-    source.close();
-
-    return ResponseBody.create(body.contentType(), body.contentLength(), buffer);
+    source.request(Long.MAX_VALUE);
+    return ResponseBody.create(body.contentType(), body.contentLength(), source.buffer().clone());
```
JakeWharton: Actually this is worse, right? I want to move the segments, not copy them.

PR #1458, opened 2016-01-11
Supply method annotations to factory when invoking requestBodyConverter.

f2prateek:
I'm using Retrofit for JSON-RPC over HTTP using a [custom converter factory](https://github.com/segmentio/java-jsonrpc/blob/master/jsonrpc/src/main/java/com/segment/jsonrpc/JsonRPCConverterFactory.java).

Currently `Converter.Factory#requestBodyConverter` is only supplied the annotations on the method parameter. In the JSON-RPC case, this requires the use 2 separate annotations — one that is available to the response converter (to transparently unwrap the response from the server ) and one that is available to the request converter (to transparently encode the request for JSON-RPC and encode the name of the RPC method).

```java
interface MultiplicationService {
    @JsonRPC @POST("/rpc")
    Call<Integer> multiply(@Body @Method("Arith.Multiply") MultiplicationArgs args);
}
```

By supplying the method annotations to both the request and response converter factory methods, we can consolidate the annotations into a single one and simplify their service declaration.

```java
interface MultiplicationService {
    @JsonRPC("Arith.Multiply") @POST("/rpc")
    Call<Integer> multiply(@Body MultiplicationArgs args);
}
```

This implementation is NOT backwards compatible — all the existing converters will have to updated (first part converters have been updated).

### DIFF
--- a/retrofit/src/main/java/retrofit2/Converter.java
+++ b/retrofit/src/main/java/retrofit2/Converter.java
@@ -57,8 +57,8 @@ public interface Converter<F, T> {
      * specified by {@link Body @Body}, {@link Part @Part}, and {@link PartMap @PartMap}
      * values.
      */
-    public Converter<?, RequestBody> requestBodyConverter(Type type, Annotation[] annotations,
-        Retrofit retrofit) {
+    public Converter<?, RequestBody> requestBodyConverter(Type type,
+        Annotation[] parameterAnnotations, Annotation[] methodAnnotations, Retrofit retrofit) {
       return null;
     }
 
--- a/retrofit/src/main/java/retrofit2/Retrofit.java
+++ b/retrofit/src/main/java/retrofit2/Retrofit.java
@@ -241,8 +241,9 @@ public final class Retrofit {
    *
    * @throws IllegalArgumentException if no converter available for {@code type}.
    */
-  public <T> Converter<T, RequestBody> requestBodyConverter(Type type, Annotation[] annotations) {
-    return nextRequestBodyConverter(null, type, annotations);
+  public <T> Converter<T, RequestBody> requestBodyConverter(Type type,
+      Annotation[] parameterAnnotations, Annotation[] methodAnnotations) {
+    return nextRequestBodyConverter(null, type, parameterAnnotations, methodAnnotations);
   }
 
   /**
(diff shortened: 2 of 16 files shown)


retrofit/src/main/java/retrofit2/Retrofit.java
```diff
@@ -282,6 +285,27 @@ public BaseUrl baseUrl() {
...
+
+  /**
+   * Returns a {@link Converter} for {@code type} to {@link RequestBody} from the available
+   * {@linkplain #converterFactories() factories} except {@code skipPast}.
+   *
+   * @throws IllegalArgumentException if no converter available for {@code type}.
+   */
+  public <T> Converter<T, RequestBody> nextRequestBodyConverter(Converter.Factory skipPast,
+      Type type, Annotation[] annotations) {
+    return nextRequestBodyConverter(skipPast, type, annotations, new Annotation[0]);
```
f2prateek: We'd probably wanna extract this into a constant.

`static final Annotation[] EMPTY_ANNOTATION_ARRAY = new Annotation[0];`

retrofit/src/main/java/retrofit2/Converter.java
```diff
@@ -63,6 +63,17 @@
     }
 
     /**
+     * Returns a {@link Converter} for converting {@code type} to an HTTP request body, or null if
+     * {@code type} cannot be handled by this factory. This is used to create converters for types
+     * specified by {@link Body @Body}, {@link Part @Part}, and {@link PartMap @PartMap}
+     * values.
+     */
+    public Converter<?, RequestBody> requestBodyConverter(Type type,
```
JakeWharton: While this was made an abstract class so we could add methods to it, I'd prefer to remove the old callback altogether. I don't want to live with both for the next 2-3 years and we're still in beta (the point of which is finding problems with the API like the one you encountered).

retrofit/src/main/java/retrofit2/Retrofit.java
JakeWharton: Ship it :shipit:

retrofit/src/main/java/retrofit2/Converter.java
f2prateek: :+1:, updated the PR!

JakeWharton:
Would love a JSON-RPC sample in `samples/` if you ever find the time.

PR #1467, opened 2016-01-13
Add support for MultipartBody.Part in @Part

JayNewstrom:
This is very much a work in progress. But it works for my needs. I'd like to push this functionality forward, but could use some direction. I'm happy to update the PR given any direction, or let someone else take over.

References #1140

### DIFF
--- a/retrofit/src/main/java/retrofit2/RequestAction.java
+++ b/retrofit/src/main/java/retrofit2/RequestAction.java
@@ -21,6 +21,7 @@ import java.lang.reflect.Array;
 import java.net.URI;
 import java.util.Map;
 import okhttp3.Headers;
+import okhttp3.MultipartBody;
 import okhttp3.RequestBody;
 
 import static retrofit2.Utils.checkNotNull;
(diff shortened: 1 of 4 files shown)


JayNewstrom:
There are a couple of interesting things I ran into with this change
- The value of `@Path` will be ignored if you pass in `MultipartBody.Part`
- There is no request body `ConverterFactory` for `MultipartBody.Part` so it blows up without the changes in `RequestAction`
- I *think* the changes are needed in `RequestAction.Part` for the iterable and array versions, however, I didn't test that.
- I also introduced an unchecked cast on `RequestFactoryParser` line 437

retrofit/src/main/java/retrofit2/RequestAction.java
```diff
@@ -202,13 +203,17 @@
     @Override void perform(RequestBuilder builder, T value) {
       if (value == null) return; // Skip null values.
 
-      RequestBody body;
-      try {
-        body = converter.convert(value);
-      } catch (IOException e) {
-        throw new RuntimeException("Unable to convert " + value + " to RequestBody", e);
+      if (MultipartBody.Part.class.isAssignableFrom(value.getClass())) {
```
JakeWharton: This is known in the `RequestFactoryParser` and should have an action chosen for the type, not as just-in-time behavior.

JakeWharton:
There are many edge cases around this which are unknown and need evaluated before decided how to support this.

JayNewstrom:
I've updated the pull request. I'm not sure there's a clean way to do it in `RequestFactoryParser` because of the iterable/array implementations. I've added another test case to cover that functionality as well.

PR #1478, opened 2016-01-14
Add support for scalars from ResponseBody

### LINKED ISSUE #1459: Add converter for scalars to response body

JayNewstrom:
Closes #1459

### DIFF
--- a/retrofit-converters/scalars/src/main/java/retrofit2/ScalarResponseBodyConverters.java
+++ b/retrofit-converters/scalars/src/main/java/retrofit2/ScalarResponseBodyConverters.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright (C) 2016 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package retrofit2;
+
+import java.io.IOException;
+import okhttp3.ResponseBody;
+
+final class ScalarResponseBodyConverters {
+  private ScalarResponseBodyConverters() {
+  }
+
+  static final class StringResponseBodyConverter implements Converter<ResponseBody, String> {
+    static final StringResponseBodyConverter INSTANCE = new StringResponseBodyConverter();
+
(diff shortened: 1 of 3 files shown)


connorMathews:
Hey Jay, can you make the converters Singletons since they're stateless SAMs?

```
static final class ByteResponseBodyConverter implements Converter<ResponseBody, Byte> {
   static final ByteResponseBodyConverter INSTANCE = ByteResponseBodyConverter();
   
    @Override public Byte convert(ResponseBody value) throws IOException {
        return Byte.valueOf(value.string());
    }
}

...

if (type == Boolean.class) {
    return BooleanResponseBodyConverter.INSTANCE;
}
```

instead of 

```
static final class ByteResponseBodyConverter implements Converter<ResponseBody, Byte> {
    @Override public Byte convert(ResponseBody value) throws IOException {
        return Byte.valueOf(value.string());
    }
}

...

if (type == Boolean.class) {
    return new BooleanResponseBodyConverter();
}
```

It'll reduce object allocations.

Thanks!

JayNewstrom:
That's how `ScalarRequestBodyConverter.INSTANCE` is done as well. Updated.

retrofit-converters/scalars/src/main/java/retrofit2/ScalarResponseBodyConverters.java
```diff
@@ -0,0 +1,81 @@
+package retrofit2;
```
JakeWharton: Copy license header from another file. Use 2016 for the date.

JakeWharton:
Looks great! Just need a license header and we're good to go.

JayNewstrom:
Whoops! Updated.

retrofit-converters/scalars/src/main/java/retrofit2/ScalarResponseBodyConverters.java
```diff
@@ -0,0 +1,96 @@
...
+    @Override public Byte convert(ResponseBody value) throws IOException {
+      return Byte.valueOf(value.string());
+    }
+  }
+
+  static final class CharacterResponseBodyConverter implements Converter<ResponseBody, Character> {
+    static final CharacterResponseBodyConverter INSTANCE = new CharacterResponseBodyConverter();
+
+    @Override public Character convert(ResponseBody value) throws IOException {
+      return value.string().charAt(0);
```
JakeWharton: I'm thinking that this should assert that the string is exactly 1 character and throw an `IOException` for a non-1 length. And a test case for both the 0 and non-1 lengths.

JakeWharton: Actually I'll just make this change locally during the merge.

JayNewstrom: :+1:

JakeWharton:
Merged manually. Thanks!

PR #1487, opened 2016-01-18
Do not trap IO from request body conversion.

JakeWharton:


retrofit-converters/gson/src/main/java/retrofit2/GsonRequestBodyConverter.java
```diff
@@ -42,12 +42,8 @@
...
     Writer writer = new OutputStreamWriter(buffer.outputStream(), UTF_8);
     JsonWriter jsonWriter = gson.newJsonWriter(writer);
-    try {
-      adapter.write(jsonWriter, value);
-      jsonWriter.flush();
-    } catch (IOException e) {
-      throw new AssertionError(e); // Writing to Buffer does no I/O.
-    }
+    adapter.write(jsonWriter, value);
+    jsonWriter.close();
```
JakeWharton: This is also a slight behavior change as it now validates the JSON is complete.

swankjesse:
Win.

PR #1491, opened 2016-01-19
Update to the latest version of wire

JayNewstrom:


retrofit-converters/wire/src/test/java/retrofit2/Phone.java
```diff
@@ -1,5 +1,5 @@
 // Code generated by Wire protocol buffer compiler, do not edit.
-// Source file: test.proto at 2:1
+// Source file: phone.proto at 6:1
```
JayNewstrom: Weird that this is different.

JakeWharton:
Thanks.

Issue #1500, opened 2016-01-21
Question about custom annotations


b22n:
I want to do something like this.

```
public interface RestService {
   @Secure
   @GET("somthing")
   Call<..> getSomthing();
}

@Target(METHOD)
@Retention(RUNTIME)
public @interface Secure {}
```

There is no way for referencing my custom annotations from intercepted Request object now maybe
Could it be supported?

danvinokour:
There might be solution according to Jack.
https://github.com/square/retrofit/issues/1496

b22n:
Yes that may be a tricky idea but not the solution that I exactly want to

andretietz:
would my pull request do the job?

andretietz:
Not the nicest solution, but it fits my needs
https://medium.com/@andretietz/custom-annotations-with-retrofit-2-8701ca7ce102#.6y800r5rm

swankjesse:
No action to take on this.

Issue #1422, opened 2015-12-30
Passing one interceptor only per request using okhttp Call proxies


vpasquier:
Hi,

I've seen here #261 the work and discussions about how to add header per request and not globally for all requests. But the solution of adding @header parameters is not convenient for the purpose of my use case (I have lots of endpoints method to implement and headers as well).

Is there a way to pass to the okhttp Call proxies an interceptor (for one request of this proxy only) dynamically?

Thank you!

JakeWharton:
You want to pass an interceptor as a parameter?

vpasquier:
Nop that is what I want to avoid.

I would like to add interceptor like we do for networkInterceptor or interceptor but not as global for all requests, just only per request. And the thing is that apparently there is no mean to pass this "local" sporadic interceptor with the Call proxy.

Except using the object com.squareup.okhttp.Request. But I would like to use Retrofit.

JakeWharton:
How do you associate the interceptor with the desired endpoint for which it is to apply?

vpasquier:
Oh yeah I understand your point. It's true that on my proxy I'm going to have multiple endpoint and it could be just applied on this group of endpoints. (why not..)

So in another way, in order to control all the global interceptors to set them just once per request using Retrofit proxies, what would be the best solution to do this? I wanted to manage my interceptors, but there is no id for them as they are in a list (com.squareup.okhttp.OkHttpClient#interceptors or network ones).

What could be the best? Setting global interceptors is interesting but sometimes you would like to set just per request a header.

JakeWharton:
I still don't understand the API you are looking for. You don't want a global interceptor but you also don't want a per-call interceptor. Can you phrase the request in the form of code?

vpasquier:
Ok sorry.

I want a per-call interceptor. How can I achieve that ? Is the only solution is to use the @header parameters?

JakeWharton:
You can use a `@Header` param for a dynamic, per-endpoint value, the `@Headers` method annotation for a fixed, per-endpoint value, or use an global OkHttp interceptor and match on the request path (or anything else) for modifying the request in any arbitrary way. We don't have any other ways to customize a single endpoint.

vpasquier:
First of all thank you for the client and this release beta3. Huge work!

Unfortunately I still stuck with Retrofit proxies to access to the original request, update the headers per call and to retrieve easily the original response.

To explain more, because it's globally the same subject: **how to get/update original request and response with proxies.**

### Request

[...]

Issue #626, opened 2014-10-07
Expose RequestAction for pluggable parameters handlers.


JakeWharton:
There's a lot of types people want to use for a lot of things. Having a mechanism to support custom parameter type handling would provide an extension point (albeit an advanced one) to prevent the further need to one-off these in the future. Bonus points if we can somehow figure out how to build the first-party behavior using this mechanism.
#613.

Carthan:
+1

matiwinnetou:
+1

podollb:
+1

jrodbx:
+1

tkdrahn:
+1

JakeWharton:
This mechanism exists (`RequestAction`) but its exposure is being punted to 2.1. That said, type handlers (aka converters) for non-body parameters which want a string (like `@Path`, `@Header`, etc.) will be supported to allow the use of alternate types (like `Date`, et. al.).

matiwinnetou:
Now that 2.1 is out is there a way to do this?

JakeWharton:
Not yet

iwo:
Have a look at my draft implementation of this feature (https://github.com/iwo/retrofit/commit/e843d9f94b114ea809d43bf7c6eaf38e4764275c) and let me know what you think.
To simplify review this commit only rearranges existing code without major flow changes or optimization. I'll extract and reuse common `ParameterHandler` code in further commits.

**Design implications**
To allow users to create custom `ParameterHandler` classes `RequestBuilder` API needs to be public. All the published methods should be reviewed and documented. The linked issues should be reviewed to ensure that the published `RequestBuilder` API provides enough low level access to allow for implementation of all the peculiar use cases mentioned.

**Code style consideration**
I moved built in `ParameterHandler` implementation to a separate package. They could be nested under  `ParameterHandler` abstract class but it would make this file really long (probably around 1k LOC). The trade-off is that package local methods can no longer be used and `Utlis` is made public. On the other hand, there is a chance that users implementing custom handlers will benefit from access to `Utils` methods so maybe it's a good thing.

Please let me know what you think about this approach.

iwo:
More changes in [pluggable-parameters-handlers](https://github.com/iwo/retrofit/commits/pluggable-parameters-handlers) branch.

The code duplication got significantly reduced so it may make sense to merge some of this changes even if you decide not to make the pluggable parameter handlers API public at this point.

matiwinnetou:
[...]

PR #1507, opened 2016-01-21
Adds RequestBuildInterceptor

andretietz:
The user can add a RequestBuildInterceptor.Factory to retrofit, which hooks into a point directly after the request has been generated via the retrofit2.RequestBuilder. On this way the request can be modified depending on whatever the user wants to add (using request.newBuilder)
This is a possible solution for #1422, #1500

PS: The naming could be improved, since "RequestBuildInterceptor.Factory" is quite a long name.

### DIFF
--- a/retrofit/src/main/java/retrofit2/RequestBuildInterceptor.java
+++ b/retrofit/src/main/java/retrofit2/RequestBuildInterceptor.java
@@ -1,0 +1,25 @@
+package retrofit2;
+
+import java.lang.reflect.Method;
+
+import okhttp3.Request;
+
+/**
+ * With this interceptor, you can intercept the request build process and modify
+ * the request as you want it to be executed. i.e. adding headers based on annotations
+ */
+public interface RequestBuildInterceptor {
+
+  /**
+   * @param request the request object as it has been generated based on the method
+   * @return a (modified) request, as you want it to be called
+   */
+  Request build(Request request);
+
+  /**
+   * Creates {@link RequestBuildInterceptor} based on the method which has been called
+   */
+  interface Factory {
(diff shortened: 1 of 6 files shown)


retrofit/src/main/java/retrofit2/RequestBuildInterceptor.java
```diff
@@ -0,0 +1,25 @@
...
+   * @param request the request object as it has been generated based on the method
+   * @return a (modified) request, as you want it to be called
+   */
+  Request build(Request request);
+
+  /**
+   * Creates {@link RequestBuildInterceptor} based on the method which has been called
+   */
+  interface Factory {
+    RequestBuildInterceptor get(Method method);
```
andretietz: instead of passing the whole method object I would be ok with passing the annotations only, but who knows what the user actually wants to read out of the method. So I decided to pass the method, to be more open for other options as well (i.e. return type dependent requests or similar things)

swankjesse:
I think a simpler solution i to just use an OkHttp interceptor, which can also rewrite the request.

andretietz:
The point is,  to create a request based on an annotation (or something else) on the request method. 
How would you do that with an okhttp interceptor?

swankjesse:
You can tell the interceptor what you want with a header. An example here:
https://publicobject.com/2016/01/17/sneaking-data-into-an-okhttp-interceptor/

andretietz:
I was talking about Custom annotations. Not the retrofit inbuild ones. 
It's a dirty solution to do it like you're suggesting IMHO

JakeWharton:
We'll be evaluating custom annotations and behaviors as part of #626.

PR #1508, opened 2016-01-21
Exception message composition correction

setheclark:
Correcting the collection used when adding the skipped converters to the exception message when no converter was found.


JakeWharton:
Ouch. This needs a test! There's ones for the other adapter methods failing. I'll add one later if you don't get to it.

PR #1515, opened 2016-01-24
baseUrl should end with '/'

manidesto:


JakeWharton:
Can you make this change against `master` instead? The website is versioned with the code in the `website/` directory. Thanks.

PR #1516, opened 2016-01-24
baseUrl should end with '/'

manidesto:


JakeWharton:
Thanks!

manidesto:
@JakeWharton I didn't know that you can have a folder pushed to separate branch in git. I guess I should thank you 👍

PR #1520, opened 2016-01-25
Refactor and fill out RxJava adapter tests.

JakeWharton:


swankjesse:
I like how uniform these are.

PR #1528, opened 2016-01-25
Ensure default methods can be invoked.

### LINKED ISSUE #1514: java.lang.IllegalAccessException with interface default method
Hello!

I have this:

    public interface LivestreamsApi {

        @GET("api/livestreams")
        Call<LivestreamsContainer> getLivestreams(@Query("limit") int limit);

        default Call<LivestreamsContainer> getLivestreams() {
            return getLivestreams(100);
        }
    }

Executing the method with the Query param works. But running the default method doesn't work. It says: http://pastebin.com/V8xGSF5K
When I add for (whatever reason?) `throws Exception` to the default method it now only says: http://pastebin.com/pZaEd0v0
But how to solve this exception and why do I have to add the `throws`?

Thanks in advance!
RandomByte

JakeWharton:
Because the service interface might not be public, we need to use a MethodHandle lookup means that ignores visibility.

Closes #1514.

### DIFF
--- a/retrofit/src/main/java/retrofit2/Platform.java
+++ b/retrofit/src/main/java/retrofit2/Platform.java
@@ -18,7 +18,8 @@ package retrofit2;
 import android.os.Build;
 import android.os.Handler;
 import android.os.Looper;
-import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodHandles.Lookup;
+import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 import java.util.concurrent.Executor;
 import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement;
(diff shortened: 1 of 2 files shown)


retrofit/src/main/java/retrofit2/Platform.java
```diff
@@ -70,8 +71,11 @@ Object invokeDefaultMethod(Method method, Class<?> declaringClass, Object object
 
     @Override Object invokeDefaultMethod(Method method, Class<?> declaringClass, Object object,
         Object... args) throws Throwable {
-      return MethodHandles.lookup()
-          .in(declaringClass)
+      // Because the service interface might not be public, we need to use a MethodHandle lookup
```
swankjesse: Why wouldn't it be public? Strange

retrofit/src/test/java/retrofit2/Java8DefaultMethodsTest.java
```diff
@@ -0,0 +1,27 @@
...
+  //  server.enqueue(new MockResponse().setBody("Hi"));
+  //
+  //  Retrofit retrofit = new Retrofit.Builder()
+  //      .baseUrl(server.url("/"))
+  //      .addConverterFactory(new ToStringConverterFactory())
+  //      .build();
+  //  Example example = retrofit.create(Example.class);
+  //
+  //  Response<String> response = example.user().execute();
+  //  assertThat(response.body()).isEqualTo("Hi");
```
swankjesse: also call user() ?

swankjesse:
LGTM

retrofit/src/main/java/retrofit2/Platform.java
JakeWharton: Don't know. Our tests and samples all use package-scoped interfaces though.

PR #1540, opened 2016-01-29
Add example of using JSON converter as a String converter.

JakeWharton:
This also adds the Retrofit instance as a param to the string converter to allow it to delegate to others directly. This brings it into parity with the method signatures of the other converter methods anyway.

### DIFF
--- a/samples/src/main/java/com/example/retrofit/JsonQueryParameters.java
+++ b/samples/src/main/java/com/example/retrofit/JsonQueryParameters.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright (C) 2016 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


samples/src/main/java/com/example/retrofit/JsonQueryParameters.java
```diff
@@ -0,0 +1,113 @@
...
+
+    static class DelegateToStringConverter<T> implements Converter<T, String> {
+      private final Converter<T, RequestBody> delegate;
+
+      DelegateToStringConverter(Converter<T, RequestBody> delegate) {
+        this.delegate = delegate;
+      }
+
+      @Override public String convert(T value) throws IOException {
+        Buffer buffer = new Buffer();
```
swankjesse: neat

swankjesse:
LGTM

samples/src/main/java/com/example/retrofit/JsonQueryParameters.java
JakeWharton: https://github.com/square/okhttp/issues/1891 :trollface:

Issue #1364, opened 2015-12-10
Multi segment path replacement 


iwo:
Looking at [StackOverflow](http://stackoverflow.com/questions/23833290/retrofit-path-replacements-replacement-over-whole-path-including) questions and other issues here (#1156), there is clearly a need for multi segment path replacement. The workaround used to be marking `@Path` as encoded and passing `/` inside a path argument but this is no longer supported. Using  `@Url` requires building whole url in the calling code which is a dirty workaround as well.

I see two elegant solutions to this problem:
* Add `multiSegment` param to the `@Path` annotation which will turn off the encoding of `/`
```
@GET(„/tree/element/{path}”)
Call<TreeElement> getTreeElement(@Path(„path”, multiSegment = true) String path)
```
* Accept `String` array argument for path segments and join them with `/` when constructing the final path
```
@GET(„/tree/element/{path}”)
Call<TreeElement> getTreeElement(@Path(„path”) String[] pathElements)
```

Personally I prefer the `String[] pathElements` variant.
Let me know what you think about this approach and I may submit a pull request.

JakeWharton:
While the latter seems more correct in terms of modeling, my guess as to when people will encounter this case is in receiving a hypermedia-like string and not want to parse it themselves which lends to the former (although, one could simply call `.split("/")` for an array). Definitely needs thought. Presumably you could also pass a `List<String>` for the latter case.

> there is clearly a need

There is a _desire_ for a lot of cases that we simply cannot support in Retrofit while maintaining a small, sane, and focused API. So it's less about the fact that someone needs to use this and more about how high-leverage a feature like this would be in the API. Does it hold its weight in term so of usage and complexity to implement?

I'm not convinced that it is as this time, so this won't be in 2.0. I'll leave the issue open for post-2.0 consideration.

jakeswenson:
I think to add to the discussion on how to make multisegment parameters easier in retrofit, i'm not a fan of the suggested solutions. 

What if instead you combine the suggestions so that it would look something like `@Path(multiSegment = true) Iterable<T> pathSegments`

This way you explicitly designate that each value you should be encoded an then joined together with `/`

Either that or using an interface to have a `pathSegments()` method so you could have a custom type, though i feel the former is more inline with the rest of retrofit.

iwo:
[...]

PR #1547, opened 2016-01-30
Bugfix/issue 1364 partial fix respect encoding for path characters

jakeswenson:
This is a partial fix for #1364 to fix the `RequestBuilder` to respect the encoded value for path parameters.

I wanted to make sure that the logic for skipping potentially dangerous chars like `\n` remains intact.

Added three tests to verify this change.

This change allows for temporary workaround of #1364, allowing the API owner to set certain path parameters as pre-encoded.

### DIFF
--- a/retrofit/src/main/java/retrofit2/RequestBuilder.java
+++ b/retrofit/src/main/java/retrofit2/RequestBuilder.java
@@ -85,10 +85,10 @@ final class RequestBuilder {
       // The relative URL is cleared when the first query parameter is set.
       throw new AssertionError();
     }
-    relativeUrl = relativeUrl.replace("{" + name + "}", canonicalize(value, encoded));
+    relativeUrl = relativeUrl.replace("{" + name + "}", canonicalizeForPath(value, encoded));
   }
 
-  private static String canonicalize(String input, boolean alreadyEncoded) {
+  private static String canonicalizeForPath(String input, boolean alreadyEncoded) {
     int codePoint;
     for (int i = 0, limit = input.length(); i < limit; i += Character.charCount(codePoint)) {
       codePoint = input.codePointAt(i);
@@ -97,7 +97,7 @@ final class RequestBuilder {
         // Slow path: the character at i requires encoding!
         Buffer out = new Buffer();
         out.writeUtf8(input, 0, i);
-        canonicalize(out, input, i, limit, alreadyEncoded);
+        canonicalizeForPath(out, input, i, limit, alreadyEncoded);
         return out.readUtf8();
       }
     }
@@ -106,8 +106,8 @@ final class RequestBuilder {
     return input;
   }
 
-  private static void canonicalize(Buffer out, String input, int pos, int limit,
-      boolean alreadyEncoded) {
+  private static void canonicalizeForPath(Buffer out, String input, int pos, int limit,
+                                          boolean alreadyEncoded) {
     Buffer utf8Buffer = null; // Lazily allocated.
     int codePoint;
     for (int i = pos; i < limit; i += Character.charCount(codePoint)) {
--- a/retrofit/src/main/java/retrofit2/RequestBuilder.java
+++ b/retrofit/src/main/java/retrofit2/RequestBuilder.java
@@ -29,7 +29,7 @@ import okio.BufferedSink;
 final class RequestBuilder {
   private static final char[] HEX_DIGITS =
       { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F' };
-  private static final String PATH_SEGMENT_ENCODE_SET = " \"<>^`{}|/\\?#";
+  private static final String PATH_SEGMENT_ALWAYS_ENCODE_SET = " \"<>^`{}|\\?#";
 
   private final String method;
 
@@ -94,8 +94,8 @@ final class RequestBuilder {
       codePoint = input.codePointAt(i);
       if (codePoint < 0x20
           || codePoint >= 0x7f
-          || (!alreadyEncoded
-              && (PATH_SEGMENT_ENCODE_SET.indexOf(codePoint) != -1 || codePoint == '%'))) {
+          || PATH_SEGMENT_ALWAYS_ENCODE_SET.indexOf(codePoint) != -1
+          || (!alreadyEncoded && (codePoint == '/' || codePoint == '%'))) {
         // Slow path: the character at i requires encoding!
         Buffer out = new Buffer();
         out.writeUtf8(input, 0, i);
@@ -118,9 +118,9 @@ final class RequestBuilder {
           && (codePoint == '\t' || codePoint == '\n' || codePoint == '\f' || codePoint == '\r')) {
         // Skip this character.
       } else if (codePoint < 0x20
-          || codePoint >= 0x7f
-          || (!alreadyEncoded
-              && (PATH_SEGMENT_ENCODE_SET.indexOf(codePoint) != -1 || codePoint == '%'))) {
+                 || codePoint >= 0x7f
+                 || PATH_SEGMENT_ALWAYS_ENCODE_SET.indexOf(codePoint) != -1
+                 || (!alreadyEncoded && (codePoint == '/' || codePoint == '%'))) {
         // Percent encode this character.
         if (utf8Buffer == null) {
           utf8Buffer = new Buffer();
--- a/retrofit/src/main/java/retrofit2/RequestBuilder.java
+++ b/retrofit/src/main/java/retrofit2/RequestBuilder.java
@@ -95,8 +95,7 @@ final class RequestBuilder {
       if (codePoint < 0x20
           || codePoint >= 0x7f
           || (!alreadyEncoded
-              && PATH_SEGMENT_ENCODE_SET.indexOf(codePoint) != -1)
-          || (codePoint == '%' && !alreadyEncoded)) {
+              && (PATH_SEGMENT_ENCODE_SET.indexOf(codePoint) != -1 || codePoint == '%'))) {
         // Slow path: the character at i requires encoding!
         Buffer out = new Buffer();
         out.writeUtf8(input, 0, i);
@@ -121,8 +120,7 @@ final class RequestBuilder {
       } else if (codePoint < 0x20
           || codePoint >= 0x7f
           || (!alreadyEncoded
-              && PATH_SEGMENT_ENCODE_SET.indexOf(codePoint) != -1)
-              || (codePoint == '%' && !alreadyEncoded)) {
+              && (PATH_SEGMENT_ENCODE_SET.indexOf(codePoint) != -1 || codePoint == '%'))) {
         // Percent encode this character.
         if (utf8Buffer == null) {
           utf8Buffer = new Buffer();
--- a/retrofit/src/main/java/retrofit2/RequestBuilder.java
+++ b/retrofit/src/main/java/retrofit2/RequestBuilder.java
@@ -92,9 +92,8 @@ final class RequestBuilder {
     int codePoint;
     for (int i = 0, limit = input.length(); i < limit; i += Character.charCount(codePoint)) {
       codePoint = input.codePointAt(i);
-      if (codePoint < 0x20 || codePoint >= 0x7f
-          || PATH_SEGMENT_ENCODE_SET.indexOf(codePoint) != -1
-          || (codePoint == '%' && !alreadyEncoded)) {
+      if (shouldSkipCodePoint(codePoint)
+          || (!alreadyEncoded && shouldEncodeCodePoint(codePoint))) {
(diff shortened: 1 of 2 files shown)


jakeswenson:
I'd think the best milestone target for this fix is 2.0, because this could be considered a breaking change. best to do it as a part of 2.0.

retrofit/src/main/java/retrofit2/RequestBuilder.java
```diff
@@ -138,6 +134,17 @@ private static void canonicalize(Buffer out, String input, int pos, int limit,
     }
   }
 
+  private static boolean shouldSkipCodePoint(final int codePoint) {
```
devshorts: Is it possible to name this something like `isUnsafeCodePoint` since these are skipped to prevent xss right?  Just a suggestion

jakeswenson: like it, @JakeWharton thoughts?

retrofit/src/main/java/retrofit2/RequestBuilder.java
```diff
@@ -107,19 +106,16 @@ private static String canonicalize(String input, boolean alreadyEncoded) {
...
       codePoint = input.codePointAt(i);
       if (alreadyEncoded
-          && (codePoint == '\t' || codePoint == '\n' || codePoint == '\f' || codePoint == '\r')) {
+          && shouldSkipCodePoint(codePoint)) {
         // Skip this character.
-      } else if (codePoint < 0x20
-          || codePoint >= 0x7f
-          || PATH_SEGMENT_ENCODE_SET.indexOf(codePoint) != -1
-          || (codePoint == '%' && !alreadyEncoded)) {
+      } else if (!alreadyEncoded && shouldEncodeCodePoint(codePoint)) {
```
swankjesse: This is incorrect when alreadyEncoded is true and the characters is out of range, like `\u0000`.

swankjesse:
I don’t like this change. The PR mixes refactoring with behavior change and ends up accidentally breaking some important cases.

I also don’t like the behavior change to the API. It means that `/` has a different meaning if the path parameter is encoded vs. not-encoded.

retrofit/src/main/java/retrofit2/RequestBuilder.java
swankjesse: This is unrelated to XSS. The actual purpose is quite hilarious: it’s due to the fact that URLs are occasionally transcribed on paper or similar, and as a consequence spaces are not significant.

retrofit/src/main/java/retrofit2/RequestBuilder.java
jakeswenson: Thanks, but not sure if i understand. If the API creator says its encoded, why should if be re-encoded? If the caller passed in a bad value with the understanding it was to be encoded, why would they have any expectation it would work?

jakeswenson:
Thanks for the discussion @swankjesse. First of all the intent was to make the behavior change, if having the refactoring in the same PR is a problem i can undo it and make only the behavior change, albeit begrudgingly. :) Its you guys as project owners call.

I'm not sure i understand your second concern, are you suggesting that `/` should always be encoded, even if a declare my parameter as having been URL encoded? Would you also want '%' to be re-encoded as well?

I'm trying to encourage consistency, as I was certainly surprised to find out that even though i declare a parameter as url encoded, parts are re-encoded in an inconsistent way (`/` vs `%`). The percent based encoding RFC specifies 18 reserved characters, and i feel that `encoded = true` should mean if they're in the value, they're intentionally in the value.

I'm using retrofit at work, and we love it. But this is a problem currently with no work around. What would you suggest we do?

retrofit/src/main/java/retrofit2/RequestBuilder.java
jakeswenson: Interesting. I guess i'm surprised that _runtime_ replacement of whitespace characters helps paper transcription. My thoughts was that it was to prevent request splitting type issues. (not xss)
Do you have a better name in mind? perhaps `shouldSkipHardToTranscribeCodePoints`? i can add a comment capturing your thoughts above.

jakeswenson:
Also if there are important cases that this change is breaking, are they codified into unit tests? 
If so, all unit tests are passing, so can you point me to the breaking change?
If not, should I file an issue to make sure those scenarios are unit tested?

retrofit/src/main/java/retrofit2/RequestBuilder.java
JakeWharton: It is not re-encoded, but canonicalized as a value inside a single path
segment. With the current behavior, a slash is an impossible character to
exist inside a path segment without being transformed to an escaped version.

The path is currently treated as an ordered list of segments in both
Retrofit and OkHttp's HttpUrl class, which is what is leading to your
misinterpretation of the current behavior. It's not wrong, just explicit in
a way you don't agree with which results in surprise.

retrofit/src/main/java/retrofit2/RequestBuilder.java
JakeWharton: I would rather these methods didn't exist and the original structure was kept.

jakeswenson: No refactoring, so noted. Will reverse these changes.

retrofit/src/main/java/retrofit2/RequestBuilder.java
jakeswenson: I agree. I think there is a disagreement between semantics, and the documentation/code not aligning with my (mis)interpretation. Are you able to see where my confusion is coming from -- why i consider re-encoding to have `alreadyEncoded = true` but also canonicalize/encode `/` to `%2F`?

If you don't want to current behavior of the encoded parameter to change, then thats fine. If you don't see why someone may misunderstand the encoded parameter thats a different thing. As a way to prevent further misunderstandings can we at least agree to update the documentation of the encoded parameter?

```
/**
 * Specifies whether the argument value to the annotated method parameter is already URL encoded.
 */
```

Thanks for the explanation of Retrofit/OkHttp using an ordered list or path segments. I suppose it would have been clearer to me if i hadn't seen the url being build with string replacements...

```
relativeUrl = relativeUrl.replace("{" + name + "}", canonicalizeForPath(value, encoded));
```
I will spend the rest of the day looking how to add in muti-segment support, instead of the changing the behavior of `encoded` if you would rather... Would you like this pull request to be closed?

JakeWharton: I'm personally fine with the change of simply not canonicalizing `/` in path replacements. It makes sense in OkHttp's HttpUrl since its APIs explicitly deal with path segments. Retrofit deals with entire paths as a single unit, and the case where someone wants to dynamically determine sections of the path which may span multiple segments is not as uncommon as I would like.

jakeswenson:
@swankjesse @JakeWharton I've reverted the refactorings, and changed the condition to canonicalize if `!alreadyEncoded`. Please have a look when you have a chance.

baoti:
I prefer split one `encoded` parameter to two.

JakeWharton:
Why?

baoti:
The single path segment is so clear as I like, but multi segments support is good to some special API.

jakeswenson:
I believe that a multi-segment parameter is different then this PR. This pr is about allowing "raw" already encoded parameter values, without further canonicalization.

I think we should take the discussion of how to make multi segment easier into the bug.

JakeWharton:
What use case is there for pre-encoded path replacement that includes unencoded `/` characters that need canonicalized (aka escaped)? I can think of none.

jakeswenson:
We have a use case when calling a structured metadata API. You can ask for all metadata can get all item metadata using `/blobs/{id}/metadata` or you can access a subset of the structured metadata using `/blobs/{id}/metadata/{metadataPath}`.
This first would return:
```
{
    "foo": { "bar": "baz" }
}
```

Using the second like `/blobs/1/metadata/foo/bar` returns:
```
"baz"
```

Further it seems as though retrofit shouldn't be setting server requirements, and should allow for the communication to server apis like this and other unpredictable needs.

jakeswenson:
Using default retrofit we were able to write a type like:
```
class MetadataPath {
    private final List<String> parts;
    private MetadataPath(List<String> parts) { this.parts = parts; }
    
    public static MetadataPath of(String ...parts) { return new MetadataPath(Arrays.asList(parts)); }

    public String toString() { return String.join("/", urlEncodeEach(parts)); }

    public static Iterable<String> urlEncodeEach(Iterable<String> pathParts) { return ... }
}
```

and use it like:
```
@PUT("api/metadata/{metadataPath}")
Call<ResponseBody> setMetadata(
            @Path(value = "metadataPath", encoded = true) MetadataPath metadataPath,
            @Body Object body);
```

jakeswenson:
Example of a need for something like this: https://www.consul.io/intro/getting-started/kv.html
Consul has a key/value storage that you need to craft nested keys to lookup values.

JakeWharton:
I'm not questioning the need for the change.

jakeswenson:
I think the only further discussion before resolving the issue would be whether Retrofit can make writing that `MetadataPath` class above easier (buy using some form of `Iterable<String>` etc.)

That is a change that can come in 2.1 as #1364 is currently scheduled for. 

The encoding fix here allows a workaround, and hopefully prevents future misunderstandings. I believe It should come as a part of 2.0, as I don't believe it should be considered a minor version change.

JakeWharton:
This issue resolves that one, so there's nothing else to be done.

jakeswenson:
I'm fine with that. @swankjesse @JakeWharton any other discussion regarding this PR?

swankjesse:
I think this is going to cause undefined behavior when `encoded` is true and the input string contains a character like `?`. In particular, that’ll cause the characters after `?` to become a part of the URL query, which I claim is undesirable.

What we might want instead is another option like `singleSegment` which is either true to turn `foo/bar` into `foo%2Fbar` or false to keep it as `foo/bar`.

@JakeWharton ?

JakeWharton:
Yes, eliminating the entire encoding set in the encoded=true case is wrong.

That said, I'm still struggling to see why we should canonicalize `/` in the encoded=true case instead of just letting it cross path segment boundaries. I think the existing (as in master) behavior makes sense for OkHttp, but Retrofit doesn't have as granular of an API and no model for a segment list so it should by default allow crossing the segment path in the encoded=true case.

swankjesse:
Works for me. So the fix is to never convert `/` into `%2F`. That works for me.

jakeswenson:
Never convert `/` if `encoded = true` 
I'll update the pr

swankjesse:
Actually I think the PR should be to never convert `/` ever.

JakeWharton:
I think that would result in surprising behavior in the default encoded=false case.

swankjesse:
We discussed offline. I’m on the same page & get the rules now...

The rule is:
1. `encoded=true` means that the caller has encoded ’em.  `/` and `%` characters are left as-is (which can be used to build multiple-segments),
2. `encoded=false` means they’re encoded by Retrofit,  which will always yield a single segment

retrofit/src/main/java/retrofit2/RequestBuilder.java
```diff
@@ -85,20 +85,21 @@ void addPathParam(String name, String value, boolean encoded) {
...
     int codePoint;
     for (int i = 0, limit = input.length(); i < limit; i += Character.charCount(codePoint)) {
       codePoint = input.codePointAt(i);
-      if (codePoint < 0x20 || codePoint >= 0x7f
-          || PATH_SEGMENT_ENCODE_SET.indexOf(codePoint) != -1
-          || (codePoint == '%' && !alreadyEncoded)) {
+      if (codePoint < 0x20
+          || codePoint >= 0x7f
+          || (!alreadyEncoded
+              && (PATH_SEGMENT_ENCODE_SET.indexOf(codePoint) != -1 || codePoint == '%'))) {
```
swankjesse: I think this might return an unexpected value for say `?` and `#`

jakeswenson:
Excellent summary @swankjesse. PR updated. Thanks! to both for the quick iterations!

weitzj:
I am sorry to hijack this issue, but I am struggling to URLEncode a dot `.`
I was expecting for the default `encoded=false` annotation to encode my dot as a `%2E` (Whitespaces get properly encoded to `%20`)

My search interface:

```
@GET("search/{payload}")
Observable<List<SearchResult>> search(@Path("payload") String stringToSearch);
```

Searching for `Foo Bar.` should yield:

`search/Foo%20Bar%2E`

but it yields to:

`search/Foo%20Bar.`


##### Notes:

- OKHTTP_VERSION = '3.0.0-RC1'
-  RETROFIT_VERSION = '2.0.0-beta3'

JakeWharton:
Squashed and merged as 01d6fb1a133cea6fac18f2ef1ecab1cbee86b569

PR #1570, opened 2016-02-04
Checking header

bloder:
Hi! First of all, I need to say you that Retrofit and his code it's amazing, and I would like to contribute with a little change in code, I don't know if it's the best way to do this but I would like to check a header then i did this, Thank you!

### DIFF
--- a/retrofit/src/main/java/retrofit2/Response.java
+++ b/retrofit/src/main/java/retrofit2/Response.java
@@ -97,6 +97,11 @@ public final class Response<T> {
     return rawResponse.headers();
   }
 
+  /** Check HTTP header */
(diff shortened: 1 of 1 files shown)


JakeWharton:
I think you probably want to add a `boolean contains(String key)` to OkHttp instead (which owns the `Headers` type). Then you can just do `response.headers().contains("Whatever")` from the Retrofit `Response`.

Please file an issue or send a pull requests over on OkHttp.

JakeWharton:
Or put another way, this is far too specific of an API to warrant including on Response. Compared to the other APIs it doesn't hold its weight. Few will actually want to solely check for the presence of a header.

PR #1573, opened 2016-02-07
OkHttp 3.1.

JakeWharton:


swankjesse:
I think we're gonna need a 3.1.1 asap to cope with the performance regression in 3.1.0.

Issue #1584, opened 2016-02-10
RxJavaCallAfapterFactory does not recognize rx.Single after proguarding


krzysiekbielicki:
After proguarding application which uses retrofit service like

```
interface Service {
  @GET("/something")
  Single<String> getSomething();
}
```

retrofit fails with

```
java.lang.IllegalArgumentException: Could not locate call adapter for e.r<java.lang.String>.
                                                   Tried:
                                                    * retrofit2.adapter.rxjava.RxJavaCallAdapterFactory
                                                    * retrofit2.ExecutorCallAdapterFactory
                                                     at retrofit2.Retrofit.nextCallAdapter(Unknown Source)
                                                     at retrofit2.Retrofit.callAdapter(Unknown Source)
```

JakeWharton:
`Single` is `@Beta` and as such [can be broken or removed at any time][1]. Retrofit will not put its presence in the critical path for all RxJava support until it graduates to stable.

 [1]: https://github.com/ReactiveX/RxJava#beta

JakeWharton:
You can tell ProGuard to not obfuscate this class for now.

david-perez:
Having this issue. `Single` is no longer `@Beta`.

JakeWharton:
And it's already been adjusted on `master` accordingly.

david-perez:
I'm sorry, was using `2.0.0-beta4`. I can't use latest version. How do I tell ProGuard not to obfuscate the class then?

```
-keep class retrofit2.adapter.rxjava.RxJavaCallAdapterFactory { *; }
```

does not seem to do the trick. Code still goes through [this execution path](https://github.com/square/retrofit/blob/2613d174be2e6f31f4f742e040412aee308849cf/retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/RxJavaCallAdapterFactory.java#L51) with the above configuration.

JakeWharton:
-keepnames class rx.Single

PR #1585, opened 2016-02-10
Rely on rx.Single class instead of className String in RxJavaCallAdapterFactory

krzysiekbielicki:
Fixes issue #1584
Rely on class not on string.
Class may be proguarded and String becomes out of date.


kkocel:
:+1:

JakeWharton:
Commented on the issue.

PR #1609, opened 2016-02-18
Add lenient option for MoshiConverterFactory

### LINKED ISSUE #1592: MoshiConverterFactory should have a lenient option
https://github.com/square/moshi/issues/132

NightlyNexus:
Closes #1592.

### DIFF
--- a/retrofit-converters/moshi/src/main/java/retrofit2/converter/moshi/MoshiConverterFactory.java
+++ b/retrofit-converters/moshi/src/main/java/retrofit2/converter/moshi/MoshiConverterFactory.java
@@ -40,20 +40,29 @@ public final class MoshiConverterFactory extends Converter.Factory {
 
   /** Create an instance using {@code moshi} for conversion. */
   public static MoshiConverterFactory create(Moshi moshi) {
-    return new MoshiConverterFactory(moshi);
+    return new MoshiConverterFactory(moshi, false);
   }
 
   private final Moshi moshi;
+  private final boolean lenient;
 
-  private MoshiConverterFactory(Moshi moshi) {
+  private MoshiConverterFactory(Moshi moshi, boolean lenient) {
     if (moshi == null) throw new NullPointerException("moshi == null");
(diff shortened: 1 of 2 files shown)


NightlyNexus:
Is this what you had in mind for #1592?

retrofit-converters/moshi/src/main/java/retrofit2/converter/moshi/MoshiConverterFactory.java
```diff
@@ -40,20 +40,36 @@ public static MoshiConverterFactory create() {
 
   /** Create an instance using {@code moshi} for conversion. */
   public static MoshiConverterFactory create(Moshi moshi) {
-    return new MoshiConverterFactory(moshi);
+    return create(moshi, false);
+  }
+
+  private static MoshiConverterFactory create(Moshi moshi, boolean lenient) {
```
JakeWharton: This doesn't need to exist. Just call the constructor.

retrofit-converters/moshi/src/main/java/retrofit2/converter/moshi/MoshiConverterFactory.java
```diff
@@ -40,20 +40,36 @@ public static MoshiConverterFactory create() {
...
 
-  private MoshiConverterFactory(Moshi moshi) {
+  private MoshiConverterFactory(Moshi moshi, boolean lenient) {
     if (moshi == null) throw new NullPointerException("moshi == null");
     this.moshi = moshi;
+    this.lenient = lenient;
+  }
+
+  public MoshiConverterFactory asLenient() {
+    if (lenient) {
```
JakeWharton: I would just delete this `if`. It'll never happen and even if it does it's not that big of a deal to create a new instance.

JakeWharton:
LGTM otherwise

JakeWharton:
Merged as fb94144db40f8d403ba57799b61e9547cb349b34. Thanks.

PR #1610, opened 2016-02-18
Fix cancellation typos

NightlyNexus:


JakeWharton:
Both are technically valid.

NightlyNexus:
Oh, shoot, my bad, sorry. I was too trusting of Intellij's dictionary. You're right. :sheep:

PR #1619, opened 2016-02-20
(feat) Add success(body, headers) to Response.

hastebrot:
**Note:** I had to add an explicit cast to `successWithNullRawResponseThrows()` in `ResponseTest`.

### DIFF
--- a/retrofit/src/test/java/retrofit2/ResponseTest.java
+++ b/retrofit/src/test/java/retrofit2/ResponseTest.java
@@ -57,12 +57,13 @@ public final class ResponseTest {
 
   @Test public void successWithHeaders() {
     Object body = new Object();
-    Response<Object> response = Response.success(body, Headers.of("foo", "bar"));
+    Headers headers = Headers.of("foo", "bar");
+    Response<Object> response = Response.success(body, headers);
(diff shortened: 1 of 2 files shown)


retrofit/src/test/java/retrofit2/ResponseTest.java
```diff
@@ -54,6 +55,28 @@
     assertThat(response.body()).isNull();
   }
 
+  @Test public void successWithHeaders() {
+    Object body = new Object();
+    Response<Object> response = Response.success(body, Headers.of("foo", "bar"));
```
JakeWharton: I'd prefer if you put the `Headers` here in a local and did `isSameAs` just like is being done with `body`.

JakeWharton:
Looks great. Just one small comment.

JakeWharton:
Ah we can't use `isSameAs` on the `Headers` object because it's defensively copied. Forgot about that. I updated it to compare with `.toMultimap()` and `isEqualTo` which provides the closest approximation to verifying the contents are the same.

Squashed and merged as 85186ec1b5232263fe0ecbda790f4911215658f6

PR #1625, opened 2016-02-24
Bumped Jackson to current version 2.7.1

s3xy4ngyc:
This is needed to Support multiple names for the same class on Subtypes.

See https://github.com/FasterXML/jackson-databind/issues/1139 for more details and code samples


JakeWharton:
FWIW you can always specify a newer version of Jackson yourself.

PR #1635, opened 2016-02-28
update config proguard for retrofit 2

quangson8128:

### DIFF
--- a/website/index.html
+++ b/website/index.html
@@ -178,6 +178,9 @@ compile 'com.squareup.retrofit2:retrofit:<span class="version pln"><em>(insert l
 -keep class retrofit2.** { *; }
 -keepattributes Signature
 -keepattributes Exceptions
+-keepclasseswithmembers class * {
+    @retrofit2.http.* <methods>;
+}
 </pre>
             </section>
 


JakeWharton:
What does this add? It looks like you're saying service interfaces were being stripped which should be impossible since you'd be using them in your application code.

quangson91:
Yes, 
When I using retrofit2 + proguard I have problem:
Eg: Api define.
```java
@GET("app/authenticate")
    Observable<AuthenticationModel> authenticate(
            @Query("email") String email,
            @Query("password") String password
    );
```
After package to APK & compile project then decompile apk:

```java
@GET("app/authenticate")
    rx.a<AuthenticationModel> a();
```
You see all parameter removed.
But after I added:
```
-keepclasseswithmembers class * {
    @retrofit2.http.* <methods>;
}
```
then retrofit2 working normally.

Notes:
On my project I use:
            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
instead:
            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'


Please check my question on SO:
https://stackoverflow.com/questions/35672922/retrofit2-proguard-remove-param

Thanks.

quangson91:
I checked, problem is because I use:

`proguardFiles getDefaultProguardFile('proguard-android-optimize.txt')`

JakeWharton:
I'm going to leave this out since these aggressive optimizations are not recommended for beginners anyway. If many people start to complain that they're being tripped up with this I'll re-consider adding it back though.

PR #1636, opened 2016-02-28
Support for primitive return types

pa314159:
If the Call pattern was not used, methods like

	@GET("...") int getValue()

were not handled properly.


JakeWharton:
Thanks! Tweaked and merged manually.

PR #1639, opened 2016-03-02
Propagate callback executor even when not explicitly specified.

### LINKED ISSUE #1632: Platform callback executor not propagated to Retrofit instance.
It would be nice if a custom `CallAdapter.Factory` could have access to the same callbackExecutor that the default `ExecutorCallAdapterFactory` has (with logic to use proper thread for Android and non-Android).

JakeWharton:
This lets custom CallAdapter implementations use the platform default for their callbacks.

Closes #1632.

### DIFF
--- a/retrofit/src/main/java/retrofit2/Retrofit.java
+++ b/retrofit/src/main/java/retrofit2/Retrofit.java
@@ -358,7 +358,10 @@ public final class Retrofit {
     return (Converter<T, String>) BuiltInConverters.ToStringConverter.INSTANCE;
   }
 
-  /** The executor used for {@link Callback} methods on a {@link Call}. */
+  /**
+   * The executor used for {@link Callback} methods on a {@link Call}. This may be {@code null},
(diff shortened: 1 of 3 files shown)


retrofit/src/main/java/retrofit2/Retrofit.java
```diff
@@ -545,9 +554,14 @@ public Retrofit build() {
         callFactory = new OkHttpClient();
       }
 
+      Executor callbackExecutor = this.callbackExecutor;
+      if (callbackExecutor == null) {
+        callbackExecutor = platform.defaultCallbackExecutor();
```
swankjesse: could still be null?

JakeWharton: yep. intentional. should I document?

swankjesse: nope, that’s fine

PR #1641, opened 2016-03-02
Add body/part/partmap option to pass nulls to underlying converter.

### LINKED ISSUE #1488: Cannot pass null to @Body
I have a call which gets "@Body RequestBody body"
I dont have always a need to pass body to it nd i cannot use null.

JakeWharton:
This also changes the default behavior of null passed to a body parameter to become an empty body.

Closes #1488.

### DIFF
--- a/retrofit-converters/gson/src/test/java/retrofit2/converter/gson/GsonConverterFactoryTest.java
+++ b/retrofit-converters/gson/src/test/java/retrofit2/converter/gson/GsonConverterFactoryTest.java
@@ -78,7 +78,7 @@ public final class GsonConverterFactoryTest {
   }
 
   interface Service {
-    @POST("/") Call<AnImplementation> anImplementation(@Body AnImplementation impl);
+    @POST("/") Call<AnImplementation> anImplementation(@Body(ignoreNull = false) AnImplementation impl);
     @POST("/") Call<AnInterface> anInterface(@Body AnInterface impl);
   }
 
@@ -131,6 +131,16 @@ public final class GsonConverterFactoryTest {
 
     RecordedRequest request = server.takeRequest();
     assertThat(request.getBody().readUtf8()).isEqualTo("{}"); // Null value was not serialized.
+    assertThat(request.getHeader("Content-Type")).isEqualTo("application/json; charset=UTF-8");
+  }
+
+  @Test public void serializeNull() throws IOException, InterruptedException {
+    server.enqueue(new MockResponse().setBody("{}"));
+
+    service.anImplementation(null).execute();
+
+    RecordedRequest request = server.takeRequest();
+    assertThat(request.getBody().readUtf8()).isEqualTo("null"); // Top-level null literal.
     assertThat(request.getHeader("Content-Type")).isEqualTo("application/json; charset=UTF-8");
   }
 
--- a/retrofit-converters/scalars/src/main/java/retrofit2/converter/scalars/ScalarRequestBodyConverter.java
+++ b/retrofit-converters/scalars/src/main/java/retrofit2/converter/scalars/ScalarRequestBodyConverter.java
@@ -23,11 +23,15 @@ import retrofit2.Converter;
 final class ScalarRequestBodyConverter<T> implements Converter<T, RequestBody> {
   static final ScalarRequestBodyConverter<Object> INSTANCE = new ScalarRequestBodyConverter<>();
   private static final MediaType MEDIA_TYPE = MediaType.parse("text/plain; charset=UTF-8");
+  private static final RequestBody EMPTY_BODY = RequestBody.create(null, new byte[0]);
 
   private ScalarRequestBodyConverter() {
   }
 
   @Override public RequestBody convert(T value) throws IOException {
+    if (value == null) {
+      return EMPTY_BODY;
+    }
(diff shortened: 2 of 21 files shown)


retrofit-converters/scalars/src/test/java/retrofit2/converter/scalars/ScalarsConverterFactoryTest.java
```diff
@@ -38,7 +37,7 @@
   interface Service {
       @POST("/") Call<ResponseBody> object(@Body Object body);
 
-      @POST("/") Call<ResponseBody> stringObject(@Body String body);
+      @POST("/") Call<ResponseBody> stringObject(@Body(ignoreNull = false) String body);
       @POST("/") Call<ResponseBody> booleanPrimitive(@Body boolean body);
       @POST("/") Call<ResponseBody> booleanObject(@Body Boolean body);
```
ndorigatti: Shoudn't be @Body(ignoreNull = false) also on Boolean object?

JakeWharton: `stringObject` is the only method tested with `null`.

retrofit-converters/gson/src/test/java/retrofit2/converter/gson/GsonConverterFactoryTest.java
```diff
@@ -78,7 +78,7 @@
   }
 
   interface Service {
-    @POST("/") Call<AnImplementation> anImplementation(@Body AnImplementation impl);
+    @POST("/") Call<AnImplementation> anImplementation(@Body(ignoreNull = false) AnImplementation impl);
```
swankjesse: this is kind of hard to read ’cause of the double negative. Should it be `nullable = true` ?

JakeWharton: It's more about what to do when we receive `null` than whether `null` is accepted. I started out with only allow/disallow but it doesn't cover a lot of common cases.

retrofit-converters/scalars/src/main/java/retrofit2/converter/scalars/ScalarRequestBodyConverter.java
```diff
@@ -23,11 +23,15 @@
 final class ScalarRequestBodyConverter<T> implements Converter<T, RequestBody> {
   static final ScalarRequestBodyConverter<Object> INSTANCE = new ScalarRequestBodyConverter<>();
   private static final MediaType MEDIA_TYPE = MediaType.parse("text/plain; charset=UTF-8");
+  private static final RequestBody EMPTY_BODY = RequestBody.create(null, new byte[0]);
 
   private ScalarRequestBodyConverter() {
   }
 
   @Override public RequestBody convert(T value) throws IOException {
+    if (value == null) {
```
swankjesse: I think this behavior is pretty strange. Is it essential?

The problem is that `""` is a valid string, but not a valid integer.

retrofit-converters/gson/src/test/java/retrofit2/converter/gson/GsonConverterFactoryTest.java
JakeWharton: The boolean controls whether or not `null` is passed to the converter or whether it just becomes an empty body and skips the converter.

Maybe `convertNull` then?

retrofit/src/main/java/retrofit2/RequestAction.java
```diff
@@ -244,21 +250,24 @@
...
-      if (value == null) {
-        throw new IllegalArgumentException("Body parameter value must not be null.");
-      }
       RequestBody body;
-      try {
-        body = converter.convert(value);
-      } catch (IOException e) {
-        throw new RuntimeException("Unable to convert " + value + " to RequestBody", e);
+      if (ignoreNull && value == null) {
+        body = Utils.EMPTY_BODY;
```
swankjesse: shouldn't we ask the converter to convert it?

swankjesse: (So we get `"null"` for JSON?)

retrofit/src/main/java/retrofit2/http/Body.java
```diff
@@ -39,4 +39,9 @@
 @Target(PARAMETER)
 @Retention(RUNTIME)
 public @interface Body {
+  /**
+   * When true a {@code null} value will be ignored and an empty body will be used. Otherwise
```
swankjesse: Ahh, so we have three cases:

 1. non-null
 2. null that maps to empty body
 3. null that is converted by the converter

Would it makes sense to punt case 2 to the converter?

retrofit/src/main/java/retrofit2/http/PartMap.java
```diff
@@ -51,4 +51,10 @@
 public @interface PartMap {
   /** The {@code Content-Transfer-Encoding} of the parts. */
   String encoding() default "binary";
+
+  /**
+   * When true a {@code null} values will cause their part to be omitted. Otherwise {@code null}
```
swankjesse: nit: `a`

JakeWharton:
Discussed offline. New PR soon.

PR #1647, opened 2016-03-04
Forbid null map values, and null maps.

JakeWharton:
As explained in #1488.


swankjesse:
Null keys?

LGTM

JakeWharton:
Already forbidden.

PR #1650, opened 2016-03-06
Combine MethodHandler and RequestFactory into ServiceMethod.

swankjesse:
This is once class that handles all of the annotations and binding for
a service method.


retrofit/src/main/java/retrofit2/ServiceMethod.java
```diff
@@ -51,62 +52,81 @@
...
+    this.baseUrl = builder.retrofit.baseUrl();
+    this.responseConverter = builder.responseConverter;
+    this.httpMethod = builder.httpMethod;
+    this.relativeUrl = builder.relativeUrl;
+    this.headers = builder.headers;
+    this.contentType = builder.contentType;
+    this.hasBody = builder.hasBody;
+    this.isFormEncoded = builder.isFormEncoded;
+    this.isMultipart = builder.isMultipart;
+    this.parameterHandlers = builder.parameterHandlers;
```
JakeWharton: All of these fields need to be made package scope. Otherwise you're adding a synthetic method for each.

JakeWharton: You can also drop all the `this` qualifiers if you want.

JakeWharton:
LGTM

PR #1652, opened 2016-03-07
Drop the mutable base URL.

swankjesse:
This was unsafe for switching between development and production
backends. This is because requests that are triggered on a UI thread
may be delayed in execution on a background thread. These could cause
requests created for a development environment to be executed against
a production environment, or vice-versa.

Retrofit is the wrong layer for switching backends for load balancing
or proximity. That can happen in an OkHttp interceptor instead.


swankjesse:
https://gist.github.com/swankjesse/8571a8207a5815cca1fb

artem-zinnatullin:
Whoa… Well, you're right that it's not safe to change url in Retrofit, but OkHttp interceptor is not a great place for that even more because not only it has the same problem with delayed requests but also you can't understand that request is coming from Retrofit inside of the interceptor (analyzing stacktrace is not an option)… Imagine you also have non-REST requests to the api endpoint that you do through same OkHttp instance without Retrofit.

Would you suggest to rebuild the `Retrofit` instance with new url and use something like `Provider<Retrofit>` instead of direct reference to the `Retrofit` or even rebuild whole graph of network dependencies to change endpoint in the application?

JakeWharton:
That has always been the recommendation, yes. `BaseUrl` existed solely for client-side load balancing or geographical proximity specialization of the same semantic host. Since there's no longer an HTTP client abstraction, we no longer need the behavior to exist in Retrofit.

artem-zinnatullin:
ok, thanks for confirmation. 

// Anyway, :+1: :+1: for more immutability, though `BaseUrl` was so convenient for functional testing with mock webserver…

swankjesse:
Yep. This new interceptor is the new best way to test with a mock webserver.

jbaginski:
in my case baseUrl contains a path segment. I don't see any other way than manually replace the old baseUrl with the new one...
@swankjesse do you know any other way to change the baseUrl (with path) using interceptor mechanism?

swankjesse:
@jbaginski you can rewrite the URL however you like. One option is to extract out the path & query in an interceptor, and add those on to the base URL.

jbaginski:
thanks @swankjesse, I ended up manipulating a string directly. it's not the best solution but it works ;)

JessYanCoding:
This solution will affect @GET (full path) @POST (full path) and @Url so that they can not work,try my solution:  https://github.com/JessYanCoding/RetrofitUrlManager

PR #1653, opened 2016-03-07
Doc error. With `encoded=true` will stop default url encoding, but your doc is misleading.

walfud:
Doc error. So, calling with `foo.list(ImmutableMap.of("foo", "foo+bar"))` yields '/search?foo=foo+bar' instead of '/search?foo=foo%2Bbar'.


swankjesse:
Thanks!

PR #1663, opened 2016-03-09
add converter for fastjson

wawlian:
[fastjson](https://github.com/alibaba/fastjson) is a fast and powerful json processor which is widely used by Chinese developers. This pull request adds an fastjson converter.  We have used this converter in our products well.


JakeWharton:
We're not going to add new converters. See https://github.com/square/retrofit/pull/988#issuecomment-148282283 for details and what to do instead.

PR #1666, opened 2016-03-10
Add support for OkHttp's MultipartBody.Part.

### LINKED ISSUE #1140: Add a way to set filename on multipart requests

JakeWharton:
To do this we need to make `@Part`'s value (i.e., name) optional, since `MultipartBody.Part` brings its name.

`MultipartBody.Part` is a replacement for Retrofit 1's `TypedOutput`/`TypedFile` for doing uploads of files with names.

Closes #1140. Closes #1467.

### DIFF
--- a/retrofit/src/main/java/retrofit2/ServiceMethod.java
+++ b/retrofit/src/main/java/retrofit2/ServiceMethod.java
@@ -29,6 +29,7 @@ import java.util.regex.Pattern;
 import okhttp3.Headers;
 import okhttp3.HttpUrl;
 import okhttp3.MediaType;
+import okhttp3.MultipartBody;
 import okhttp3.Request;
 import okhttp3.RequestBody;
 import okhttp3.ResponseBody;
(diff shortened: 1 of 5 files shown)


retrofit/src/main/java/retrofit2/ServiceMethod.java
```diff
@@ -535,33 +536,47 @@ private Headers parseHeaders(String[] headers) {
...
-              retrofit.requestBodyConverter(arrayComponentType, annotations, methodAnnotations);
-          return new ParameterHandler.Part<>(headers, converter).array();
+
+          return ParameterHandler.RawPart.INSTANCE;
         } else {
-          Converter<?, RequestBody> converter =
-              retrofit.requestBodyConverter(type, annotations, methodAnnotations);
-          return new ParameterHandler.Part<>(headers, converter);
+          Headers headers =
+              Headers.of("Content-Disposition", "form-data; name=\"" + partName + "\"",
```
swankjesse: Probably want to eventually restrict this to not contain `"`, etc.
Or escape it somehow

retrofit/src/test/java/retrofit2/RequestBuilderTest.java
```diff
@@ -1393,6 +1394,97 @@
...
+      @Multipart //
+      @POST("/foo/bar/") //
+      Call<ResponseBody> method(@Part RequestBody part) {
+        return null;
+      }
+    }
+
+    try {
+      buildRequest(Example.class, new Object[] { null });
+      fail();
```
swankjesse: 🍩

retrofit/src/test/java/retrofit2/RequestBuilderTest.java
```diff
@@ -1393,6 +1394,97 @@
...
+      @POST("/foo/bar/") //
+      Call<ResponseBody> method(@Part MultipartBody.Part part) {
+        return null;
+      }
+    }
+
+    MultipartBody.Part part = MultipartBody.Part.createFormData("kit", "kat");
+    Request request = buildRequest(Example.class, part);
+    assertThat(request.method()).isEqualTo("POST");
+    assertThat(request.headers().size()).isZero();
```
swankjesse: Neat

retrofit/src/main/java/retrofit2/http/Part.java
```diff
@@ -26,12 +26,16 @@
...
+ * <li>If the type is {@link okhttp3.MultipartBody.Part} the contents will be used directly. Omit
+ * the name from the annotation (i.e., {@code @Part MultipartBody.Part part}).</li>
  * <li>If the type is {@link okhttp3.RequestBody RequestBody} the value will be used
- * directly with its content type.</li>
+ * directly with its content type. Supply the part name in the annotation (e.g.,
+ * {@code @Part("foo") RequestBody foo}).</li>
  * <li>Other object types will be converted to an appropriate representation by using
- * {@linkplain Converter a converter}.</li>
+ * {@linkplain Converter a converter}. Supply the part name in the annotation (e.g.,
+ * {@code @Part("foo") RequestBody foo}).</li>
```
baoti: May not be `RequestBody` ?

JakeWharton: Good catch, thanks. 210be051ddbbd3383f02d51e3d15f021fb0aac48

PR #1673, opened 2016-03-14
Retrofit web crawler example.

swankjesse:


JakeWharton:
Fun, although not the best usage of Retrofit one could argue. The converter part is probably the coolest takeaway here.

swankjesse:
Yeah. I was trying to think of other fun stuff to scrape with Retrofit & JSoup.

PR #1703, opened 2016-03-30
Support iterable and array-based OkHttp Part parts.

### LINKED ISSUE #1700: Feature request: multiple file uploading via @Part List<MultipartBody.Part> or @PartMap<String, MultipartBody.Part>
[Feature Request] 
Since 2.0.0, setting a filename for an upload file is resolved by OkHttp's MultipartBody.Part,
but the support limited to one file per param via (@Part MultipartBody.Part part).

Multiple file uploading via @PartMap or @Part List<> is still unsupported.

There're two approaches to achieve that.

1) Support @Part List&lt;MultipartBody.Part> as method parameter
When parsing @Part in ServiceMethod.parseParameterAnnotation, allow the parameterizedType of Iterable to be MultipartBody.Part and allow each part setting its OWN headers, instead of setting a common header for all parts.

2) Support @PartMap Map&lt;String, MultipartBody.Part> as method parameter
When parsing @PartMap in ServiceMethod.parseParameterAnnotation, allow the parameterizedType of Iterable to be MultipartBody.Part.
I noticed that, in the code block parsing @PartMap, an exception will be thrown when parameterizedType is detected to be MultipartBody.Part. 
The exception message is "@PartMap values cannot be MultipartBody.Part. Use @Part List&lt;Part> or a different value type instead."
Sadly @Part List&lt;Part> is also not supported yet. I supposed it should be @Part List&lt;MultipartBody.Part>.

I prefer to the first approach, because I sometimes need to upload multiple files with same field name (but possibly different filenames), which can't be done with @PartMap unless multimap type support is added.

Thanks.

JakeWharton:
Also forbid OkHttp Part parts in iterables and arrays when a part name was specified.

Closes #1700.

### DIFF
--- a/retrofit/src/main/java/retrofit2/ServiceMethod.java
+++ b/retrofit/src/main/java/retrofit2/ServiceMethod.java
@@ -540,12 +540,33 @@ final class ServiceMethod<T> {
         String partName = part.value();
         Class<?> rawParameterType = Utils.getRawType(type);
         if (partName.isEmpty()) {
-          if (!MultipartBody.Part.class.isAssignableFrom(rawParameterType)) {
+          if (Iterable.class.isAssignableFrom(rawParameterType)) {
+            if (!(type instanceof ParameterizedType)) {
+              throw parameterError(p, rawParameterType.getSimpleName()
+                  + " must include generic type (e.g., "
+                  + rawParameterType.getSimpleName()
+                  + "<String>)");
+            }
+            ParameterizedType parameterizedType = (ParameterizedType) type;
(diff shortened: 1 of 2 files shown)


WesleyWu:
Hi JakeWharton,
Thanks for the quick patch.
In RequestBuilderTest probably should add a new test multipartOkHttpIterablePartWithFilename for testing setting filenames.

JakeWharton:
File names are handled by `MultipartBody.Part` and `MultipartBody` which are OkHttp types so its behavior doesn't need tested in this library.

PR #1713, opened 2016-04-03
Eliminate some allocation/state overhead by implementing subscription directly.

JakeWharton:
`Call`'s internal `volatile boolean` for tracking canceled was already mirroring this state so it can be used directly.


swankjesse:
Nice.

PR #1715, opened 2016-04-03
Update CompletableFutureTest.java

DimitrisMalamas:
Simply added hasMessage() method to responseFailure() and bodyFailure().


JakeWharton:
There's no `hasMessage` here because the message hasn't historically been stable across OS's.

PR #1742, opened 2016-04-14
replace with foreach

chenDoInG:


JakeWharton:
This allocates an Iterator whereas the indexed for loop does not.

chenDoInG:
In the method "stringConverter"  we don't need the indexed for loop, because the index is between 0 and the sizes of loop

JakeWharton:
I know that an indexed for loop isn't needed, but it is what's desired. The for-each loop allocates an Iterator in order to traverse the list contents and pays more heavily in virtual method calls. As a library (and you'll find this in many of Square's libraries) we try to minimize the memory and performance impact incurred. It's small, I know, but since we do it everywhere there's loops it slowly adds up to a non-small amount.

If you want to learn more about it I gave a talk a while back which covers why we do this: http://jakewharton.com/presentation/2015-06-25-streamlining-android-apps/

chenDoInG:
I got it! thank you very much

PR #1745, opened 2016-04-16
fix Polymorph class of Jackson

ZeroBrain:
If request body is defined super class and actual body is child,
then request body is always serialized of super.
so I change to object-mapper from object-writer in converter-jackson

ex)
```java
class A { String name; }
class B extends A { int age; }

interface Service {
  Call<B> b(A a);
}
```

If I request like this
`Service.b(new B("steve", 10));`
then it was like this
```json
//expect
{"name":"steve","age":10}

// actual
{"name":"steve"}
```

It was occured by ObjectWriter.
so I change to use ObjectMapper.


ZeroBrain:
it is related to #1192 

#1192 was rejected for cached performance. I agree with you.
but consider this problem.

JakeWharton:
If you need polymorphic conversion of body types then you will need to supply your own converter factory with this behavior (which is what we said in #1192). It's rare that this is actually needed, and we don't want to make the common case slower to accommodate it.

Issue #1736, opened 2016-04-13
Add @HeadersMap


artshell:
**> First:**
```java
    public interface ApiService {
        @Headers({
                "User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/49.0.2623.110 Safari/537.36",
                "Accept: text/html,text/plain,application/json",
                "Accept-Language: zh-cn,zh,en-US,en",
                "Accept: text/html,text/plain,application/json",
                "Accept-Language: zh-cn,zh,en-US,en",
                "Accept-Charset: utf-8,GB2312",
                "Connection: keep-alive"
        })
        @GET("Api-{module}-{method}")
        Call<TokenEntity> asyncGet(@Path("module") String pModule, @Path("method") String pMethod, @QueryMap Map<String, String> pParameterOptions);
    }
```
"Headers" modify to "HeaderMap" easy reading

**> Second:**
    This feature is exist or implement new feature?

```java
    private static class SingletonHolder {
        private static final OkHttpClient S_CLIENT;
        private static final Retrofit.Builder S_BUILDER;

        static {
            OkHttpClient.Builder nClientBuilder = new OkHttpClient.Builder();
            /* 连接超时2分钟 */
            nClientBuilder.connectTimeout(2 * 60 * 1000, TimeUnit.MILLISECONDS);
            /* 数据读取超时10分钟 */
            nClientBuilder.readTimeout(10 * 60 * 1000, TimeUnit.MILLISECONDS);
            /* 数据上传超时10分钟 */
            nClientBuilder.writeTimeout(10 * 60 * 1000, TimeUnit.MILLISECONDS);
            /* 连接失败后尝试重新连接 */
            nClientBuilder.retryOnConnectionFailure(true);
            S_CLIENT = nClientBuilder.build();

            /* Retrofit结合OkHttpClient*/
            S_BUILDER = new Retrofit.Builder();
            S_BUILDER.baseUrl("http://ybirds.lp/").addConverterFactory(GsonConverterFactory.create()).validateEagerly(true).client(S_CLIENT);
        }
    }

    public static Retrofit.Builder getRetrofitBuilder(){
        return SingletonHolder.S_BUILDER;
    }

    public interface ApiService {
        @GET("Api-{module}-{method}")
        <T> Call<T> asyncGet(@Path("module") String pModule, @Path("method") String pMethod);
    }

[...]

PR #1751, opened 2016-04-20
Implemented @HeaderMap

venilnoronha:
I've created `HeaderMap` which supports parameters of type `Map` ~~and `Multimap`~~. ~~`Multimap` enables one to specify multiple header values for the same name.~~ See #1736 for more info.

I've signed the ICLA. Please review and merge.

Thanks,
Venil

### DIFF
--- a/retrofit/src/main/java/retrofit2/http/HeaderMap.java
+++ b/retrofit/src/main/java/retrofit2/http/HeaderMap.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2013 Square, Inc.
+ * Copyright (C) 2016 Square, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


retrofit/pom.xml
```diff
@@ -32,6 +32,11 @@
     </dependency>
 
     <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+    </dependency>
```
JakeWharton: We can not depend on Guava

venilnoronha: Oh okay. Should we have our own `MultiMap`? Or should we support `Map<String, Collection<?>>` instead?

JakeWharton: Just ignore that case for now, we're going to look into it as part of #626.

venilnoronha: Sure, thanks!

venilnoronha:
Fixed checkstyle issues, squashed.

JakeWharton:
Looks great at a quick glance! I'll do a full review later.

venilnoronha:
Sure, thank you!

retrofit/src/main/java/retrofit2/http/HeaderMap.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+import java.util.Map;
+
+/**
+ * Adds headers specified in the {@link Map}.
+ * <p>
+ * Values are converted to strings using {@link String#valueOf(Object)}.
```
JakeWharton: This is false, but I suspect this is copy/paste from other map annotations so I'll fix them all once this is merged.

JakeWharton:
Merged as adfe99884bfc01dd3a1164a768fef96d6550208b. Thanks again!

venilnoronha:
Thanks a lot @JakeWharton!

PR #1755, opened 2016-04-24
Portuguese Translation

viniciuslcpereira97:
A lot of Brazilians use Retrofit, so we've translated the website to portuguese.


JakeWharton:
While I appreciate you taking the time to do this, we can't merge or support translations natively since we have no way of ensuring their accuracy and no way to keep them up-to-date as the normal website changes.

PR #1789, opened 2016-05-10
Adding ProGuard lines in the Readme file

DiSiqueira:
It aims to simplify installation for new users


JakeWharton:
I don't want to have these in two places because they will get out of sync.

Issue #1786, opened 2016-05-09
POST Request with JSON-Converters


dbaudach:
I have an application that has a connection to a server. 
The server responds with JSON , I would like to convert into Java objects.

Interface looks like:
```
 @Headers({
            "Authorization: Basic YWRtaW46YWRtaW4=",
            "Content-Type: application/x-www-form-urlencoded",
            "Accept: application/sparql-results+json"
    })
    @POST("/annex/LichtScene/sparql/query")
    Observable<ScenesResponse> getScenes(@Body String userRequest);
```

When I call a POST with:

```
Retrofit retro = new Retrofit.Builder().baseUrl("http://10.0.3.2:5820/")
                .addConverterFactory(GsonConverterFactory.create())
                .addCallAdapterFactory(RxJavaCallAdapterFactory.create())
                .build();
        stardog = retro.create(Stardog.class);

        String body = "query=SELECT+%3FIDScene+WHERE+%7B+%3FIDScene+%3AhatPOI+%3FPOI+%7D";
        stardog.getScenes(body).subscribeOn(Schedulers.io()).observeOn(AndroidSchedulers.mainThread()).subscribe(new Observer<ScenesResponse>() {
            @Override
            public void onCompleted() {
            }

            @Override
            public void onError(Throwable e) {
                Timber.e(e.getMessage());
            }

            @Override
            public void onNext(ScenesResponse scenesResponse) {
            }
        });
```

I have the following problem: The request to the server is assembled incorrectly.
Wireshark output:
```
Form item: ""query" = "SELECT ?IDScene WHERE { ?IDScene :hatPOI ?POI }""
    Key: "query
    Value: SELECT ?IDScene WHERE { ?IDScene :hatPOI ?POI }"
```
There are inserted too many apostrophes.

If the return type is set to String and the converter to Scalar it works fine.
Wireshark output:
```
Form item: "query" = "SELECT ?IDScene WHERE { ?IDScene :hatPOI ?POI }"
    Key: query
    Value: SELECT ?IDScene WHERE { ?IDScene :hatPOI ?POI }
```
No extra apostrophes here.

iagreen:
[...]

PR #1795, opened 2016-05-15
Don't create gson converters for scalars

iagreen:
Using a gson converter on scalar types is almost universally an error since serializing them results in invalid JSON.  By not creating gson converters for the scalar types, code will fail faster and the error will (hopefully) be more discoverable.

Inspired by this question -- https://github.com/square/retrofit/issues/1786.


JakeWharton:
The JSON spec allows scalar literals as root objects. A bug was (erroneously) filed on Retrofit for the lack of Gson supporting this feature. I made changes in both Gson and Moshi a few months ago to support top-level scalars (Jackson already supported it) so this is not only a valid case but one that someone needed. It can be counter intuitive, but since Gson supports these types it's not Retrofit's place to restrict them.

PR #1812, opened 2016-05-23
@JSON annotation

kevinrandrup:
I'm working with an JSON API which commonly uses a root key to encode a value.
A example for a login call would be:

    {"user":{ "username":Kevin,"password":"ThePassword"}}

I have added an `@JSON` tag to set `Content-Type` to `application\json` and a `@Root` tag to provide a root key for the object. My main goal is to avoid losing type information with method that just accepts a `Map` and avoid creating a bunch of dummy objects to maintain type and key information. 

The end result for sending a LoginRequest is shown below.

```java
public class LoginRequest {
    public String email;
    public String password;
}

// Before
@Headers("Content-Type: application/json")
@POST("users/sign_in.json")
Call<User> attemptLogin(@Body Map<String, LoginRequest> loginRequest);

Map<String, LoginRequest> request = new HashMap<>();
LoginRequest loginRequest = new LoginRequest("MyUsername", "MyPassword");
request.put("user", loginRequest);
attemptLogin(request)

// After
@JSON
@POST("login")
Call<User> attemptLogin(@Root("user") LoginRequest loginRequest);

LoginRequest loginRequest = new LoginRequest("MyUsername", "MyPassword");
attemptLogin(loginRequest);
```

I have only implemented having a single `@Root` key, but here is a more complicated example from the [GitHub API](https://developer.github.com/v3/git/commits/#create-a-commit) for creating a commit that would use many `@Root` keys.
```java
{
  "message": "my commit message",
  "author": {
    "name": "Scott Chacon",
    "email": "schacon@gmail.com",
    "date": "2008-07-09T16:13:30+12:00"
  },
  "parents": [
    "7d1b31e74ee336d15cbd21741bc88a537ed063a0"
  ],
  "tree": "827efc6d56897b048c772eb4087f854f46256132"
}

@JSON
@POST("login")
Call<Map<String, String> attemptLogin(
    @Root("message") String commitMessage,
    @Root("author") Author author,
    @Root("parents") List<String> parents,
    @Root("tree") String parentSHA);
```

I have added some tests and everything passes besides the C-style arrays (`rootJSONArray()` and `rootJSONPrimitiveArray()` - verified with `mvn clean verify`). I would really appreciate some assistance implementing that (as well as any other improvements). 

I realize this fork isn't ready to be merged but I wanted to get some initial feedback/assistance.

### DIFF
--- a/retrofit/src/main/java/retrofit2/JSONBody.java
+++ b/retrofit/src/main/java/retrofit2/JSONBody.java
@@ -1,0 +1,80 @@
+package retrofit2;
+
+import java.io.IOException;
+
+import okhttp3.MediaType;
+import okhttp3.RequestBody;
+import okio.Buffer;
+import okio.BufferedSink;
+
+public class JSONBody extends RequestBody {
+  
+  private static final MediaType CONTENT_TYPE =
+      MediaType.parse("application/json");
+  private String jsonString;
+  
+  private JSONBody(String jsonString) {
+    this.jsonString = jsonString;
+  }
+
+  @Override
+  public MediaType contentType() {
+    return CONTENT_TYPE;
+  }
+
+  @Override public long contentLength() {
+    return writeOrCountBytes(null, true);
+  }
+
+  @Override public void writeTo(BufferedSink sink) throws IOException {
+    writeOrCountBytes(sink, false);
+  }
+
+  /**
+   * Either writes this request to {@code sink} or measures its content length. We have one method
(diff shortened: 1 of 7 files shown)


JakeWharton:
Retrofit is serialization format agnostic and so we can't add anything JSON-specific in the core.

You don't need any support, though, as this can be implemented already using Retrofit's existing mechanisms. The nearest example is JSON RPC for which you can find an implementation for inspiration here: https://github.com/segmentio/retrofit-jsonrpc.

PR #1818, opened 2016-05-25
Update okhttp to 3.3.0

sparty02:
Update to latest okhttp release.  Among other things, this has improved support for properly handling TrustManagers on various JVMs.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,7 +49,7 @@
 
     <!-- Dependencies -->
     <android.version>4.1.1.4</android.version>
-    <okhttp.version>3.2.0</okhttp.version>
+    <okhttp.version>3.3.0</okhttp.version>
     <animal.sniffer.version>1.14</animal.sniffer.version>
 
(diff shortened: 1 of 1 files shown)


JakeWharton:
Thanks. But FYI, we will never make a release for things like this, so if you want to use 3.3.0 you'll need to declare the newer dependency manually in your build for now.

sparty02:
Understood.  I just wasn't sure what the strategy was for updating deps and wanted to make sure this was in there for the next retrofit release.

JakeWharton:
Yep. We definitely want to track them here to help force people onto the
latest if they otherwise don't update. But just making sure I set
expectations that there wouldn't be an immediate release containing this.

artem-zinnatullin:
Another reason to do this is to run tests with newer deps. For example in
this case: to check that Retrofit works without problems with new OkHttp.
Same with RxBinding + RxJava, etc.

PR #1837, opened 2016-06-03
second wait on callbacks

anasanasanas:


JakeWharton:
Because this is test code we relax on conventions like using constants for shared values. It's not super important that these all actually be the same value.

Thanks for the PR though.

PR #1865, opened 2016-06-19
Allows Observable and Completable to be transformed

tsuijten:
Allows Observable and Completable to be transformed before returning then from RxJavaCallAdapterFactory.


JayNewstrom:
Looks like the build failed because of style related issues. See Squares code style here: https://github.com/square/java-code-styles

JakeWharton:
This would need to support `Single` as well.

retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/Transformer.java
```diff
@@ -0,0 +1,15 @@
...
+import rx.Completable;
+import rx.Observable;
+
+/**
+ * A class that allows {@link Observable} and {@link Completable} to be transformed before
+ * returning then from {@link RxJavaCallAdapterFactory}.
+ *
+ * For example, you can use the {@link Transformer} to apply a observeOn or subscribeOn {@link rx.Scheduler}
+ */
+public interface Transformer {
```
JakeWharton: Not a huge fan of needing our own type when the concept exists in RxJava already. Let me think about this a bit.

tsuijten: I agree, we could also go for a `TransformerProvider` which returns a Rx `Transformer` and a `CompletableTransformer` which we apply using `compose` on the `Observable` and `Completable`

tsuijten: PS `Single` is supported because `RxJavaCallAdapterFactory` uses the same `CallAdapter` as used by `Observable` and just calls `toSingle` on it

tsuijten:
I refactored to a `TransformerProvider` adds boilerplate and basically does the same as before, but now we don't have a custom `Transformer` anymore and just use Rx's `compose` method.

`Single` is supported because `RxJavaCallAdapterFactory` uses the same `CallAdapter` as used by `Observable` and just calls `toSingle` on it. If you'd want a custom provider for `Single` I'd have to make some changes.

koperko:
Are there any news regarding this ? It would be an amazing boilerplate mitigator.. Will it get to some version ?

4gus71n:
Same as @koperko did this changes made it into any release?

JakeWharton:
Nope, there's no need for it: https://github.com/square/retrofit/blob/bd8e38d406fff637ed69b1b312a8e439aa28b012/samples/src/main/java/com/example/retrofit/RxJavaObserveOnMainThread.java

vidhithakrar:
@JakeWharton I think the Transformer still makes sense in case we want to bind the Observable with Activity/Fragment's lifecycle.

PR #1871, opened 2016-06-21
Fix test to not rely on the order

alexgyori:
The test assumes that the order in which annotations are returned by the JVM are the same across all implementations and is consistent with the declaration order; this assumption does not hold.

This fixes the assumption by allowing any of the two possible orders to appear in the exception message.


JakeWharton:
Thanks!

PR #1877, opened 2016-06-23
Avoid leak on newJsonReader exception

frapontillo:
#### Problem

When converting a `ResponseBody` with the `GsonResponseBodyConverter`, if `gson.newJsonReader` throws an unchecked exception (such as `NullPointerException`), the body will remain open and will never be closed.

This was introduced in 473fd2473cc41e68eb61c6d2d4316d720edd46fb.

#### Solution

Mirroring what's in `ProtoResponseBodyConverter`, I have moved the creation of the `JsonReader` inside the `try` block.
This will ensure that after opening the body stream, it will always be closed.

CLA has been signed.

### DIFF
--- a/retrofit-converters/gson/src/main/java/retrofit2/converter/gson/GsonResponseBodyConverter.java
+++ b/retrofit-converters/gson/src/main/java/retrofit2/converter/gson/GsonResponseBodyConverter.java
@@ -18,7 +18,10 @@ package retrofit2.converter.gson;
 import com.google.gson.Gson;
 import com.google.gson.TypeAdapter;
 import com.google.gson.stream.JsonReader;
+
 import java.io.IOException;
+import java.io.Reader;
+
 import okhttp3.ResponseBody;
 import retrofit2.Converter;
 
@@ -32,8 +35,9 @@ final class GsonResponseBodyConverter<T> implements Converter<ResponseBody, T> {
   }
 
   @Override public T convert(ResponseBody value) throws IOException {
-    JsonReader jsonReader = gson.newJsonReader(value.charStream());
+    Reader charStream = value.charStream();
     try {
+      JsonReader jsonReader = gson.newJsonReader(charStream);
       return adapter.read(jsonReader);
     } finally {
       value.close();


swankjesse:
I don’t see the benefit. A program that’s throwing unexpected exceptions is unstable. What motivated this?

swankjesse:
Ie. is this response to a particular NullPointerException, or is it speculative? In general I don’t think we should worry too much about unchecked exceptions _unless_ we’re calling into user code.

frapontillo:
It's not the unchecked exception in itself, users will have to handle those.
The wrong behaviour is that, when it's thrown, the body is not closed, therefore leaking memory.

I am having this issue on AWS Lambda, where Retrofit at some point complains about "A connection to https://api.something.com/ was leaked. Did you forget to close a response body?".
Since I'm not handling `ResponseBody` myself, only this could be the cause.

swankjesse:
Ahh. Are you seeing any unchecked exceptions?

frapontillo:
Not explicitly because the handlers could be swallowing it (and debugging on AWS Lambda is impossible), plus it's not that reproducible of an error.

Of course this PR is formally correct if the body is opened in `gson.newJsonReader` and not in `adapter.read`.

JakeWharton:
Hmm I can't see any way for `newJsonReader` to throw NPE unless `value.charStream()` returned `null` which should also be impossible.

frapontillo:
I'll try to debug the issue further then.

JakeWharton:
Android's new VM prints out really nice NPEs which point to the exact statement which triggers it. I wish the JVM had the equivalent. It would make debugging these a lot more simple.

JakeWharton:
Do you have a full stack trace here? (Or at least the last few stack frames from inside Retrofit?)

I've looked again and it should be impossible for this to happen.

frapontillo:
I will stress test the system in a few days and report back as soon as that happens again.

frapontillo:
I may have found out the original cause of the issue: we have a custom Interceptor that may throw when a specific header is found. In that case, we're not closing the response after reading the headers and before throwing the Exception, which leads to the leak:

```java
class RateLimitHandlerInterceptor implements Interceptor {

    private static final int HTTP_STATUS_CODE_UNAUTHORIZED = 403;

    @Override
    public Response intercept(Chain chain) throws IOException {
        Request request = chain.request();
        Response response = chain.proceed(request);

        if (response.code() == HTTP_STATUS_CODE_UNAUTHORIZED) {
            // we should close the response here
            throw new RateLimitEncounteredException("Rate limit encountered, please retry.");
        }

        return response;
    }
}
```

JakeWharton:
Yep. That will do it! Thanks for following up.

Issue #1821, opened 2016-05-27
Support protobuf 3.0 when released.


brilliance-boy:
@JakeWharton 
Hi Wharton，
Retrofit is really a delightful networking framework for android. But we have recently found “ProtoConverterFactory” is not support ProtoBuf 3.0. So we want to know when you can update the ProtoConverter.

Thank you very much.

JakeWharton:
We've been waiting for the API to stabilize since we don't support
pre-release versions.

fabriciointerama:
Yes, please. I'm waiting for it too.

@brilliance-boy 
Meanwhile, I'm using this workaround. I just copied the `ProtoConverterFactory` class code for a own class (I couldn't extend it because it is `final` and private :cry:) and I added the following:

```
Field field = c.getDeclaredField("PARSER"); // find this line
field.setAccessible(true);                  // insert this line
```

Or just get [this gist code](https://gist.github.com/fabriciointerama/22178fe28528bd2919a02f171680ca13).

echavda:
@fabriciointerama @brilliance-boy 

Hello Both, 

can you please give me some example how we can use protobuf in retrofit - i tried but its failed with some error , let me give you sample of my implementation on that.

i hope you guys will help me. 


ApiInterface.java
```
public interface ApiInterface {
    @GET
    Call<CommonProto.Country> makeGetRequest(@Url String url);
}
```

ApiClient.java
```
public class ApiClient {

    public static final String BASE_URL = "**************************";

    private static Retrofit retrofit = null;


    public static Retrofit getClient() {
        if (retrofit==null) {
            retrofit = new Retrofit.Builder()
                    .baseUrl(BASE_URL)
                    .addConverterFactory(Proto3ConverterFactory.create())
                    .build();
        }
        return retrofit;
    }

}

```

MainActivity.java
```
ApiInterface apiService =
                ApiClient.getClient().create(ApiInterface.class);

Call<CommonProto.Country> call = apiService.makeGetRequest("Services/CountryServices/GetAllCountry");

        call.enqueue(new Callback<CommonProto.Country>() {
                         @Override
                         public void onResponse(Call<CommonProto.Country> call, Response<CommonProto.Country> response) {
                             String bodyString = null;
                             try {
                                 Log.e("RETROFIT  ::::::: ", String.valueOf(response.body())+"TEST");
                             } catch (Exception e) {
                                 Log.e("RETROFIT ERROR  ::::::: ", e.getMessage()+"TEST");
[...]

PR #1879, opened 2016-06-24
Retrofit2 converter for Protobuf Nano

staticgears:
This patch adds a converter implementation for MessageNano protos, which are a lightweight implementation of Protocol Buffers.

I adapted the work from https://github.com/square/retrofit/pull/596 to the structure of the existing protobuf converter. Credit to @jnlopar for the groundwork.

As I understand it the main objection to not including protobuf-javanano support was that it was not available in maven central. This patch (and time) has addressed that and I have added the module so that it is built and tested as part of the main converter build cycle.

Cheers,
Keith


jnlopar:
Cool, thanks a ton for picking this up! Looking forward to seeing it get integrated. :)

retrofit-converters/protobuf-nano/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+To build this module, you will need to [install][2] the Android external\_protobuf module.
+
+To use the converter, depend on the `nano` classifier:
+
+```xml
+<dependency>
+  <groupId>com.squareup.retrofit</groupId>
+  <artifactId>retrofit-converters</artifactId>
+  <classifier>nano</classifier>
```
JakeWharton: Looks like this isn't accurate. There's no classifier and the artifact ID is 'converter-protobuf-nano'.

retrofit-converters/protobuf-nano/README.md
```diff
@@ -0,0 +1,19 @@
+Google Nano Protocol Buffer Converter
+=====================================
+
+A `Converter` which uses [Nano Protocol Buffer][1] binary serialization.
+
+To build this module, you will need to [install][2] the Android external\_protobuf module.
```
JakeWharton: Delete

retrofit-converters/protobuf-nano/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+    <dependency>
+      <groupId>com.squareup.retrofit2</groupId>
+      <artifactId>retrofit</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.protobuf.nano</groupId>
+      <artifactId>protobuf-javanano</artifactId>
+      <version>3.0.0-alpha-5</version>
+    </dependency>
```
JakeWharton: This should be defined in the root `pom.xml` like the other dependencies so that it can share the same version with the normal protobuf converter.

retrofit-converters/protobuf-nano/src/main/java/retrofit2/converter/protobuf/nano/ProtoConverterFactory.java
```diff
@@ -0,0 +1,85 @@
...
+import okhttp3.ResponseBody;
+import retrofit2.Converter;
+import retrofit2.Retrofit;
+
+/**
+ * A {@linkplain Converter.Factory converter} which uses Protocol Buffers Nano.
+ * <p>
+ * This converter only applies for types which extend from {@link MessageNano} (or one of its
+ * subclasses).
+ * @author Keith Banks - Architect at siilo.com
```
JakeWharton: Delete. We don't include author tags anywhere.

retrofit-converters/protobuf-nano/src/main/java/retrofit2/converter/protobuf/nano/ProtoRequestBodyConverter.java
```diff
@@ -0,0 +1,35 @@
...
+package retrofit2.converter.protobuf.nano;
+
+import com.google.protobuf.nano.MessageNano;
+import java.io.IOException;
+import okhttp3.MediaType;
+import okhttp3.RequestBody;
+import retrofit2.Converter;
+
+/**
+ * * @author Keith Banks - Architect at siilo.com
```
JakeWharton: Delete

retrofit-converters/protobuf-nano/src/main/java/retrofit2/converter/protobuf/nano/ProtoResponseBodyConverter.java
```diff
@@ -0,0 +1,44 @@
...
+package retrofit2.converter.protobuf.nano;
+
+import com.google.protobuf.nano.InvalidProtocolBufferNanoException;
+import com.google.protobuf.nano.MessageNano;
+import java.io.IOException;
+import okhttp3.ResponseBody;
+import retrofit2.Converter;
+
+/**
+ * * @author Keith Banks - Architect at siilo.com
```
JakeWharton: Delete

retrofit-converters/protobuf-nano/src/test/java/retrofit2/converter/protobuf/nano/ProtoConverterTest.java
```diff
@@ -0,0 +1,126 @@
+package retrofit2.converter.protobuf.nano;
```
JakeWharton: nit: needs license header

JakeWharton:
I'm fine with merging this once 3.0 proper is released. We'll also be doing #1821 at the same time.

retrofit-converters/protobuf-nano/src/main/java/retrofit2/converter/protobuf/nano/ProtoResponseBodyConverter.java
```diff
@@ -0,0 +1,44 @@
...
+final class ProtoResponseBodyConverter<T extends MessageNano>
+        implements Converter<ResponseBody, T> {
+  private final T instance;
+  ProtoResponseBodyConverter(T instance) {
+    this.instance = instance;
+  }
+
+  @Override public T convert(ResponseBody value) throws IOException {
+    try {
+      return MessageNano.mergeFrom(instance, value.bytes());
```
JakeWharton: Does this update `instance` or return a new copy? Because if it's the latter then a new instance needs created for every invocation of this method. We only create the converters once but call them many times.

retrofit-converters/protobuf-nano/src/main/java/retrofit2/converter/protobuf/nano/ProtoResponseBodyConverter.java
```diff
@@ -0,0 +1,59 @@
...
+    } catch (InvalidProtocolBufferNanoException e) {
+      throw new RuntimeException(e); // Despite extending IOException, this is data mismatch.
+    } finally {
+      value.close();
+    }
+  }
+
+  private T getNanoProtoInstance(final Type type) {
+    Class<T> c = (Class<T>) type;
+    if (!(MessageNano.class.isAssignableFrom(c))) {
```
JakeWharton: This was already checked in the factory. No need to check again.

retrofit-converters/protobuf-nano/src/main/java/retrofit2/converter/protobuf/nano/ProtoResponseBodyConverter.java
```diff
@@ -0,0 +1,59 @@
...
+import com.google.protobuf.nano.MessageNano;
+import java.io.IOException;
+import java.lang.reflect.Type;
+
+import okhttp3.ResponseBody;
+import retrofit2.Converter;
+
+final class ProtoResponseBodyConverter<T extends MessageNano>
+        implements Converter<ResponseBody, T> {
+  private final Type type;
```
JakeWharton: This can also be `Class<T>` which avoids the need to cast every time

retrofit-converters/protobuf-nano/src/main/java/retrofit2/converter/protobuf/nano/ProtoConverterFactory.java
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright (C) 2013 Square, Inc.
```
baoti: 2016

JakeWharton:
@staticgears once #1953 is merged can you rebase this and update it to the latest things (like it being called 'lite' now not 'nano')?

jnlopar:
So, lite and nano are actually different. Lite is basically a lighter subset of the standard Java protobuf API that is more optimized for Android (uses MessageLite), but basically API compatible, while nano is a totally different API (using MessageNano) that was specifically created for mobile use from the beginning, but given it prefers fields over methods loses some of the API niceness. Both types are still in use on various projects and they're unfortunately not quite interchangeable yet, so I think a nano change would still be valid on its own here.

PR #1881, opened 2016-06-25
headers replace just like @Path replace

inpot:
here is a simple example:
@Headers({"Range:byte={start}-{end}"})
use test(@Head("start")String start,@Head("end")String end) to replace the "{start}" and "{end}"
it's just like @Path replace

### DIFF
--- a/retrofit/src/main/java/retrofit2/ParameterHandler.java
+++ b/retrofit/src/main/java/retrofit2/ParameterHandler.java
@@ -93,6 +93,25 @@ abstract class ParameterHandler<T> {
     }
   }
 
+  static final class Head<T> extends ParameterHandler<T> {
+    private final String name;
+    private final Converter<T, String> valueConverter;
+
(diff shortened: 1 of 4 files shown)


JakeWharton:
We already offer `@Header` for supplying a runtime header value as a parameter. You could use an object like Guava's `Range` to encapsulate the two values as a single value and then annotate the parameter with `@ByteRange` to control the "rendering" to a string in a custom string converter. You could also more simply just have a `ByteRange` type whose `toString()` rendered in that format.

Thanks for your contribute, but I don't want to have two mechanisms for the same effect.

wangfuda:
@JakeWharton i've not found the annotation "@ByteRange" in the latest Retrofit 2.1.0, can u give me a guide? thans a lot.

u are my idol, i feel so happy to leave a message to u.

JakeWharton:
You would have to create it and detect it inside the string converter that
you would have to write.

PR #1882, opened 2016-06-26
Use string converter for String type.

### LINKED ISSUE #1876: retrofit2.BuildInConverters$StringConverter let me confused
I am so confused,when send request like this

 public interface API {
       @FormUrlEncoded
       @POST("login.do?os=" + ConstantValue.OS + "&version=" + ConstantValue.VERSION)
       Observable<LoginInfo> login(@Field("cmd") String params);
  } 
the method of my custom Converter stringConverter method do not work.. (to encrypt request)
retrofit2.BuildInConvert$StringConverter is first before mine execute.. help~please~
retrofit version 2.1.0 @

JakeWharton:
Unlike RequestBody and ResponseBody which are library types that must be handled by default, String is a platform type and we want to offer the flexibility of doing things like using annotations to transform the contents.

Closes #1876.

### DIFF
--- a/retrofit/src/main/java/retrofit2/BuiltInConverters.java
+++ b/retrofit/src/main/java/retrofit2/BuiltInConverters.java
@@ -27,10 +27,9 @@ final class BuiltInConverters extends Converter.Factory {
   public Converter<ResponseBody, ?> responseBodyConverter(Type type, Annotation[] annotations,
       Retrofit retrofit) {
     if (type == ResponseBody.class) {
-      if (Utils.isAnnotationPresent(annotations, Streaming.class)) {
-        return StreamingResponseBodyConverter.INSTANCE;
-      }
-      return BufferingResponseBodyConverter.INSTANCE;
+      return Utils.isAnnotationPresent(annotations, Streaming.class)
+          ? StreamingResponseBodyConverter.INSTANCE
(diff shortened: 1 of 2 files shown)


retrofit/src/main/java/retrofit2/BuiltInConverters.java
```diff
@@ -27,10 +27,9 @@
   public Converter<ResponseBody, ?> responseBodyConverter(Type type, Annotation[] annotations,
       Retrofit retrofit) {
     if (type == ResponseBody.class) {
-      if (Utils.isAnnotationPresent(annotations, Streaming.class)) {
-        return StreamingResponseBodyConverter.INSTANCE;
-      }
-      return BufferingResponseBodyConverter.INSTANCE;
+      return Utils.isAnnotationPresent(annotations, Streaming.class)
+          ? StreamingResponseBodyConverter.INSTANCE
+          : BufferingResponseBodyConverter.INSTANCE;
```
JakeWharton: This is unrelated... it just looks nicer.

eneim: Did the IDE tell you to re-format this and you just accidentally pressed yes :curious_face: ?

swankjesse: 🍩

PR #1904, opened 2016-07-08
Fix typo in Response documentation

vanniktech:
Could also be that you want `(200..300)` if so let me know and I'll adjust it.


egor-n:
`[200..300)` means 200 is included and 300 is not, [which is the desired behavior](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Response.java#L106). I think the docs were correct.

vanniktech:
Ups then sorry for the noise.

PR #1906, opened 2016-07-09
Update to OkHttp 3.4.0.

JakeWharton:


retrofit-converters/moshi/pom.xml
```diff
@@ -23,7 +23,7 @@
       <groupId>com.squareup.moshi</groupId>
       <artifactId>moshi</artifactId>
     </dependency>
-    <!-- TODO remove this dependency once Moshi or OkHttp ships with 1.9.0 or newer. -->
+    <!-- TODO remove this dependency once Moshi ships with 1.9.0 or newer. -->
```
JakeWharton: Turns out the more immediate dependency on Moshi wins with the transitive dep.

PR #1908, opened 2016-07-10
throw body while gson convert err

nightbear1009:
hi,

I think it's good to know more detail about "response" while gson convert error.
so I modify the exception. we can use getBody to get more infomation about the response


JakeWharton:
The response will have been partially consumed and should be consodered unusable at this point since the converter will have closed it.

PR #1909, opened 2016-07-11
Feature/throw headers

nightbear1009:
hi,

I think it's good to know more detail about "headers" while gson convert error.
so I modify the exception. we can use getHeaders to get more infomation such as url ...


JakeWharton:
This masks the original exception that's thrown by the converter making them all appear the same, and it would be a fairly large behavior change at this point which I'm not comfortable doing without a major release.

PR #1924, opened 2016-07-19
Update to OkHttp 3.4.1

amincheloh:


JakeWharton:
Thanks!

FYI we never make releases to track releases of our dependencies, so if you want OkHttp 3.4.1 make sure you're declaring it explicitly in your build.

amincheloh:
Ok, thanks.

PR #1931, opened 2016-07-21
Feature/add query list

wonhyeongju:
Add - @QueryList
https://github.com/square/retrofit/pull/1184

Multiple Query Parameters of Same Name.
Like ?user_id=aaa&user_id=bbb&user_id=ccc

example:
```@GET("v1/user/list")
    Call<List<User>> list(@Query("api_key") String apiKey, @QueryList List<KeyValue<String, String>> options);```

### DIFF
--- a/retrofit/src/main/java/retrofit2/KeyValue.java
+++ b/retrofit/src/main/java/retrofit2/KeyValue.java
@@ -2,25 +2,25 @@ package retrofit2;
 
 public class KeyValue<K, V> {
 
-    public K key;
+  public K key;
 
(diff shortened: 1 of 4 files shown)


JakeWharton:
This is already supported: https://github.com/square/retrofit/blob/fccedbeb4d5181c926fff450cdf5c5116ef0eeaa/retrofit/src/test/java/retrofit2/RequestBuilderTest.java#L1054-L1068

wonhyeongju:
it's different. QueryList is for options.

PR #1934, opened 2016-07-22
Feature/add query list. 

wonhyeongju:
It is not same list with @Query("key").
QueryList is for options parameter.


JakeWharton:
How is it different from `@QueryMap`?

wonhyeongju:
Querymap not allow same query key.

JakeWharton:
I would rather support iterable values in maps and we also are tracking custom parameter type handling with #626.

PR #1943, opened 2016-07-25
NDJSON support converter

CotunaAurelian:
Adds converter for the NDJSON format.

There is currently no standard for transporting instances of JSON text within a stream protocol, apart from `Websockets`, which is unnecessarily complex for non-browser applications.

A common use case for NDJSON is delivering multiple instances of JSON text through streaming protocols like TCP or UNIX Pipes. It can also be used to store semi-structured data.

NDJSON example
     {"some":"thing"}
     {"foo":17,"bar":false,"quux":true}
     {"may":{"include":"nested","objects":["and","arrays"]}}

### DIFF
--- a/retrofit-converters/ndjson/README.md
+++ b/retrofit-converters/ndjson/README.md
@@ -1,0 +1,17 @@
+NDJSON Converter
+==============
+
+A `Converter` for the [NDJSON][1] format serialization.
+There is currently no standard for transporting instances of JSON text within a stream protocol,
+apart from `Websockets`, which is unnecessarily complex for non-browser applications.
+A common use case for `NDJSON` is delivering multiple instances of JSON text through streaming
+protocols like TCP or UNIX Pipes. It can also be used to store semi-structured data.
+
+An instance of `NDJsonConverter` will be created and it will use a `ResponseBody` to create
+a `NDJsonResponse` object.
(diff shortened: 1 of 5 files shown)


retrofit-converters/ndjson/src/BaseResponse.java
```diff
@@ -0,0 +1,94 @@
...
+import okhttp3.ResponseBody;
+import okio.BufferedSource;
+
+/**
+ * Base class for all responses received from the server
+ * <p>
+ * Created by Cotuna Aurelian on 7/2/2016.
+ */
+
+public class BaseResponse extends ResponseBody implements Serializable{
```
FabianTerhorst: add a space between Serializable and {

CotunaAurelian: Added. Thank you

swankjesse:
Retrofit doesn’t intend to keep a full inventory of converters. I think you should distribute this as a standalone GitHub project. Otherwise it puts some burden on us to keep this working, and I don’t have any understanding in NDJSON.

https://github.com/ligboy/retrofit-converter-fastjson
https://gist.github.com/aurae/8427b93b27483763d9cb
https://gist.github.com/ararog/fd1405124d2d64bc9a38

Issue #1846, opened 2016-06-08
Order of attributes in POST request


josephlam1984:
The Amazon S3 REST API requires the attributes in a POST request to be submitted in a specific order (i.e. the key field needs to go before the file field).

I understand that the official JSON specification suggests that the order of attributes should not matter in a transmission. However, given the above requirement, can retrofit support a manual, explicit ordering of attributes when putting together a POST request?

Thanks!

JakeWharton:
Ordering is based on what you provide.

josephlam1984:
Thanks Jake. In my code below, the file field is received first, then the key field:
```

RequestBody requestFile = RequestBody.create(MediaType.parse("image/*"), file);
RequestBody requestBody = RequestBody.create(MediaType.parse("application/json"), gson.toJson(media.getUploadInfo().getFields()));

Call<AWSS3Asset> s3Call = mAWSS3WebService.upload(
	media.getUploadInfo().getURL(),
        requestBody,
        requestFile
);
```

The interface looks like this:

```
    @Multipart
    @POST
    Call<AWSS3Asset> upload(@Url String url,
                            @Part("data") RequestBody data,
                            @Part("file") RequestBody file);
```

JakeWharton:
I'll write up a test case soon

matbos:
@JakeWharton
I have created a PR(#1949) with tests that ensure `@Part`s are being translated into the `Request` in proper order.

jaredsburrows:
@josephlam1984 @JakeWharton We can close this now that @matbos's PR is in.

PR #1949, opened 2016-07-28
Add tests for order of multipart request's parts. Issue #1846.

matbos:


retrofit/src/test/java/retrofit2/RequestBuilderTest.java
```diff
@@ -2486,6 +2486,45 @@
...
+
+    Buffer buffer = new Buffer();
+    body.writeTo(buffer);
+    String readBody = buffer.readUtf8();
+
+    assertThat(readBody.indexOf("firstParam")).isLessThan(readBody.indexOf("secondParam"));
+    assertThat(readBody.indexOf("secondParam")).isLessThan(readBody.indexOf("thirdParam"));
+  }
+
+  @Test public void multipartPartsShouldBeInOrderWithUrlParam() throws IOException {
```
JakeWharton: Why this extra variant of the test?

matbos: I only have added this second version to reflect a code sample provided in the #1846.
Come to think of it twice, it is not needed, will remove.

JakeWharton:
Thanks!

PR #1953, opened 2016-08-03
Update to support protobuf 3.0.

### LINKED ISSUE #1821: Support protobuf 3.0 when released.

JakeWharton:
Closes #1821.

### DIFF
--- a/retrofit-converters/protobuf/src/main/java/retrofit2/converter/protobuf/ProtoConverterFactory.java
+++ b/retrofit-converters/protobuf/src/main/java/retrofit2/converter/protobuf/ProtoConverterFactory.java
@@ -20,6 +20,8 @@ import com.google.protobuf.MessageLite;
 import com.google.protobuf.Parser;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.lang.reflect.Type;
 import okhttp3.RequestBody;
 import okhttp3.ResponseBody;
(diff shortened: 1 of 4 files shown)


retrofit-converters/protobuf/src/main/java/retrofit2/converter/protobuf/ProtoConverterFactory.java
```diff
@@ -61,12 +63,20 @@ private ProtoConverterFactory(ExtensionRegistryLite registry) {
...
     try {
-      Field field = c.getDeclaredField("PARSER");
+      Method method = c.getDeclaredMethod("parser");
       //noinspection unchecked
-      parser = (Parser<MessageLite>) field.get(null);
-    } catch (NoSuchFieldException | IllegalAccessException e) {
-      throw new IllegalArgumentException(
-          "Found a protobuf message but " + c.getName() + " had no PARSER field.");
+      parser = (Parser<MessageLite>) method.invoke(null);
+    } catch (NoSuchMethodException | InvocationTargetException | IllegalAccessException ignored) {
```
swankjesse: pedantic: if it’s an InvocationTargetException, falling back doesn’t seem right

swankjesse:
LGTM! Nice trick supporting two versions simultaneously.

JiongBull:
Thanks, i'm waiting for it, when it can be used for proto3.

JakeWharton:
There's nothing preventing the current converter from being used with
proto3.

JiongBull:
@JakeWharton  sincere thanks

retrofit-converters/protobuf/src/main/java/retrofit2/converter/protobuf/ProtoConverterFactory.java
JakeWharton: Rethrow the cause? It should never happen.

fabriciointerama:
@JakeWharton  Hi. I'm new with java and android dev. That said, sorry if I say some silly stuff here. I was trying your updated factory class but compiler says this error for your [multi-catch](https://github.com/square/retrofit/blob/f7a8ba0b448a8643bdb8e989f19468d6c365e8ee/retrofit-converters/protobuf/src/main/java/retrofit2/converter/protobuf/ProtoConverterFactory.java#L71): 

> **Multi-catch** with these reflection exceptions requires API level 19 (current min is 16) because they get compiled to the common but new super type ReflectiveOperationException. As a workaround either create individual catch statements, or catch Exception.

And since [Retrofit says that](https://github.com/square/retrofit):

> Retrofit requires at minimum Java 7 or **Android 2.3**.

I don't know if this is an issue, so I thought it would be better to note it.

JakeWharton:
Thanks I'll have a look.

JakeWharton:
This was determined to not be an issue in the library. We actually have a compile-time check in place which prevents this from happening.

PR #1969, opened 2016-08-11
Defer retrieval of Gson's TypeAdapters until convert() is called

ronocod:
Prevents Gson adapters from being created on the main thread when a client method is called from the main thread.

### DIFF
--- a/retrofit-converters/gson/src/main/java/retrofit2/converter/gson/GsonAdapterProvider.java
+++ b/retrofit-converters/gson/src/main/java/retrofit2/converter/gson/GsonAdapterProvider.java
@@ -1,0 +1,8 @@
+package retrofit2.converter.gson;
+
+import com.google.gson.TypeAdapter;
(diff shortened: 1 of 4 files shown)


JakeWharton:
This defeats the whole purpose of the factory/converter system which is to build the required state once and re-use it for all subsequent calls. Additionally it only changes one converter and leaves all the others with the existing behavior.

Why do you want to move adapter lookup off the main thread?

artem-zinnatullin:
Alternatively, you can shift creation of the Retrofit instance to another
thread, for example to the thread that performs network request. If you use
Dagger you can make Retrofit Lazy and access it on the worker thread to
achieve that effect.

ronocod:
I've only just seen the `validateEagerly` option on the `Retrofit.Builder` class 🙈. Using that and creating the Retrofit instance on a background thread will solve our issues. Thanks for your help guys!

PR #1977, opened 2016-08-16
Refactor Rx tests to avoid the use of blocking operators.

JakeWharton:


retrofit-adapters/rxjava/src/test/java/retrofit2/adapter/rxjava/ObservableTest.java
```diff
@@ -103,31 +97,32 @@
...
     server.enqueue(new MockResponse().setBody("Hi"));
 
-    BlockingObservable<Response<String>> o = service.response().toBlocking();
-    Response<String> response = o.first();
+    TestSubscriber<Response<String>> subscriber = new TestSubscriber<>();
+    service.response().subscribe(subscriber);
+    Response<String> response = subscriber.getOnNextEvents().get(0);
     assertThat(response.isSuccessful()).isTrue();
     assertThat(response.body()).isEqualTo("Hi");
+    subscriber.assertCompleted();
```
artem-zinnatullin: Maybe add `assertNoErrors`?

retrofit-adapters/rxjava/src/test/java/retrofit2/adapter/rxjava/ObservableTest.java
```diff
@@ -103,31 +97,32 @@
...
     server.enqueue(new MockResponse().setResponseCode(404).setBody("Hi"));
 
-    BlockingObservable<Response<String>> o = service.response().toBlocking();
-    Response<String> response = o.first();
+    TestSubscriber<Response<String>> subscriber = new TestSubscriber<>();
+    service.response().subscribe(subscriber);
+    Response<String> response = subscriber.getOnNextEvents().get(0);
     assertThat(response.isSuccessful()).isFalse();
     assertThat(response.errorBody().string()).isEqualTo("Hi");
+    subscriber.assertCompleted();
```
artem-zinnatullin: And here

artem-zinnatullin:
:+1: (I would add more assertions of the state since TestSubscriber swallows errors and to check that Retrofit does not violate the Observable contract)

retrofit-adapters/rxjava/src/test/java/retrofit2/adapter/rxjava/ObservableTest.java
artem-zinnatullin: And in similar places

retrofit-adapters/rxjava/src/test/java/retrofit2/adapter/rxjava/ObservableTest.java
JakeWharton: I don't see why. `SafeSubscriber` dictates the `TestSubscriber` will only ever see a completed event or an error event, so the lack of a completed event will already fail the test.

artem-zinnatullin: Well, it's implementation detail, but ok

JakeWharton: SafeSubscriber is an implementation detail, but that is the observable
contract which is not. If we want to test whether the contract is honored
then the tests need ported to use unsafeSubscribe and there's a lot more
cases to cover.

retrofit-adapters/rxjava/src/test/java/retrofit2/adapter/rxjava/ObservableTest.java
```diff
@@ -103,31 +97,32 @@
...
-    try {
-      o.first();
-      fail();
-    } catch (RuntimeException t) {
-      assertThat(t.getCause()).isInstanceOf(IOException.class);
-    }
+    TestSubscriber<Response<String>> subscriber = new TestSubscriber<>();
+    service.response().subscribe(subscriber);
+    subscriber.assertNoValues();
+    subscriber.assertError(IOException.class);
```
swankjesse: I like that there’s no boxed exception here.

swankjesse:
LGTM!

PR #1978, opened 2016-08-17
Simplify observable creation and reduce runtime allocations.

JakeWharton:


JakeWharton:
License headers are different because these new files are copied from a personal project https://github.com/JakeWharton/retrofit2-rxjava2-adapter

retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/BodyOnSubscribe.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override public void onError(Throwable throwable) {
+      if (!terminated) {
+        subscriber.onError(throwable);
+      }
+    }
+
+    @Override public void onCompleted() {
+      if (!terminated) {
+        subscriber.onCompleted();
```
swankjesse: terminated = true ?

swankjesse: ahh, I think `terminated` means that it was by error.

retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/CallOnSubscribe.java
```diff
@@ -0,0 +1,81 @@
...
+
+  @Override public void call(Subscriber<? super Response<T>> subscriber) {
+    // Since Call is a one-shot type, clone it for each new subscriber.
+    Call<T> call = originalCall.clone();
+    CallArbiter<T> arbiter = new CallArbiter<>(call, subscriber);
+    subscriber.add(arbiter);
+    subscriber.setProducer(arbiter);
+  }
+
+  private static final class CallArbiter<T> extends AtomicBoolean
```
swankjesse: extending AtomicBoolean is weird. Why is that even allowed?

retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/BodyOnSubscribe.java
JakeWharton: I'm cheating here because `onCompleted` is a terminal event. This means we know that no other events will come after it so there's no need to update the state. I can comment to that effect.

retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/CallOnSubscribe.java
```diff
@@ -0,0 +1,81 @@
...
+      if (amount == 0 || !compareAndSet(false, true)) {
+        return;
+      }
+
+      try {
+        Response<T> response = call.execute();
+        if (!call.isCanceled()) {
+          subscriber.onNext(response);
+        }
+        if (!call.isCanceled()) {
```
swankjesse: subtle

swankjesse: Looks right . . . probably worth a test though.

retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/CallOnSubscribe.java
JakeWharton: It definitely shouldn't be allowed. But it has become fairly common in Rx code to save bytes in the runtime allocation and indirection when calling its methods.

retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/CallOnSubscribe.java
JakeWharton: There's a bunch of correctness tests missing (and some missing behavior around correctness as well). Didn't want to refactor and change behavior at the same time since this code is hard enough to read. I have at least one follow-up coming around this.

retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/BodyOnSubscribe.java
JakeWharton: Also there should be an `else` branch that sends the error to a handler, but I didn't want to change behavior while refactoring.

retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/CallOnSubscribe.java
```diff
@@ -0,0 +1,81 @@
...
+    subscriber.add(arbiter);
+    subscriber.setProducer(arbiter);
+  }
+
+  private static final class CallArbiter<T> extends AtomicBoolean
+      implements Subscription, Producer {
+    private final Call<T> call;
+    private final Subscriber<? super Response<T>> subscriber;
+
+    CallArbiter(Call<T> call, Subscriber<? super Response<T>> subscriber) {
```
swankjesse: Curious: is _Arbiter_ a standard concept in Rx? Don’t love the term though I can’t come up with anything that’s much better.

JakeWharton: It's an internal concept in RxJava's implementation, but not a public API concept. Yet again, it's just smashing together responsibilities into a single object for the sake of saving bytes and reducing indirections.

In RxJava 2 all this stuff is cleaned up a ton. Compare this class to https://github.com/JakeWharton/retrofit2-rxjava2-adapter/blob/master/src/main/java/com/jakewharton/retrofit2/adapter/rxjava2/CallObservable.java

retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/RxJavaCallAdapter.java
```diff
@@ -0,0 +1,87 @@
...
+    if (isSingle) {
+      return SingleHelper.toSingle(observable);
+    }
+    if (isCompletable) {
+      return CompletableHelper.toCompletable(observable);
+    }
+    return observable;
+  }
+
+  /** Separate static class defers classloading and bytecode verification. */
```
swankjesse: I’m curious what the motivation of this is. Seems like we’re trading one method call for another.

swankjesse:
LGTM

retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/CallOnSubscribe.java
swankjesse: SGTM!

retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/CallOnSubscribe.java
swankjesse: Works for me. Fortunately this doesn’t leak into our API.

retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/RxJavaCallAdapter.java
JakeWharton: Single and Completable (and thus these methods) are experimental APIs in the current version. This means they're subject to backwards-incompatible changes. The goal here (and in the factory which compares classes by name) is to make this code resilient against these APIs breaking. Ideally both `Single` and `Completable` types and these methods could be removed from the library and this call adapter would continue to work.

Unfortunately that's a bit hard to test, but I have some ideas–just not the time.

retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/BodyOnSubscribe.java
swankjesse: Works for me.

retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/RxJavaCallAdapter.java
JakeWharton: If they ever become stable (which usually takes two minor versions), we can update the dependency to that RxJava version and remove this indirection and the name-based matching in the factory.

swankjesse: Ahh, got it. The comment didn’t cause me to realize that “deferring verification” meant “avoiding verification” !

PR #1980, opened 2016-08-18
Switch to a better, custom validating subscriber.

JakeWharton:
TestSubscriber is a poor implementation for correctness testing. This new subscriber ensures all events are tested by being a JUnit rule. It also consumes events as they're validated so that assertion becomes more like a script than a ledger.

This also switches tests to use unsafeSubscribe to ensure that any contract violations are not being suppressed. The latest version of RxJava is required for accessing this method on Completable.

### DIFF
--- a/retrofit-adapters/rxjava/src/test/java/retrofit2/adapter/rxjava/ObservableTest.java
+++ b/retrofit-adapters/rxjava/src/test/java/retrofit2/adapter/rxjava/ObservableTest.java
@@ -25,13 +25,13 @@ import retrofit2.Response;
 import retrofit2.Retrofit;
 import retrofit2.http.GET;
 import rx.Observable;
-import rx.observers.TestSubscriber;
 
 import static okhttp3.mockwebserver.SocketPolicy.DISCONNECT_AFTER_REQUEST;
 import static org.assertj.core.api.Assertions.assertThat;
 
 public final class ObservableTest {
   @Rule public final MockWebServer server = new MockWebServer();
+  @Rule public final RecordingSubscriber.Rule subscriberRule = new RecordingSubscriber.Rule();
 
   interface Service {
     @GET("/") Observable<String> body();
@@ -53,42 +53,37 @@ public final class ObservableTest {
   @Test public void bodySuccess200() {
     server.enqueue(new MockResponse().setBody("Hi"));
 
-    TestSubscriber<String> subscriber = new TestSubscriber<>();
-    service.body().subscribe(subscriber);
-    subscriber.assertValues("Hi");
-    subscriber.assertCompleted();
+    RecordingSubscriber<String> subscriber = subscriberRule.create();
+    service.body().unsafeSubscribe(subscriber);
+    subscriber.assertValue("Hi").assertCompleted();
   }
 
   @Test public void bodySuccess404() {
     server.enqueue(new MockResponse().setResponseCode(404));
 
-    TestSubscriber<String> subscriber = new TestSubscriber<>();
-    service.body().subscribe(subscriber);
-    subscriber.assertNoValues();
-    Throwable cause = subscriber.getOnErrorEvents().get(0);
-    assertThat(cause).isInstanceOf(HttpException.class).hasMessage("HTTP 404 Client Error");
+    RecordingSubscriber<String> subscriber = subscriberRule.create();
+    service.body().unsafeSubscribe(subscriber);
+    subscriber.assertError(HttpException.class, "HTTP 404 Client Error");
   }
 
   @Test public void bodyFailure() {
     server.enqueue(new MockResponse().setSocketPolicy(DISCONNECT_AFTER_REQUEST));
 
-    TestSubscriber<String> subscriber = new TestSubscriber<>();
-    service.body().subscribe(subscriber);
-    subscriber.assertNoValues();
+    RecordingSubscriber<String> subscriber = subscriberRule.create();
+    service.body().unsafeSubscribe(subscriber);
     subscriber.assertError(IOException.class);
   }
 
   @Test public void bodyRespectsBackpressure() {
     server.enqueue(new MockResponse().setBody("Hi"));
 
-    TestSubscriber<String> subscriber = new TestSubscriber<>(0);
-    Observable<String> o = service.body();
-
-    o.subscribe(subscriber);
+    RecordingSubscriber<String> subscriber = subscriberRule.createWithInitialRequest(0);
+    service.body().unsafeSubscribe(subscriber);
     assertThat(server.getRequestCount()).isEqualTo(0);
 
     subscriber.requestMore(1);
     assertThat(server.getRequestCount()).isEqualTo(1);
+    subscriber.assertAnyValue().assertCompleted();
 
     subscriber.requestMore(Long.MAX_VALUE); // Subsequent requests do not trigger HTTP requests.
     assertThat(server.getRequestCount()).isEqualTo(1);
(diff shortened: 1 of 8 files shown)


retrofit-adapters/rxjava/src/test/java/retrofit2/adapter/rxjava/RecordingSubscriber.java
```diff
@@ -0,0 +1,146 @@
...
+import org.junit.runners.model.Statement;
+import rx.Notification;
+import rx.Subscriber;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/** A test {@link Subscriber} and JUnit which guarantees all events are asserted. */
+final class RecordingSubscriber<T> extends Subscriber<T> {
+  private final long initialRequest;
+  private final Deque<Notification<T>> events = new ArrayDeque<>();
```
artem-zinnatullin: Maybe also store accumulative list of received notifications for debug/print purposes?

retrofit-adapters/rxjava/src/test/java/retrofit2/adapter/rxjava/RecordingSubscriber.java
```diff
@@ -0,0 +1,146 @@
...
+    if (notification == null) {
+      throw new AssertionError("No event found!");
+    }
+    return notification;
+  }
+
+  public T takeValue() {
+    Notification<T> notification = takeNotification();
+    assertThat(notification.isOnNext())
+        .overridingErrorMessage("Expected onNext event but was %s", notification)
```
artem-zinnatullin: Printing whole list of received notifications will be helpful I guess, same for all other exception messages below

JakeWharton: The assertions that were successful prior to this one act as a list. Why
repeat them?

artem-zinnatullin: To easily see actual state of subscriber without diving into details of test case code, like:

```
Expected onNext event but was onError.

State: onNext("a"), onNext("b"), onError("IllegalStateException(whoa)")
```

JakeWharton: This isn't a reusable utility so I don't feel compelled to bother. The
error message already makes it really easy to understand what went wrong
since you can only assert one thing at a time and the stacktrace indicates
which assertion failed.

artem-zinnatullin: ok, I thought that you will extract it to a library as usual 😸 

// RxBinding has pretty similar `RecordingObserver`

JakeWharton: I have a non-public library of it but it has a very different
implementation and it does have much more detailed messaging. I'll consider
this suggestion for it.

artem-zinnatullin:
👍 

But sad that `RecordingSubscriber` itself doesn't have tests which in worst possible scenario may lead to false-positive tests of Retrofit RxJavaAdapter -> affect users.

JakeWharton:
It won't.

retrofit-adapters/rxjava/src/test/java/retrofit2/adapter/rxjava/RecordingSubscriber.java
```diff
@@ -0,0 +1,146 @@
...
+import java.util.List;
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+import rx.Notification;
+import rx.Subscriber;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/** A test {@link Subscriber} and JUnit which guarantees all events are asserted. */
```
swankjesse: and JUnit _rule_ ?

swankjesse: Neat idea

retrofit-adapters/rxjava/src/test/java/retrofit2/adapter/rxjava/RecordingSubscriber.java
```diff
@@ -0,0 +1,146 @@
...
+  public void assertCompleted() {
+    Notification<T> notification = takeNotification();
+    assertThat(notification.isOnCompleted())
+        .overridingErrorMessage("Expected onCompleted event but was %s", notification)
+        .isTrue();
+    assertNoEvents();
+  }
+
+  public void assertError(Throwable throwable) {
+    assertThat(takeError()).isSameAs(throwable);
```
swankjesse: isEqualTo() ?

swankjesse: (same result in practice)

retrofit-adapters/rxjava/src/test/java/retrofit2/adapter/rxjava/RecordingSubscriber.java
```diff
@@ -0,0 +1,146 @@
...
+      subscribers.add(subscriber);
+      return subscriber;
+    }
+
+    @Override public Statement apply(final Statement base, Description description) {
+      return new Statement() {
+        @Override public void evaluate() throws Throwable {
+          base.evaluate();
+          for (RecordingSubscriber<?> subscriber : subscribers) {
+            subscriber.assertNoEvents();
```
swankjesse: assertNoMoreEvents() ?

retrofit-adapters/rxjava/src/test/java/retrofit2/adapter/rxjava/RecordingSubscriber.java
```diff
@@ -0,0 +1,146 @@
...
+
+  public void assertNoEvents() {
+    assertThat(events).as("Unconsumed events found!").isEmpty();
+  }
+
+  public void requestMore(long amount) {
+    request(amount);
+  }
+
+  public static final class Rule implements TestRule {
```
swankjesse: Should the top-level class implement Rule directly?

swankjesse: Ahhh . . . this is the factory of subscribers. That works for me.

retrofit-adapters/rxjava/src/test/java/retrofit2/adapter/rxjava/ObservableTest.java
```diff
@@ -144,51 +133,40 @@
...
   }
 
   @Test public void resultRespectsBackpressure() {
     server.enqueue(new MockResponse().setBody("Hi"));
 
-    TestSubscriber<Result<String>> subscriber = new TestSubscriber<>(0);
-    Observable<Result<String>> o = service.result();
-
-    o.subscribe(subscriber);
+    RecordingSubscriber<Result<String>> subscriber = subscriberRule.get(0);
```
swankjesse: This line doesn’t read very well . . . kind of looks like getting the first element from a list.

Would it be simpler to say something like this:
```
RecordingSubscriber<Result<String>> subscriber = subscriberRule.get().initialRequest(0);
```

swankjesse:
LGTM!

retrofit-adapters/rxjava/src/test/java/retrofit2/adapter/rxjava/ObservableTest.java
swankjesse: (Might be overthinking this for a test API)

JakeWharton: Yeah. This method should be called `create` too. `get` is super weird. I'll do `createWithInitialRequest` because you can only set the initial request once and I don't want to deal with additional state management to check when you call `initialRequest`.

retrofit-adapters/rxjava/src/test/java/retrofit2/adapter/rxjava/RecordingSubscriber.java
JakeWharton: This was the original name. I renamed it to assertNoEvents because it's a public method that's used in tests and it sounded weird when not used as the final assertion, but in the middle of the test.

PR #1984, opened 2016-08-21
Add toString implementations to Response and Result.

JakeWharton:


swankjesse:
(I'm wondering if the raw response toString is decent)

JakeWharton:
It is!

PR #1996, opened 2016-08-28
add baseHttpsUrl for https request.

ly85206559:
I found a lot of projects have the same baseUrl,but some requests are http and some requests are https.
For example: baseUrl is "github.com/",if we want sometimes baseUrl is like "http://github.com/" and sometimes baseUrl is like "https://github.com/" for https request, we have to create 2 retrofit instance.


swankjesse:
I don’t think this extra behavior is worth the complexity. Could you change your code to have two interfaces, one for HTTP endpoints and another for HTTPS endpoints? Even better, consider migrating everything to HTTPS.

ly85206559:
thank you.

PR #2003, opened 2016-08-31
Pluggable parameters handlers

iwo:
This is an implementation of https://github.com/square/retrofit/issues/626.

It would be great if you could review my comments in this issue.

This PR makes `RequestBuilder` and couple methods in `Utils` public. Before releasing this changes they would probably need to be reviewed and documented.


swankjesse:
I don’t think this is a good fit for Retrofit. We’re trying to offer features in a small, easy-to-learn API. This makes that API much larger for a use case that doesn’t strictly need to be a part of Retrofit.

iwo:
This change is making the public API a bit bigger but as @JakeWharton said in #626 it is an advanced extension point so I don't think this is such a big deal.

The ParameterHandler API is rather small (just like Converter or CallAdapter). If you are concerned about exposing the implementation of built in ParameterHandlers they could be hidden.

If you decide that exposing such API is too much (which I can understand) #626 should probably be closed.

I would appreciate if you could have a look at how I reorganized the code anyway. It significantly reduces code duplication and allows to add new features, like multimap support, by modifying code in one place instead of 4-6. Let me know if you'd be interested in a PR with code cleanup without exposing a public API.

matemink:
Hi. How to use this ParameterHandlers? I want to use something like this
 @POST(Constants.Url)
  Single<Response<Void>> getByApi(@Body CustomObject params);

I took your code and built *.jar file and included it in my project

taojoe:
I think this idea fit this situation:

when the server use json request,  the client api still be simple.

`
@POST(...)
  Call<SomeThing> work(@JsonProp("username") String username, @JsonProp("password"));
`

Issue #1960, opened 2016-08-05
Query with dynamic key and value as list


kabindra:
I want to create dynamic query for elasticsearch facet search thus require query in form
?q=t-shirt&size[]=S&size[]=L&category=men&color[]=Black&color[]=Grey

I couldn't use List<String> for Query values as query names are dynamic also I couldn't place multiple entries in Map as key has to be unique. Thus `@QueryMap(Map<String,String> map)` or `@Query(List<String> list)` or `@QueryMap(Map<String, List<String>> map)` just don't fit in.

JakeWharton:
This has been requested multiple times and we've bucketed it under #626. That said, I think the current plan is to implement this natively for 2.2 so I'll just leave this open to track it.

iwo:
If `@QueryMap Map<String, List<String>> map` created a separate query argument for each value in the list would that solve your problem @kabindra? If so, we could check the second parameter of `Map` type and if it's `Iterable` enable the repeating behavior just like we do for `@Query List<String> list`.

@JakeWharton do you think we can enable this behavior for all `@*Map` annotations by default or would you add some additional flag to the annotations that use request body converters?
Someone may have a custom `Converter` registered for the `List<Foo>` type in which case enabling this by default would be a breaking change.

kabindra:
@iwo If we can put Iterable list as key-value pairs instead of key and values as a single pair, I think it will solve it.

JakeWharton:
Tracking with #1324.

Issue #1324, opened 2015-11-25
@QueryMap not supporting multi values


pawel-kaminski-krk:
hi,
we are using retrofit 1.9 and we declared our endpoint as
```
@GET("/path")
Response get(@QueryMap Map<String, String> queryParams);
```
which works fine and is expanded to ex. ```host/path?rq=1&ra=2``` for input ```{"rq": "1", "ra": "2"}```. 

right now we need to support multivalue same as described here https://futurestud.io/blog/retrofit-multiple-query-parameters-of-same-name but for @QueryMap. and declaring endpoint as
```
@GET("/path")
Response get(@QueryMap Map<String, List<String>> queryParams);
```
doesn't help at all as value is expanded to its string equivalent ex. ```host/path?rq=[1,2]``` for input ```{"rq": ["1", "2"]}```. we would like to get same behavior as for @Query ex. ```host/path?rq=1&rq=2``` for input ```{"rq": ["1", "2"]}```.

is there any work around for now?

JakeWharton:
None of the maps support this currently, no.

pawel-kaminski-krk:
is there any road map. we solve the issue adding another parameter with @Query annotation which supports multi value. but this is ugly

liorzam:
In my project I serialized my object via GSON to string and with @Query annotation I succeed to send it probably

pawel-kaminski-krk:
it is not about serializing but properly handling multivalues for one key

raniejade:
`@PartMap ` does not support multi values as well. One trick around it is to extend `Map` and overriding `Map#entrySet()`, though this can break as it depends on the internals.

ming13:
Most likely #1184 provides some help with this.

pawel-kaminski-krk:
thanks! it seems will do the trick!

mandybess:
in the meantime, i hacked this together for anyone who absolutely needs this and can't wait - https://gist.github.com/mandybess/dca2e8a0527aff2d8e0688c17297c945

jm-lim:
I have a simple trick. 

```
public class ProxyRetrofitQueryMap extends HashMap<String, Object> {
    public ProxyRetrofitQueryMap(Map<String, Object> m) {
        super(m);
    }

    @Override
    public Set<Entry<String, Object>> entrySet() {
        Set<Entry<String, Object>> originSet = super.entrySet();
        Set<Entry<String, Object>> newSet = new HashSet<>();

[...]

PR #2005, opened 2016-09-01
Add supprot for multi-values Maps

iwo:
Implements support for multimaps requested in #1960, #1324 and numerous closed issues referenced from #626.

Multiple values are generated for map declared as `Map<String, Iterable<T>>` and `Map<String, T[]>`.

To maintain backward compatibility, this feature is disabled by default and needs to be enabled by setting `multimap` annotation argument to `true` (e.g. `@QueryMap(multimap = true) Map<String, List<String>> queries`).


swankjesse:
Sorry for dropping the ball on this. I think this was just a big scary PR and we didn’t know how to approach it, and now it’s been stalled for a long time.

PR #2006, opened 2016-09-01
Replaced invocation of deprecated factory methods in JacksonConverterFactory

ArazAbishov:
Replaced invocation of deprecated factory methods in favor of newer counterparts in JacksonConverterFactory.


JakeWharton:
Thanks!

PR #2016, opened 2016-09-17
Update RxJava adapter to 1.2.0.

JakeWharton:
Use a more idiomatic creation mechanism which eliminates a lot of responsibility for our implementation.


retrofit-adapters/rxjava/src/test/java/retrofit2/adapter/rxjava/ObservableTest.java
```diff
@@ -86,83 +81,16 @@
 
     RecordingSubscriber<String> subscriber = subscriberRule.createWithInitialRequest(0);
     service.body().unsafeSubscribe(subscriber);
-    assertThat(server.getRequestCount()).isEqualTo(0);
+    assertThat(server.getRequestCount()).isEqualTo(1);
+    subscriber.assertNoEvents();
```
JakeWharton: This is an unrelated internal behavior change I've been meaning to make. It has no effect on what the downstream consumer of the stream sees in terms of behavior.

retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/RxJavaCallAdapter.java
```diff
@@ -46,24 +44,23 @@
   }
 
   @Override public Object adapt(Call<R> call) {
-    OnSubscribe<Response<R>> callFunc = new CallOnSubscribe<>(call);
+    ResponseCallable<R> resultCallable = new ResponseCallable<>(call);
```
swankjesse: I like this new callable-driven approach

PR #2017, opened 2016-09-17
Promote adapter response body type parameter to class level.

### LINKED ISSUE #1985: CallAdapter's generic should be defined on the type, not method.
After all, the whole point of this class is that it adapts a single type. I _think_ we can do this change in a binary-compatible way but breaking source compatibility.

JakeWharton:
This is a source-incompatible but binary-compatible change. Since we don't expect a lot of third-party adapters, I think this is acceptable.

Closes #1985.

### DIFF
--- a/retrofit/src/main/java/retrofit2/Retrofit.java
+++ b/retrofit/src/main/java/retrofit2/Retrofit.java
@@ -21,9 +21,9 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executor;
 import okhttp3.HttpUrl;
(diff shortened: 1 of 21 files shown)


retrofit/src/main/java/retrofit2/Retrofit.java
```diff
@@ -158,12 +159,14 @@ private void eagerlyValidateMethods(Class<?> service) {
     }
   }
 
-  ServiceMethod loadServiceMethod(Method method) {
-    ServiceMethod result;
+  ServiceMethod<?, ?> loadServiceMethod(Method method) {
+    ServiceMethod<?, ?> result = serviceMethodCache.get(method);
+    if (result != null) return result;
+
     synchronized (serviceMethodCache) {
```
swankjesse: nit: weird to mix different sync mechanisms here. Not sure how to structure it better

JakeWharton: Oops that wasn't supposed to be in this PR but a separate one. Hard to do anything better since we don't have Java 8's `computeIfAbsent` to override.

PR #2023, opened 2016-09-20
Close ResponseBody on 204/205 to avoid connection leak

joaham:
I was getting "A connection to https://... was leaked. Did you forget to close a response body?" on all 204 responses.

### DIFF
--- a/retrofit/src/main/java/retrofit2/OkHttpCall.java
+++ b/retrofit/src/main/java/retrofit2/OkHttpCall.java
@@ -203,6 +203,7 @@ final class OkHttpCall<T> implements Call<T> {
     }
 
     if (code == 204 || code == 205) {
+      rawBody.close();
       return Response.success(null, rawResponse);
     }
 


JakeWharton:
Good catch! I'm going to look at writing a test for this before merging because I think we already have one for non-200 responses that should be easy to port.

JakeWharton:
There's not a good test setup right now for this. I have an idea of how to do it but I'll add it later.

adrien-aubel:
Thanks @joaham! Happy to see this commit, I've been having this issue as well.

@JakeWharton is a release including this planned anytime soon? Would be awesome as this is impacting heavily applications calling APIs returning `204` / `205`, especially if it's making a lot of those calls.

JakeWharton:
We don't give ETAs for releases because these are projects and not
products. That said the next release is blocked by OkHttp releasing which
is blocked by Okio releasing. We publish snapshot versions of all our
projects to Sonatype's snapshot repo which you can depend on or download
.jars of for fixed versions for now.

adrien-aubel:
Sounds good, I'll use 2.1.1-SNAPSHOT for now. Thanks!

Issue #2037, opened 2016-10-01
Explicit usage of File objects in @Multipart methods with @Part and @PartMap


papcio28:
I want to use File object explicitly in Retrofit interfaces without manually creating `MultipartBody.Part`.  It should be available in single parameters with `@Part` and `@PartMap` annotations.

Example :

```
public interface FileService {
    @Multipart
    @POST("/upload")
    Call<ResponseBody> sendFile(@Part("file") File fileToSend);
}
```

I will provide a pull request with tests and implementation.

JakeWharton:
This can be accomplished with:
```java
public final class FileBodyConverter implements Converter<File, RequestBody> {
  public static final Converter.Factory FACTORY = new Factory() {
    @Override public Converter<?, RequestBody> requestBodyConverter(Type type,
        Annotation[] parameterAnnotations, Annotation[] methodAnnotations, Retrofit retrofit) {
      return File.class.equals(type) ? new FileBodyConverter() : null;
    }
  }
  
  @Override public RequestBody convert(File value) throws IOException {
    return RequestBody.create(null, value);
  }
}
```
and since it's unclear what content-type to use for a `File` I don't want to ship a general version of this.

PR #2038, opened 2016-10-01
Explicit usage of File objects in @Multipart methods

papcio28:
This feature allow developers to explicitly use `File` objects in Retrofit services with `@Part("name")` and `@PartMap` (`Map<String, File>`) annotations - issue #2037

Provided implementation and tests.

### DIFF
--- a/retrofit/src/main/java/retrofit2/ParameterHandler.java
+++ b/retrofit/src/main/java/retrofit2/ParameterHandler.java
@@ -257,6 +257,10 @@ abstract class ParameterHandler<T> {
     }
 
     @Override void apply(RequestBuilder builder, File value) throws IOException {
+      apply(builder, name, value);
+    }
+
+    static void apply(RequestBuilder builder, String name, File value) {
       if (value != null) { // Skip null values.
         builder.addPart(MultipartBody.Part.createFormData(name,
                 value.getName(),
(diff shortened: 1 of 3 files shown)


jaredsburrows:
@papcio28 You might want to setup your git config: https://help.github.com/articles/setting-your-username-in-git/.

papcio28:
@jaredsburrows - I didn't noticed that I have a wrong setup. Thank you. I will try to fix it soon.

JakeWharton:
This implementation looks a bit overkill. I'm pretty sure this could be accomplished by writing a `Converter` that took a `File` and wrapped it in the `ResponseBody` and then the normal mechanisms that are already in place would just work.

ylfzq:
Actually we can't archieve this by writing a `Converter` elegantly. Because if we want to make `Converter` effective, we must use `@Body` annotation, in that way, we can't use `@MultiPart` as <https://github.com/square/retrofit/blob/master/retrofit/src/main/java/retrofit2/ServiceMethod.java#L214>. 

```java
@MultiPart
Call<SomeResponse> uploadFile(@Body File file);	// it's error, for the link above. it requires at least one @Part annoation in method params
```

Alternatively, we can define like:

```java
@Headers("ContentType: multipart/form-data; boundary=--custom_boundary")
Call<SomeResponse> uploadFile(@Body File file);	// it's ok, but I said it's not elegant
```



I really want to add this feature when my first use of Retrofit. And I did so(I forked retrofit, and achieved this). But nowadays, I find the support of File type is not a necessary function, although it's used quiet often. We can achieve elegantly without any source code changing like:

```java
public interface SampleInterface {

  @POST("samplePath")
  @Multipart
  Call<ResponseBody> uploadApk(@PartMap Params params);
}
```

the `Params` definition is:

```java
public final class Params extends HashMap<String, RequestBody> {

  public Params(File someFile, String appId, String title, String description) {
    putFile("file", someFile);
    putString("app_id", appId);
    putString("title", title);
    putString("description", description);
    putString("pid", "1");
    putString("secret", "1");
  }

  public void putString(String key, String value) {
    put(key, RequestBody.create(MediaType.parse("text/*"), value));
  }

  public void putFile(String key, File file) {
    put(key + "\"; filename=\"" + file.getName(),
        RequestBody.create(MediaType.parse("*/*"), file));
  }
}
```

so, we can use like this:
```java
uploadApk(new Params(new File("someFile"), "app_id", "title", "description"));
```
it's quiet elegant, I think

papcio28:
@ylfzq Thank for support. As for your solution - it's sounds great, but you don't have any validation of required parameters (because you need to use `@PartMap`) and still have to remember about correct adding File as part (`put(key + "\"; filename=\"" + file.getName(), RequestBody.create(MediaType.parse("*/*"), file));`). I think my solution will be quite usable, because developer will be able to put Files in request without any extra effort and boilerplate code and still it doesn't change or broke any other Retrofit features.

JakeWharton:
Request body converters are used for multipart parts. I don't think we need to ship this because it can be implemented thusly:
```java
public final class FileBodyConverter implements Converter<File, RequestBody> {
  public static final Converter.Factory FACTORY = new Factory() {
    @Override public Converter<?, RequestBody> requestBodyConverter(Type type,
        Annotation[] parameterAnnotations, Annotation[] methodAnnotations, Retrofit retrofit) {
      return File.class.equals(type) ? new FileBodyConverter() : null;
    }
  }
  
  @Override public RequestBody convert(File value) throws IOException {
    return RequestBody.create(null, value);
  }
}
```

Issue #2054, opened 2016-10-20
[Request] Add "newBuilder" to Retrofit class


jaredsburrows:
Can we add `newBuilder` builder similar to Okhttp3 (see [here](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/OkHttpClient.java#L390)) to `Retrofit.Builder`?

I would like to change the URL of an already created builder.

JakeWharton:
Why do you need it?

jaredsburrows:
The same reason it is availble in OkHttp3.

For instance:

```
// Some Client
    OkHttpClient client = new OkHttpClient.Builder()
      .addInterceptor(new HttpLoggingInterceptor()
        .setLevel(HttpLoggingInterceptor.Level.BODY))
      .cache(cache)
      .build();
    
// Some client 2
    OkHttpClient client2 = client.newBuilder().connectTimeout(CLIENT_TIME_OUT, TimeUnit.SECONDS)
      .writeTimeout(CLIENT_TIME_OUT, TimeUnit.SECONDS)
      .readTimeout(CLIENT_TIME_OUT, TimeUnit.SECONDS)
      .build();
```

I can reuse the existing builder to create another client and change some the existing parameters. I would like to be able to do the same with Retrofit 2.

JakeWharton:
I know what it's function is, but what specifically would you use it for?
Unlike OkHttpClient the Retrofit instance is much more dumb and
necessitates something like this much less. So actual motivations are much
more compelling than theoretical ones (which is why it's currently absent).

jaredsburrows:
It would be nice to easily call `newBuilder` and pass in the url from the Okhttp mockserver for testing. But in general, it would be nice to change some the existing parameters.

ly85206559:
U can hold one Retrofit.Builder instance, then you can create different Retrofit instance with this builder.
like this:
Retrofit.Builder builder = Retrofit.Builder()
                .addConverterFactory(GsonConverterFactory.create())
                .addCallAdapterFactory(RxJavaCallAdapterFactory.create())
                .client(okHttpClient);
Retrofit retrofit1 = builder
                .baseUrl("xxx")
                .build();

Retrofit retrofit2 = builder
                .baseUrl("xxxxx")
                .build();
Do you mean this?

ly85206559:
U can hold one Retrofit.Builder instance, then you can create different Retrofit instance with this builder.
like this:
[...]

PR #2056, opened 2016-10-22
Add newBuilder method to retrofit

jaredsburrows:
@JakeWharton Based on our discussion here: #2054

### DIFF
--- a/retrofit/src/main/java/retrofit2/Retrofit.java
+++ b/retrofit/src/main/java/retrofit2/Retrofit.java
@@ -59,16 +59,18 @@ import static retrofit2.Utils.checkNotNull;
 public final class Retrofit {
   private final Map<Method, ServiceMethod<?, ?>> serviceMethodCache = new ConcurrentHashMap<>();
 
-  private final okhttp3.Call.Factory callFactory;
-  private final HttpUrl baseUrl;
-  private final List<Converter.Factory> converterFactories;
-  private final List<CallAdapter.Factory> adapterFactories;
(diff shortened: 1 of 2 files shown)


retrofit/src/main/java/retrofit2/Retrofit.java
```diff
@@ -405,6 +432,16 @@ public Builder() {
       this(Platform.get());
     }
 
+    public Builder(Retrofit retrofit) {
+      this.platform = retrofit.platform;
+      this.callFactory = retrofit.callFactory;
+      this.baseUrl = retrofit.baseUrl;
+      this.converterFactories = retrofit.converterFactories;
```
JakeWharton: All lists should use `addAll` and not retain these unmodifiable references.

retrofit/src/main/java/retrofit2/Retrofit.java
```diff
@@ -53,22 +53,45 @@
  * Response&lt;User&gt; user = api.getUser().execute();
  * </code></pre>
  *
+ * <h3>Customize your rest adapter with newBuilder()</h3>
```
JakeWharton: What's a "rest adapter"?!? I don't think we want any of this doc. Unlike OkHttp, this isn't something we expect many to do (seeing as we haven't had the ability for Retrofit's 5 years of existence).

retrofit/src/main/java/retrofit2/Retrofit.java
jaredsburrows: Good eye.

PR #2057, opened 2016-10-22
RetrofitError.httpError() has unknown cause

pyricau:
In Retrofit 1.x, the exception created by RetrofitError.httpError() has a null cause (aka unknown cause), instead of having a nonexistent cause.

This causes issues with RxJava assembly tracking, see [here](https://github.com/ReactiveX/RxJava/issues/4737).

The following code crashes:

```
RxJavaHooks.enableAssemblyTracking()
// Will crash if the endpoint returns an http error.
myService.endpointReturningHttpError().subscribe(...);
```


pyricau:
We all should move to 2.x, but this takes time in large codebases. In the meantime, it'd be nice if we could still enable Rx Java assembly tracking.

PR #2082, opened 2016-11-15
Add README docs for adapters

Jawnnypoo:
It can be confusing or tricky to understand what the purpose is or what exactly a `CallAdapter.Factory` does. For instance, I spent quite a bit of time struggling and Googling before I realized you could get an `Observable<Response<Model>>` with an RxJava adapter. This adds some basic docs to each of the supported adapters, similar to how each `Converter` has its own docs. 

I am not super familiar with what all types are supported in the RxJava 1.X adapter, so that should be checked. I pulled most of the README for RxJava 2.X adapter from [Jake's project](https://github.com/JakeWharton/retrofit2-rxjava2-adapter)

Happy to make changes as needed!

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,5 +47,5 @@ License
 
 
  [1]: http://square.github.io/retrofit/
- [2]: https://search.maven.org/remote_content?g=com.squareup.retrofit&a=retrofit&v=LATEST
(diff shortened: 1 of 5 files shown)


Jawnnypoo:
Updated this to match #2093 and also as a bonus, fixed the JAR link on the root README to point to the latest group ID

JakeWharton:
Thanks for doing this! I combined it with the other PR and merged manually.

PR #2090, opened 2016-11-19
proguard update

RicardoBravoA:
Doesn't allow generate an apk when using proguard
- Warning:there were 14 unresolved references to classes or interfaces.
- Warning:Exception while processing task java.io.IOException: Please correct the above warnings first.
- Error:Execution failed for task ':app:transformClassesAndResourcesWithProguardForRelease'.
> java.io.IOException: Please correct the above warnings first.

### DIFF
--- a/website/index.html
+++ b/website/index.html
@@ -183,6 +183,7 @@ compile 'com.squareup.retrofit2:retrofit:<span class="version pln"><em>(insert l
 -keepattributes Signature
 # Retain declared checked exceptions for use by a Proxy instance.
 -keepattributes Exceptions
+-dontwarn okio.**
 </pre>
             </section>
 


Jawnnypoo:
I think since this is actually a ProGuard issue with Okio, it should if anything be a note or a link below the actual proguard configuration, to differentiate.

JakeWharton:
It's not really an issue, but yes it's for Okio and not Retrofit directly.
We should document the Okio rules and just link to them. I do not want them
duplicated.

Jawnnypoo:
Would you like me to make a PR on Okio? Or are you on it?

JakeWharton:
See #2119

Issue #2102, opened 2016-12-01
NPE when retrofit used by IntelliJ IDEA plugin


baron1405:
What kind of issue is this?

 - [ ] Question. This issue tracker is not the place for questions. If you want to ask how to do
       something, or to understand why something isn't working the way you expect it to, use Stack
       Overflow. https://stackoverflow.com/questions/tagged/retrofit

 - [X ] Bug report. If you’ve found a bug, spend the time to write a failing test. Bugs with tests
       get fixed. Here’s an example: https://gist.github.com/swankjesse/6608b4713ad80988cdc9

 - [ ] Feature Request. Start by telling us what problem you’re trying to solve. Often a solution
       already exists! Don’t send pull requests to implement new features without first getting our
       support. Sometimes we leave features out on purpose to keep the project small.

We are developing a plugin for IntelliJ IDEA Ultimate. The plugin uses retrofit for communication with a service. The plugin is crashing due to a NullPointerException resulting from retrofit's platform detection getting confused by the presence of an Android class on the IntelliJ IDEA classpath. The class is present due to IntelliJ's bundled Android Support plugin. The stack trace is:
```
2016-11-30 14:30:27,654 [   7389]  ERROR - plication.impl.ApplicationImpl - null
java.lang.ExceptionInInitializerError
        at java.lang.Class.forName0(Native Method)
        at java.lang.Class.forName(Class.java:264)
        at retrofit2.Platform.findPlatform(Platform.java:37)
        at retrofit2.Platform.<clinit>(Platform.java:29)
        at retrofit2.Retrofit$Builder.<init>(Retrofit.java:402)
        ...
        at com.intellij.openapi.application.impl.ApplicationImpl$3.call(ApplicationImpl.java:331)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.NullPointerException
        at android.os.SystemProperties_Delegate.native_get(SystemProperties_Delegate.java:44)
        at android.os.SystemProperties.native_get(SystemProperties.java)
        at android.os.SystemProperties.get(SystemProperties.java:64)
        at android.os.Build.getString(Build.java:819)
        at android.os.Build.<clinit>(Build.java:39)
[...]

PR #2103, opened 2016-12-01
Prevent NPE when retrofit is used by an IntelliJ plugin. The Android …

baron1405:
…Support plugin in IntelliJ IDEA Ultimate makes retrofit incorrectly think it is running in an Android environment. This is a fix for retrofit issue #2102.

### DIFF
--- a/retrofit/src/main/java/retrofit2/Platform.java
+++ b/retrofit/src/main/java/retrofit2/Platform.java
@@ -32,12 +32,14 @@ class Platform {
   }
 
   private static Platform findPlatform() {
-    try {
-      Class.forName("android.os.Build");
-      if (Build.VERSION.SDK_INT != 0) {
-        return new Android();
+    if ("Dalvik".equals(System.getProperty("java.vm.name"))) {
+      try {
+        Class.forName("android.os.Build");
+        if (Build.VERSION.SDK_INT != 0) {
+          return new Android();
+        }
+      } catch (ClassNotFoundException ignored) {
       }
-    } catch (ClassNotFoundException ignored) {
     }
     try {
       Class.forName("java.util.Optional");


retrofit/src/main/java/retrofit2/Platform.java
```diff
@@ -32,12 +32,14 @@ static Platform get() {
   }
 
   private static Platform findPlatform() {
-    try {
-      Class.forName("android.os.Build");
-      if (Build.VERSION.SDK_INT != 0) {
-        return new Android();
+    if ("Dalvik".equals(System.getProperty("java.vm.name"))) {
```
JakeWharton: This seems like enough of an indicator where we don't even need to bother with the `Class.forName` check?

@swankjesse thoughts?

swankjesse: Will it work if we look for `android.os.Looper` instead? Seems like if that class is present then we should have Android behavior

baron1405: BTW you can replace the forName check for Java 8 by using the `java.class.version` property. Any value >= 52 is Java 8 or newer.

baron1405: I am concerned that attempting to load any Android class runs the risk of undesirable consequences especially in an IDE environment. Case is point is the android.os.Builder class was not on the classpath in the IDEA 2016.2 version of the Android Support plugin but has appeared on the classpath in the 2016.3 version. In other words, it is impossible to predict what Google and JetBrains will expose either intentionally or accidentally. A property-based approach would be the safest.

JakeWharton: True. It also affects our ability to test this behavior since the Android classes are on the classpath but aren't actually usable.

ZhangKuixun:
I delete app/build file ,then run app. ok

PR #2107, opened 2016-12-03
Revert back to an OnSubscribe-based RxJava adapter.

### LINKED ISSUE #2048: RxJavaCallAdapterFactory and HTTP requests cancellation
The d6666438ea50137df49957d25e62639b86b59197 commit changed `RxJavaCallAdapterFactory` internals to work with `Observable.fromCallable` instead of custom `OnSubscribe`. It greatly simplifies the implementation, but it seems like it omits such important thing as HTTP request cancellation.

The previous implementation called `retrofit2.Call.cancel` which in its turn called `okhttp3.Call.cancel` under the hood. As the `retrofit2.Call.cancel` documentation states `Cancel this call. An attempt will be made to cancel in-flight calls, and if the call has not yet been executed it never will be`. It seems like the `Observable.fromCallable` covers the second part about never executing the request if it wasn’t already, but misses the in-flight call situation. Was this change intentional? As far as I understand OkHttp reacts on `okhttp3.Call.cancel` by closing a connection socket which is actually a great thing.

JakeWharton:
This allows proper cancelation of the underlying Call which wasn't supported in the fromCallable method.

Closes #2048. This started as a revert of d6666438ea50137df49957d25e62639b86b59197.


JakeWharton:
meh. not fixing checkstyle tonight.

artem-zinnatullin:
Maybe better force movement fromEmitter to stable API?

It'll help many library developers and you won't need to fix checkstyle in
this PR!

JakeWharton:
Sure, but we need a release here sooner than any 1.3.0 could be released. I
can't wait until it's stable.

JakeWharton:
@swankjesse ping!

PR #2109, opened 2016-12-05
Update checkstyle to the latest version.

JakeWharton:
Should fix the failures on #2107.


checkstyle.xml
```diff
@@ -123,6 +122,7 @@
         <module name="UpperEll"/>
         <module name="Indentation">
           <property name="basicOffset" value="2"/>
+          <property name="caseIndent" value="2"/>
```
JakeWharton: Looks like this was the only thing needed. Kept the rest anyway.

JakeWharton:
TBR

PR #2119, opened 2016-12-09
Okio proguard rules link

Jawnnypoo:
As discussed in https://github.com/square/retrofit/pull/2090 this links out to Okio for its ProGuard rules. 

capture


JakeWharton:
Brilliant! Thanks.

PR #2126, opened 2016-12-16
Add optional headers to @Part annotation

iagreen:
Add a `headers` property to the `Part` annotation to allow for custom headers inside each part of the multipart body.

Inspired by this [SO question](http://stackoverflow.com/q/41164123/1904517).

### DIFF
--- a/retrofit/src/main/java/retrofit2/ServiceMethod.java
+++ b/retrofit/src/main/java/retrofit2/ServiceMethod.java
@@ -594,9 +594,10 @@ final class ServiceMethod<R, T> {
                 "@Part annotation must supply a name or use MultipartBody.Part parameter type.");
           }
         } else {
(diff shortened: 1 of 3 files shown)


JakeWharton:
Honestly I'm really hesitant about exposing this as an API. Can you talk about your specific use case more? I'm tempted to say that you should just use a `MultipartBody.Part` as the type which allows specifying custom headers in the runtime instance and not the compile-time Retrofit annotations.

anninos:
So how can you pass/set custom headers to a part?

iagreen:
@JakeWharton This totally fell of my radar, but I agree with you.  It is probably better to build the `Part` than annotate it.

anninos:
Because I am not really sure what you mean by that, can you give me an example or something?

PR #2134, opened 2016-12-24
Add Completable to documentation in RxJava2CallAdapterFactory

vanniktech:


retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/RxJava2CallAdapterFactory.java
```diff
@@ -41,7 +41,7 @@
  * }
  * </code></pre>
  * There are three configurations supported for the {@code Observable}, {@code Flowable},
```
JakeWharton: Four

vanniktech: Nope. With configurations: `Observable<User>`, `Observable<Response<User>>` and `Observable<Result<User>>` are meant. Also if so there are 5 supported types (Observable, Flowable, Maybe, Single and Completable).

JakeWharton: What a terribly worded sentence this is then.

PR #2159, opened 2017-01-17
Add AnnotatedConverterFactory sample

NightlyNexus:
Stole the idea and most of the code from one of Jake's recent talks.

Not sure what you had in mind here, or if you want it in the samples at all, but here it is in case it's helpful to anyone else.

### DIFF
--- a/samples/src/main/java/com/example/retrofit/AnnotatedConverters.java
+++ b/samples/src/main/java/com/example/retrofit/AnnotatedConverters.java
@@ -1,0 +1,140 @@
+package com.example.retrofit;
+
+import com.google.gson.GsonBuilder;
+import java.io.IOException;
+import java.lang.annotation.Annotation;
+import java.lang.annotation.Retention;
+import java.lang.reflect.Type;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import okhttp3.RequestBody;
+import okhttp3.ResponseBody;
+import okhttp3.mockwebserver.MockResponse;
+import okhttp3.mockwebserver.MockWebServer;
+import org.simpleframework.xml.Attribute;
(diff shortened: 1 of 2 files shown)


samples/src/main/java/com/example/retrofit/AnnotatedConverters.java
```diff
@@ -0,0 +1,140 @@
+package com.example.retrofit;
```
JakeWharton: license

samples/src/main/java/com/example/retrofit/AnnotatedConverters.java
```diff
@@ -0,0 +1,140 @@
...
+          return factory.responseBodyConverter(type, annotations, retrofit);
+        }
+      }
+      return null;
+    }
+
+    @Override public Converter<?, RequestBody> requestBodyConverter(Type type,
+        Annotation[] parameterAnnotations, Annotation[] methodAnnotations, Retrofit retrofit) {
+      for (Annotation annotation : parameterAnnotations) {
+        Converter.Factory factory = factories.get(annotation.annotationType());
```
NightlyNexus: Oops, I would have done a patch commit if I had seen you had seen this first.
I fixed this just now; I had `annotation.getClass()` before.

samples/src/main/java/com/example/retrofit/AnnotatedConverters.java
```diff
@@ -1,3 +1,18 @@
+/*
+ * Copyright (C) 2016 Square, Inc.
```
JakeWharton: 2017

JakeWharton:
Thanks!

PR #2160, opened 2017-01-20
Fix Error Message for Maybe if it's not parameterized and add RxJava2CallAdapterFactoryTest tests.

vanniktech:
- fixes an error message when using `Maybe`
- adds RxJava2CallAdapterFactoryTest that are based off RxJavaCallAdapterFactoryTest with additions of `Flowable` & `Maybe`


retrofit-adapters/rxjava2/src/test/java/retrofit2/adapter/rxjava2/RxJava2CallAdapterFactoryTest.java
```diff
@@ -0,0 +1,274 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
```
vanniktech: Wasn't sure whether it should be 2015 or 2016. Sticked to 2015 since that's also the date from `RxJavaCallAdapterFactoryTest`

JakeWharton:
Thanks

PR #2165, opened 2017-01-25
Add failOnUnknown to MoshiConverterFactory

NightlyNexus:
I don't actually have a use case. It just seems missing.
`asLenient` does not have a test, either; should it have some simple test?
`asFailOnUnknown` sounds awkward...

### DIFF
--- a/retrofit-converters/moshi/src/main/java/retrofit2/converter/moshi/MoshiConverterFactory.java
+++ b/retrofit-converters/moshi/src/main/java/retrofit2/converter/moshi/MoshiConverterFactory.java
@@ -51,27 +51,37 @@ public final class MoshiConverterFactory extends Converter.Factory {
   /** Create an instance using {@code moshi} for conversion. */
   public static MoshiConverterFactory create(Moshi moshi) {
     if (moshi == null) throw new NullPointerException("moshi == null");
-    return new MoshiConverterFactory(moshi, false, false);
+    return new MoshiConverterFactory(moshi, false, false, false);
   }
 
   private final Moshi moshi;
   private final boolean lenient;
+  private final boolean failOnUnknown;
   private final boolean serializeNulls;
 
-  private MoshiConverterFactory(Moshi moshi, boolean lenient, boolean serializeNulls) {
+  private MoshiConverterFactory(Moshi moshi, boolean lenient, boolean failOnUnknown,
+      boolean serializeNulls) {
     this.moshi = moshi;
     this.lenient = lenient;
+    this.failOnUnknown = failOnUnknown;
     this.serializeNulls = serializeNulls;
   }
 
   /** Return a new factory which uses {@linkplain JsonAdapter#lenient() lenient} adapters. */
   public MoshiConverterFactory asLenient() {
-    return new MoshiConverterFactory(moshi, true, serializeNulls);
+    return new MoshiConverterFactory(moshi, true, failOnUnknown, serializeNulls);
+  }
+
+  /**
+   * Return a new factory which uses {@link JsonAdapter#failOnUnknown()} adapters.
+   */
+  public MoshiConverterFactory failOnUnknown() {
+    return new MoshiConverterFactory(moshi, lenient, true, serializeNulls);
   }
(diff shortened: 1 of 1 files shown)


retrofit-converters/moshi/src/main/java/retrofit2/converter/moshi/MoshiConverterFactory.java
```diff
@@ -81,16 +92,21 @@ public MoshiConverterFactory withNullSerialization() {
     if (lenient) {
       adapter = adapter.lenient();
     }
+    if (failOnUnknown) {
+      adapter.failOnUnknown();
```
swankjesse: this would need to be `adapter = adapter.failOnUnknown()`

NightlyNexus: haha, ouch. thanks.

retrofit-converters/moshi/src/main/java/retrofit2/converter/moshi/MoshiConverterFactory.java
```diff
@@ -51,27 +51,38 @@ public static MoshiConverterFactory create() {
...
   }
 
   /** Return a new factory which uses {@linkplain JsonAdapter#lenient() lenient} adapters. */
   public MoshiConverterFactory asLenient() {
-    return new MoshiConverterFactory(moshi, true, serializeNulls);
+    return new MoshiConverterFactory(moshi, true, failOnUnknown, serializeNulls);
+  }
+
+  /**
+   * Return a new factory which uses {@linkplain JsonAdapter#failOnUnknown()}  failOnUnknown}
```
JakeWharton: Extra `}` and double space

JakeWharton: Also since you're using `linkplain` it should either say "fail on unknown" or be changed to just `link`.

retrofit-converters/moshi/src/main/java/retrofit2/converter/moshi/MoshiConverterFactory.java
```diff
@@ -51,27 +51,38 @@ public static MoshiConverterFactory create() {
...
+   * adapters.
+   */
+  public MoshiConverterFactory asFailOnUnknown() {
```
JakeWharton: Does "as" make sense as a prefix here? Or should it be "with"? both sound weird.

NightlyNexus:
>Does "as" make sense as a prefix here? Or should it be "with"? both sound weird.

Yeah, "as" definitely sounds awkward.
It'd be nice if `failOnUnknown` could line up nicely with `asLenient`, but it might not be worth the awkwardness.
Maybe no prefix?

retrofit-converters/moshi/src/main/java/retrofit2/converter/moshi/MoshiConverterFactory.java
```diff
@@ -51,27 +51,37 @@ public static MoshiConverterFactory create() {
...
   /** Return a new factory which uses {@linkplain JsonAdapter#lenient() lenient} adapters. */
   public MoshiConverterFactory asLenient() {
-    return new MoshiConverterFactory(moshi, true, serializeNulls);
+    return new MoshiConverterFactory(moshi, true, failOnUnknown, serializeNulls);
+  }
+
+  /**
+   * Return a new factory which uses {@link JsonAdapter#failOnUnknown()} adapters.
+   */
+  public MoshiConverterFactory asFailOnUnknown() {
```
JakeWharton: I think we just go with `failOnUnknown()` here.

NightlyNexus: Yep, I think killing the prefix is the best option. will fix.

JakeWharton:
Merged manually as 4ec773178f8a992f9750ec4e7a57d2839530a075 with a test.

PR #2168, opened 2017-01-28
Rename LICENSE.txt to LICENSE

SUPERCILEX:
This is the new default for GitHub and looks much nicer.


JakeWharton:
This would probably be annoying for Windows users. Even disregarding that, I see no reason for the change.

SUPERCILEX:
@JakeWharton Since GitHub uses `LICENSE` by default, license displayer libraries are going to make that their default. As for it being annoying for Windows users, who reads the license anyway? I doubt that will much of an issue. However, it makes it easier for license libraries if you follow the GitHub default.

JakeWharton:
No license checker of any quality would omit checking for a .txt version.

SUPERCILEX:
Of course not, but they would start by checking for GitHub's default and
then the .txt version. Heck, even you use the non .txt version for some of
your repos: [
https://github.com/ReactiveX/RxAndroid/blob/2.x/**LICENSE**](https://github.com/ReactiveX/RxAndroid/blob/2.x/LICENSE)
.

JakeWharton:
I didn't create that repo

SUPERCILEX:
@JakeWharton Yeah, but you maintain it.

@JakeWharton Fine, I tried. I created https://github.com/yshrsmz/LicenseAdapter/pull/22 anyway to fix this problem so it's good, I guess. I still prefer `LICENSE` instead of `LICENSE.txt`. 😄

Issue #2132, opened 2016-12-21
Evaluate adding createAsync() to the RxJava adapters.


JakeWharton:
This would use `enqueue` instead of `execute` to defer threading to OkHttp's `Dispatcher` instead of blocking a scheduler (or application) thread. This would allow re-using 90% of the built-in machinery around things like type parsing, `Result`/body handling, etc.

PR #2169, opened 2017-01-28
Add createAsync() to RxJava 2.x CallAdapter.Factory.

JakeWharton:
This uses Call.enqueue() instead of Call.execute() when invoking the underlying Call.

Refs #2132.

### DIFF
--- a/retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/CallEnqueueObservable.java
+++ b/retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/CallEnqueueObservable.java
@@ -22,50 +22,67 @@ import io.reactivex.exceptions.CompositeException;
 import io.reactivex.exceptions.Exceptions;
 import io.reactivex.plugins.RxJavaPlugins;
 import retrofit2.Call;
+import retrofit2.Callback;
 import retrofit2.Response;
 
-final class CallObservable<T> extends Observable<Response<T>> {
+final class CallEnqueueObservable<T> extends Observable<Response<T>> {
   private final Call<T> originalCall;
 
-  CallObservable(Call<T> originalCall) {
+  CallEnqueueObservable(Call<T> originalCall) {
     this.originalCall = originalCall;
   }
 
   @Override protected void subscribeActual(Observer<? super Response<T>> observer) {
     // Since Call is a one-shot type, clone it for each new observer.
     Call<T> call = originalCall.clone();
-    observer.onSubscribe(new CallDisposable(call));
+    CallCallback<T> callback = new CallCallback<>(call, observer);
+    observer.onSubscribe(callback);
+    call.enqueue(callback);
+  }
 
+  private static final class CallCallback<T> implements Disposable, Callback<T> {
+    private final Call<?> call;
+    private final Observer<? super Response<T>> observer;
     boolean terminated = false;
-    try {
-      Response<T> response = call.execute();
-      if (!call.isCanceled()) {
+
+    CallCallback(Call<?> call, Observer<? super Response<T>> observer) {
+      this.call = call;
+      this.observer = observer;
+    }
+
+    @Override public void onResponse(Call<T> call, Response<T> response) {
+      if (call.isCanceled()) return;
+
+      try {
         observer.onNext(response);
-      }
-      if (!call.isCanceled()) {
-        terminated = true;
-        observer.onComplete();
-      }
-    } catch (Throwable t) {
-      Exceptions.throwIfFatal(t);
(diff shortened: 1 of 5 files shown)


JakeWharton:
Will do RxJava 2.x next.

eneim:
Sorry, this is actually the implementation for RxJava2 I think :-?

retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/RxJava2CallAdapterFactory.java
```diff
@@ -59,7 +59,15 @@
    * by default.
    */
   public static RxJava2CallAdapterFactory create() {
-    return new RxJava2CallAdapterFactory(null);
+    return new RxJava2CallAdapterFactory(null, false);
+  }
+
+  /**
+   * Returns an instance which creates asynchronous observables. Applying
```
artem-zinnatullin: Returns an instance which creates asynchronous observables, network requests will be performed on OkHttp's `Dispatcher` instead of blocking a scheduler (or application) thread?

retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/CallEnqueueObservable.java
```diff
@@ -0,0 +1,96 @@
...
+
+      try {
+        observer.onNext(response);
+
+        if (!call.isCanceled()) {
+          terminated = true;
+          observer.onComplete();
+        }
+      } catch (Throwable t) {
+        if (terminated) {
```
artem-zinnatullin: Maybe test this since it's a major part of Observable contract?

retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/RxJava2CallAdapterFactory.java
JakeWharton: OkHttp is not guaranteed

JakeWharton:
@eneim So it is. They're all the same to me.

retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/RxJava2CallAdapterFactory.java
artem-zinnatullin: Ah, right, "network requests will be performed by `Call` dispatcher (ie OkHttp) instead of blocking a scheduler or application thread"? 

I mean, would be great to include some pros of using it into javadoc (and release notes)!

retrofit-adapters/rxjava2/src/test/java/retrofit2/adapter/rxjava2/AsyncTest.java
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  @Test public void async() throws InterruptedException {
+    server.enqueue(new MockResponse());
+
+    TestObserver<Void> observer = new TestObserver<>();
+    service.completable().subscribe(observer);
+    assertFalse(observer.await(1, SECONDS));
+
+    triggerResponse();
```
swankjesse: fun fact: if you wanted to do all the async through QueueDispatcher, this line becomes just `server.enqueue(MockResponse())` and you don’t need your own triggers

JakeWharton: Is that true? I recall trying that but Dispatcher returns a default response if one isn't available when I looked.

JakeWharton: It was `peek()` that returns a default, not `dispatch()`. Updated

retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/RxJava2CallAdapterFactory.java
JakeWharton: A dispatcher is an implementation detail of OkHttp. All we can guarantee is that `Call.enqueue` is used and I'm not sure that specific information is any more useful than calling them "asynchronous".

retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/CallEnqueueObservable.java
JakeWharton: Added

PR #2172, opened 2017-01-30
Add createAsync() to RxJava 1.x CallAdapter.Factory.

### LINKED ISSUE #2132: Evaluate adding createAsync() to the RxJava adapters.

JakeWharton:
This uses Call.enqueue() instead of Call.execute() when invoking the underlying Call.

Closes #2132.


retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/CallEnqueueOnSubscribe.java
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  @Override public void call(Subscriber<? super Response<T>> subscriber) {
+    // Since Call is a one-shot type, clone it for each new subscriber.
+    Call<T> call = originalCall.clone();
+    final CallArbiter<T> arbiter = new CallArbiter<>(call, subscriber);
+    subscriber.add(arbiter);
+    subscriber.setProducer(arbiter);
+
+    call.enqueue(new Callback<T>() {
```
JakeWharton: The only new code in this PR is this `Callback` implementation. Everything else was here previously.

PR #2178, opened 2017-01-30
Allow HTTP errors to be created as part of mock behavior.

### LINKED ISSUE #738: MockRetrofit failure mode should support non-2xx responses.
If not as a first-class citizen, then transition to something where we get a callback and can either throw directly or return a Response to enable that functionality.

JakeWharton:
Closes #738.

### DIFF
--- a/retrofit-mock/src/main/java/retrofit2/mock/NetworkBehavior.java
+++ b/retrofit-mock/src/main/java/retrofit2/mock/NetworkBehavior.java
@@ -17,7 +17,10 @@ package retrofit2.mock;
 
 import java.io.IOException;
 import java.util.Random;
+import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
+import okhttp3.ResponseBody;
+import retrofit2.Response;
 
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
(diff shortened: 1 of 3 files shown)


retrofit-mock/src/main/java/retrofit2/mock/NetworkBehavior.java
```diff
@@ -120,28 +130,76 @@ public int failurePercent() {
...
+  public void setErrorPercent(int errorPercent) {
+    if (errorPercent < 0 || errorPercent > 100) {
+      throw new IllegalArgumentException("Error percentage must be between 0 and 100.");
+    }
+    this.errorPercent = errorPercent;
+  }
+
+  /**
+   * Set the error response factory to be used when an error is triggered. This factory may only
+   * return responses for which {@link Response#isSuccessful()} returns false.
```
swankjesse: Nice doc.

retrofit-mock/src/test/java/retrofit2/mock/NetworkBehaviorTest.java
```diff
@@ -72,6 +76,97 @@
...
+      private int code = 500;
+
+      @Override public Response<?> call() throws Exception {
+        return Response.error(code++, ResponseBody.create(null, new byte[0]));
+      }
+    });
+
+    assertEquals(500, behavior.createErrorResponse().code());
+    assertEquals(501, behavior.createErrorResponse().code());
+    assertEquals(502, behavior.createErrorResponse().code());
```
swankjesse: Nice feature

PR #2181, opened 2017-01-30
Add support for query parameters with no value.

### LINKED ISSUE #1997: [Feature Request] Possibility to add empty parameters
My app has to communicate with an API that uses RQL. RQL uses parameters such as limit(x,y) to perform certain tasks on the DB. Is there a way to add these parameters? If I pass a parameter without a value, the parameters are not added to the link. 

Example: `api.acme.com/users`  is my endpoint to list all the users. If I want to limit this list to 10 users, I have to use `api.acme.com/users?limit(10,0)`. So I have to use only the key of the parameter, without a value. 

The second problem with these parameters is that part of the name has to change depending on the needs. The placeholder method `/limit({limit},{offset}` gives me an error, So I have to pass the whole generated URL for every request. 

I know RQL is a relative new technique and isn't that widely implemented yet, but it would be a nice feature to have.

JakeWharton:
Closes #1997.

### DIFF
--- a/retrofit/src/main/java/retrofit2/http/QueryName.java
+++ b/retrofit/src/main/java/retrofit2/http/QueryName.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2014 Square, Inc.
+ * Copyright (C) 2013 Square, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


retrofit/src/main/java/retrofit2/http/QueryName.java
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Query parameter appended to the URL that has no value.
+ * <p>
+ * Passing a {@link java.util.List List} or array will result in a query parameter for each
+ * non-{@code null} item.
+ * <p>
+ * Simple Example:
+ * <pre><code>
+ * &#64;GET("/list")
```
swankjesse: Nit: the word list here is a bit distracting because the examples deal with multiplicity.

retrofit/src/test/java/retrofit2/RequestBuilderTest.java
```diff
@@ -1099,6 +1100,93 @@
...
+
+    Object[] values = { 1, 2, null, "three" };
+    Request request = buildRequest(Example.class, new Object[] { values });
+    assertThat(request.method()).isEqualTo("GET");
+    assertThat(request.headers().size()).isZero();
+    assertThat(request.url().toString()).isEqualTo("http://example.com/foo/bar/?1&2&three");
+    assertThat(request.body()).isNull();
+  }
+
+  @Test public void getWithQueryNameParamPrimitiveArray() {
```
swankjesse: Maybe also test a repeated key? Should Just Work.

JakeWharton:
@swankjesse we don't allow `null` values in any maps currently so I thought it weird to allow it just for this one. Also it doesn't allow repeated values (although that seems like a rare need in this case.

retrofit/src/test/java/retrofit2/RequestBuilderTest.java
JakeWharton: Added

retrofit/src/main/java/retrofit2/http/QueryName.java
JakeWharton: Changed to `/friends`

PR #2185, opened 2017-02-02
Request example using static and dynamic headers together

calvernaz:


JakeWharton:
I'm not sure this adds a lot of value. Was it not clear that you can use both? We want to keep the website lean and are specifically looking to _not_ make it comprehensive.

calvernaz:
No not really, even after a few searches there is some examples out there with string interpolation to achieve the same result, when I'm not sure it ever worked even if it worked is not really necessary. Lean I can understand, but not comprehensive ? Anyway it's your call.

JakeWharton:
Comprehensive is impossible. Everyone tends to eventually have something they think should be added to the website. We can't target everyone with it because there's simply so many configurations. We're trying to keep the website more like a marketing site rather than a long set documentation.

NightlyNexus:
If it's something that took some time to realize, I'd encourage you to make a StackOverflow question that you [answer yourself](http://stackoverflow.com/help/self-answer).
SO is very searchable without being too cluttered, and you have a good chance of helping others who run into the same question.

JakeWharton:
I think that's a great idea. Especially because we'd need a section about mixing static and dynamic query params as well.

Issue #2171, opened 2017-01-30
Use Moshi's JsonAdapter.serializeNulls() method.


JakeWharton:
Once it's released.

https://github.com/square/moshi/pull/238

PR #2188, opened 2017-02-05
Use Moshi 1.4's JsonAdapter.serializeNulls()

NightlyNexus:
Closes https://github.com/square/retrofit/issues/2171


retrofit-converters/moshi/src/main/java/retrofit2/converter/moshi/MoshiConverterFactory.java
```diff
@@ -94,6 +94,9 @@ public MoshiConverterFactory withNullSerialization() {
     if (failOnUnknown) {
       adapter = adapter.failOnUnknown();
     }
+    if (serializeNulls) {
+      adapter = adapter.serializeNulls();
```
NightlyNexus: Useless, but best for symmetry here, I believe.

JakeWharton:
I figured you were going to send this! Thanks.

PR #2193, opened 2017-02-10
Explain which types are supported for each adapter

Jawnnypoo:
I think it would be much more useful to tell people which types are supported for each adapter. Otherwise they would have to dig into the source to find out, which isn't everyones cup o tea. 

The only one I was not positive on was `adapter-rxjava`, please let me know if I got em wrong.


JakeWharton:
This is great. Thanks!

PR #2195, opened 2017-02-11
Unify HttpException

Jawnnypoo:
Per https://github.com/square/retrofit/issues/2194

PR #2196, opened 2017-02-13
Unify HttpException into a single shared type.

### LINKED ISSUE #2194: Promote HttpException to core artifact.
Change adapter implementations to be deprecated but extend it.

JakeWharton:
Closes #2195. Closes #2194.


retrofit/src/main/java/retrofit2/HttpException.java
```diff
@@ -0,0 +1,52 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package retrofit2;
+
+/** Exception for an unexpected, non-2xx HTTP response. */
+public class HttpException extends Exception {
```
NightlyNexus: I suppose this could have been final, and the deprecated types could have delegated to this.
probably not worth those extra lines, though.

retrofit/src/main/java/retrofit2/HttpException.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright (C) 2016 Square, Inc.
```
swankjesse: 2017 ?

retrofit/src/main/java/retrofit2/HttpException.java
swankjesse: Probably can’t compose because `catch` clauses are expecting a specific type.

swankjesse:
LGTM

retrofit/src/main/java/retrofit2/HttpException.java
JakeWharton: Not actually a new type. Copied from one of the adapters.

PR #2205, opened 2017-02-18
Removal of Unnecessary Creation of String Builder

guptasourabh:
if typeArguments has length 0 there is no point to create stringBuilder for that. that condition check needs to happen before stringBuilder creation.


guptasourabh:
@JakeWharton @swankjesse for review

JakeWharton:
Thanks

PR #2216, opened 2017-02-24
code corrections

guptasourabh:
Moved duplicate code at one place.
If condition needs to be in else if block.
Added errorMessaging in case of NPE exception

### DIFF
--- a/retrofit-mock/src/main/java/retrofit2/mock/NetworkBehavior.java
+++ b/retrofit-mock/src/main/java/retrofit2/mock/NetworkBehavior.java
@@ -19,7 +19,6 @@ import java.io.IOException;
 import java.util.Random;
 import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
-
 import okhttp3.ResponseBody;
 import retrofit2.Response;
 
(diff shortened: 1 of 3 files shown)


retrofit/src/main/java/retrofit2/OkHttpCall.java
```diff
@@ -200,9 +200,7 @@ private void callSuccess(Response<T> response) {
       } finally {
         rawBody.close();
       }
-    }
-
-    if (code == 204 || code == 205) {
+    } else if (code == 204 || code == 205) {
```
JakeWharton: `else` is not needed after a `return`. Please revert.

retrofit-mock/src/main/java/retrofit2/mock/NetworkBehavior.java
```diff
@@ -206,4 +200,10 @@ public long calculateDelay(TimeUnit unit) {
     long callDelayMs = (long) (delayMs * delayPercent);
     return MILLISECONDS.convert(callDelayMs, unit);
   }
+
+  private void checkPercentageValidity(int percentage, String message) {
```
JakeWharton: +`static`

guptasourabh:
@JakeWharton changes done

JakeWharton:
Thanks!

PR #2217, opened 2017-02-24
Removal of unnecessary http method parsing

guptasourabh:
If httpMethod is null then we don't need to parse and set the other values.


JakeWharton:
Your PR contains changes that were already merged. Please rebase and remove the superfluous commits.

PR #2218, opened 2017-02-24
removal of unnecssary parsing if httpMethod is null

guptasourabh:
If httpMethod is null then we don't need to parse and set the other values

### DIFF
--- a/retrofit/src/main/java/retrofit2/ServiceMethod.java
+++ b/retrofit/src/main/java/retrofit2/ServiceMethod.java
@@ -282,6 +282,8 @@ final class ServiceMethod<R, T> {
         throw methodError("Only one HTTP method is allowed. Found: %s and %s.",
             this.httpMethod, httpMethod);
       }
+      if (null == httpMethod) return;
+
       this.httpMethod = httpMethod;
       this.hasBody = hasBody;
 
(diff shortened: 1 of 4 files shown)


retrofit/src/main/java/retrofit2/ServiceMethod.java
```diff
@@ -282,6 +282,8 @@ private void parseHttpMethodAndPath(String httpMethod, String value, boolean has
         throw methodError("Only one HTTP method is allowed. Found: %s and %s.",
             this.httpMethod, httpMethod);
       }
+      if (null == httpMethod) return;
```
JakeWharton: This will never be true.

guptasourabh: if we go through the code in the ServiceMethod build after parseMethodAnnotations , we are checking if httpMethod is null or not and then throwing the exception , so if we check at parseHttpMethodPath itself , we don't need to do value parsing.

JakeWharton:
The exception checks that the method isn't called twice. The value of
httpMethod is guaranteed to never be null by Java since it comes from an
annotation.

guptasourabh:
@JakeWharton requested changes done

guptasourabh:
Moved it into new PR https://github.com/square/retrofit/pull/2219

PR #2226, opened 2017-02-28
moved validation first before cloning

guptasourabh:
First Validate the TypeArguments and then do the cloning of that.


guptasourabh:
@JakeWharton please do the review

jaredsburrows:
LGTM. That's a good idea.

guptasourabh:
@JakeWharton please check the PR.

JakeWharton:
On vacation

JakeWharton:
Rarely is there value in optimizing a code path that results in an exception because the overhead of creating and throwing the exception will dwarf any small change like this.

PR #2243, opened 2017-03-08
Allowing skipping of query params if the conveters converted to a null

devshorts:
First let me say, love retrofit. It's a wonderfully written library.  

We're using retrofit with a custom adapter to use with scala.  So far things work swimmingly, we have our own jackson converter factory that lets us use scala types for request and response bodies. However, we ran into an issue with optional query parameters using the scala None object.  There doesn't seem to be a way to properly indicate "do not use this query param" other than passing in a null object. While this works in java land, its very anti scala.  

Given that parameters can be converted, this PR submits that if a string converter returns a valid converter who _then_ returns a null object, that this indicates that the object should not be used as a query/header/etc param.  This lets us hook into the conversion factory and delegate all type resolution of scala wonkery to the converter.

### DIFF
--- a/retrofit/src/test/java/retrofit2/CallTest.java
+++ b/retrofit/src/test/java/retrofit2/CallTest.java
@@ -18,7 +18,6 @@ package retrofit2;
 import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
-import java.util.Map;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
@@ -40,8 +39,6 @@ import retrofit2.http.Body;
 import retrofit2.http.GET;
 import retrofit2.http.POST;
 import retrofit2.http.Path;
-import retrofit2.http.Query;
-import retrofit2.http.QueryMap;
 import retrofit2.http.Streaming;
 
 import static java.util.concurrent.TimeUnit.SECONDS;
--- a/retrofit/src/test/java/retrofit2/CallTest.java
+++ b/retrofit/src/test/java/retrofit2/CallTest.java
@@ -18,7 +18,6 @@ package retrofit2;
 import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -36,7 +35,6 @@ import okio.ForwardingSource;
 import okio.Okio;
 import org.junit.Rule;
 import org.junit.Test;
-import retrofit2.helpers.NullObjectConverterFactory;
 import retrofit2.helpers.ToStringConverterFactory;
 import retrofit2.http.Body;
 import retrofit2.http.GET;
@@ -63,8 +61,6 @@ public final class CallTest {
     @GET("/") @Streaming Call<ResponseBody> getStreamingBody();
     @POST("/") Call<String> postString(@Body String body);
     @POST("/{a}") Call<String> postRequestBody(@Path("a") Object a);
-    @POST("/query") Call<String> queryPath(@Query("a") Object a);
-    @POST("/query") Call<String> queryPathMap(@QueryMap Map<String, String> a);
   }
 
   @Test public void http200Sync() throws IOException {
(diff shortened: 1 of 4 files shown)


retrofit/src/test/java/retrofit2/CallTest.java
```diff
@@ -978,4 +985,66 @@
     }
     assertThat(writeCount.get()).isEqualTo(1);
   }
+
+  @Test public void queryParamsSkippedIfConvertedToNull() throws IOException, InterruptedException {
```
NightlyNexus: Do these tests belong in `RequestBuilderTest`? They just need to ensure that the request is created correctly. I don't think they care about execution of calls.

devshorts: Yeah, I wasn't sure where to put them. Open to suggestions and I'll make that change

devshorts:
Hmm, I'm not sure I see how is a breaking change, as if you return null right now you just get a null pointer error as the query builders try to use the value. If someone wants to legitimately query with the word null, they'd be returning it as a string right?

NightlyNexus:
Yeah, you might be right about compatibility. 🆒 ❗️

devshorts:
@NightlyNexus I moved the tests to request builder and applied your PR feedback. Is there any way I could get another review?

retrofit/src/test/java/retrofit2/RequestBuilderTest.java
```diff
@@ -2578,6 +2579,44 @@
...
+
+    final HashMap<String, String> queryMap = new HashMap<String, String>() {{
+      put("Ignored", "Always Null");
+    }};
+
+    try {
+      buildRequest(Example.class, retrofitBuilder, queryMap);
+      fail();
+    } catch (IllegalArgumentException e) {
+      assertThat(e).hasMessageContaining("Query map contained null value for key");
```
NightlyNexus: nit: hasMessage(...) and have the full message.

retrofit/src/test/java/retrofit2/CallTest.java
```diff
@@ -39,6 +40,8 @@
 import retrofit2.http.GET;
 import retrofit2.http.POST;
 import retrofit2.http.Path;
+import retrofit2.http.Query;
+import retrofit2.http.QueryMap;
```
NightlyNexus: Revert unused imports.

retrofit/src/test/java/retrofit2/RequestBuilderTest.java
```diff
@@ -2578,6 +2579,44 @@
     assertThat(readBody.indexOf("secondParam")).isLessThan(readBody.indexOf("thirdParam"));
   }
 
+
+  @Test public void queryParamsSkippedIfConvertedToNull() throws IOException, InterruptedException {
```
NightlyNexus: you can always just to `throws Exception` to save your fingers some typing. :)

retrofit/src/test/java/retrofit2/RequestBuilderTest.java
```diff
@@ -2578,6 +2579,44 @@
...
+    Retrofit.Builder retrofitBuilder = new Retrofit.Builder()
+            .baseUrl("http://example.com")
+            .addConverterFactory(new NullObjectConverterFactory());
+
+    Request request = buildRequest(Example.class, retrofitBuilder, "Ignored");
+
+    if(request.url().toString().contains("Ignored")){
+      fail();
+    }
+  }
```
NightlyNexus: I _think_ tests tend to have some assertion that succeeds to pass. Maybe assert that request.url().queryParameter("Ignored") is null?

retrofit/src/test/java/retrofit2/RequestBuilderTest.java
```diff
@@ -2578,6 +2579,44 @@
...
+
+  @Test public void queryParamMapsConvertedToNullShouldError() throws IOException, InterruptedException {
```
NightlyNexus: 👍

retrofit/src/test/java/retrofit2/RequestBuilderTest.java
```diff
@@ -2578,6 +2579,44 @@
...
+    class Example {
+      @POST("/query") Call<String> queryPath(@QueryMap HashMap<String, String> a) { return null; }
+    }
+
+    Retrofit.Builder retrofitBuilder = new Retrofit.Builder()
+            .baseUrl("http://example.com")
+            .addConverterFactory(new NullObjectConverterFactory());
+
+    final HashMap<String, String> queryMap = new HashMap<String, String>() {{
```
NightlyNexus: just btw, [always prefer LinkedHashMap](https://publicobject.com/2016/02/08/linkedhashmap-is-always-better-than-hashmap/).

devshorts:
@NightlyNexus appreciate it anyways. Is @JakeWharton the main maintainer? Who is the primary reviewer?

JakeWharton:
I'm on vacation. Not reviewing.

retrofit/src/test/java/retrofit2/RequestBuilderTest.java
```diff
@@ -2598,16 +2637,12 @@ private static void assertBody(RequestBody body, String expected) {
...
-        .baseUrl("http://example.com/")
-        .addConverterFactory(new ToStringConverterFactory())
-        .callFactory(callFactory)
-        .build();
+    Retrofit retrofit = builder.callFactory(callFactory).build();
 
     Method method = TestingUtils.onlyMethod(cls);
     //noinspection unchecked
     ServiceMethod<T, Call<T>> serviceMethod =
-        (ServiceMethod<T, Call<T>>) retrofit.loadServiceMethod(method);
```
salarsk: good

salarsk: 20

JakeWharton:
Yeah this looks good. I'm going to make some style changes locally and then merge.

PR #2258, opened 2017-03-26
Fix completable type checking

devholic:
`Completable` in `adapter-rxjava` is not working with proguard, so I fixed it.


JakeWharton:
Completable is not stable API and thus cannot be relied on the be present. Exclude it from obfuscation.

PR #2286, opened 2017-04-18
Fix NoSuchElementException when using Rx 1.x Singles.

tonycosentini:
We're seeing quite a few `NoSuchElementException` errors in our apps and I think I've narrowed it down to this issue.

It looks like `CallArbiter` checks if the subscriber is still subscribed when calling onNext, but not when calling onCompleted. This can lead to `NoSuchElementException` when a subscriber unsubscribes during a `deliverResponse` call.

**Repro steps:**
I can't 100% reproduce this in the wild, but I can trigger it via the debugger. If you run this [snippet](https://gist.github.com/tonycosentini/d22f05431af34123ef725a5d4cf08e59), add a breakpoint to `deliverResponse`, and manually unsubscribe via the debugger, it will crash.

There is also a test in `AsyncTest` that will reproduce this by failing when the fix is removed.

**TODO:**
* [x] Tests? If this is a valid bug, some guidance on where / how to test this would be super appreciated.

### DIFF
--- a/retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/CallArbiter.java
+++ b/retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/CallArbiter.java
@@ -140,9 +140,7 @@ final class CallArbiter<T> extends AtomicInteger implements Subscription, Produc
 
     if (!isUnsubscribed()) {
       try {
-        if (!isUnsubscribed()) {
-          subscriber.onError(t);
-        }
+        subscriber.onError(t);
       } catch (Throwable inner) {
         Exceptions.throwIfFatal(inner);
         CompositeException composite = new CompositeException(t, inner);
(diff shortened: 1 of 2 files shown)


naturalwarren:
This could also happen with RxJava 2 right?

tonycosentini:
> This could also happen with RxJava 2 right?

It can happen with an Rx 1.x Single interopped to an Rx 2.x Single.

Fo the native Rx 2.x adapter, it looks like it does check before calling onComplete.

retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/CallArbiter.java
```diff
@@ -138,7 +140,9 @@ void emitError(Throwable t) {
 
     if (!isUnsubscribed()) {
       try {
-        subscriber.onError(t);
+        if (!isUnsubscribed()) {
```
hzsweers: This seems redundant? It's checked two lines above

tonycosentini: oops - good point

tonycosentini:
@hzsweers is your rejection stale now? Diff should be good now.

Any Square folks able to review this? There is a test now + we rolled out a fork rolling out to confirm the fix.

hzsweers:
LGTM

JakeWharton:
Merged manually with some test tweaks as a5357cb72309e4e32b2cbee2463dd1894a880ede

hzsweers:
Thanks!

tonycosentini:
Thank you!

Issue #624, opened 2014-10-07
Evaluate a nullable static analysis library


JakeWharton:
In general, the disparate nature of these tools depress me. It would be nice to have something assisting with this, however.

JakeWharton:
Can't get on board with any, sadly. Too much fragmentation in this area.

joshfriend:
I've tried to guess why each of the [supported nullablilty annotations supported by Kotlin](https://github.com/JetBrains/kotlin/blob/1.0.1/core/descriptor.loader.java/src/org/jetbrains/kotlin/load/java/JvmAnnotationNames.kt#L21-L42) wont work:

* The AOSP/Android support annotations are obviously out since Retrofit is not an Android-only project
* JSR305 is marked as "Dormant" (read: dead?)
* IDE Specific annotations (`org.eclipse.annotations`) not available everywhere
* CheckerFramework/FindBugs are relatively heavy dependencies for what you would use them for

`org.jetbrains.annotations` however consists solely of 4 annotations (`Mutable`, `Nullable`, `NonNull`, and `ReadOnly`) and is available as a self contained dependency (#604 did not explain why they were not usable either)

Could you explain what you found when you investigated this which prevents using Kotlin's `org.jetbrains` annotations?

JakeWharton:
You nailed most of the reasoning, yep!

The JetBrains ones are okay. We've actually used them in the past. Now that they're [fixing their mistake][1] of bundling them inside Kotlin instead of just depending on them they're even more appealing.

I'll bring this up with @swankjesse and see if we can't decide on one because it would likely apply to all our Java projects, not just this one.

 [1]: https://discuss.kotlinlang.org/t/org-jetbrains-annotations-notnull-problem-with-android-build/629/6

swankjesse:
I think I’m ready for this. I generally prefer JSR 305 but not well-motivated. JetBrains’ annotations seem reasonable.

Can we assume everything not-annotated is non-null? I’d like to make non-null our default if the tools will honor that.

vanniktech:
Then you might want to use [@ParametersAreNonnullByDefault](https://www.jetbrains.com/help/idea/2016.3/parametersarenonnullbydefault-annotation.html)

joshfriend:
@vanniktech kotlin does not seem to support `@ParametersAreNonnullByDefault` at the moment, as noted in a similar issue for OkHttp: https://github.com/square/okhttp/pull/2969#issuecomment-259974936

[...]

PR #2290, opened 2017-04-20
[WIP] Add nullability annotations to public APIs

joshfriend:
WIP implementation of #624

What I think I've covered:

* [x] Retrofit public APIs
* [x] Retrofit Adapters
* [ ] Retrofit Converters

It also turns out that `org.jetbrains:annotations-java5` contains _only_ `@Nullable` and `@NotNull`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -51,6 +51,7 @@
     <android.version>4.1.1.4</android.version>
     <okhttp.version>3.6.0</okhttp.version>
     <animal.sniffer.version>1.14</animal.sniffer.version>
+    <annotations.version>15.0</annotations.version>
 
     <!-- Adapter Dependencies -->
     <rxjava.version>1.2.0</rxjava.version>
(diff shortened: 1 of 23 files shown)


joshfriend:
moshi/pull#297 used jsr305 annotations, I assume retrofit should use the same?

JakeWharton:
Yep. JSR 305 was added to all our libraries. The PR for Retrofit was here: #2315.

Sorry I didn't provide feedback sooner. I meant to update this but it fell off my radar. We spent some time playing with the different libraries internally before deciding on which one to use. I appreciate you taking the time to send the PR for this though. It helped kickstart the effort.

joshfriend:
Thats fine, I'm just happy that it's happening! I wasn't exactly pleased with how littered the code looked with `@Nonnull` everywhere, I think `@ParametersAreNonnullByDefault` is a much better solution. Hopefully with you and @swankjesse weighing in on the [Kotlin issue](https://youtrack.jetbrains.com/issue/KT-10942) it will get some traction.

PR #2299, opened 2017-04-21
Fix annotation docs to advertise string converters.

NightlyNexus:


retrofit/src/main/java/retrofit2/http/Field.java
```diff
@@ -15,17 +15,22 @@
...
+import retrofit2.Retrofit;
 
 import static java.lang.annotation.ElementType.PARAMETER;
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
 
 /**
  * Named pair for a form-encoded request.
  * <p>
- * Values are converted to strings using {@link String#valueOf(Object)} and then form URL encoded.
+ * Values are converted to strings using {@link Retrofit#stringConverter(Type, Annotation[])}
```
digitalbuddha: converted to ~~strings~~Strings?

retrofit/src/main/java/retrofit2/http/Field.java
```diff
@@ -15,17 +15,22 @@
...
+ * (or {@link Object#toString()}, if no matching string Converter is installed)
```
digitalbuddha: same, `String` should be class case?

JakeWharton: I'd prefer if "Converter" was lowercased instead. Saying "string converter" is fine.

JakeWharton:
Thanks!

PR #2303, opened 2017-04-29
Allow Crawler to crawl urls that aren't present in the fetched urls set

lucastsantos:
The subsequent calls of `crawlPage` was only occurring if the url that is being iterated is already in the fetched urls set.

With this modification, `crawlPage`  only occurs in the first time that the url is being added to the set


JakeWharton:
Thanks!

Issue #2302, opened 2017-04-28
Response from 202 code in to JSON


trinadhkoya:
i want to convert the Response in to JSON and show it to the User .I was able to display 200 response but in 202,i am failed.

Gson gson = new Gson();
SuccessResponse signupResponse = response.body();
String sSignupResponse = gson.toJson(signUpResponse, SuccessResponse.class);
should i do the same for 202 too? i created another POJO like ErrorResponse with message and code ,replaced the SuccessRespone with ErrorResponse

JakeWharton:
If you need to convert 202s differently than 200s you'll have to write a custom call adapter which returns a `responseType()` of `ResponseBody.class` and then does the conversion manually based on the response code. There isn't a good sample of this currently, but I would welcome one.

PR #2307, opened 2017-05-02
Add CallAdapter example for multi-deserialization.

NightlyNexus:
A proposal for #2302.

### DIFF
--- a/samples/src/main/java/com/example/retrofit/MultipleDeserializedTypes.java
+++ b/samples/src/main/java/com/example/retrofit/MultipleDeserializedTypes.java
@@ -1,0 +1,280 @@
+/*
+ * Copyright (C) 2017 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


samples/src/main/java/com/example/retrofit/MultipleDeserializedTypes.java
```diff
@@ -0,0 +1,279 @@
...
+    final class Response<R, T> {
+      /**
+       * Null iff the response is 202.
+       */
+      public final retrofit2.Response<R> response;
+
+      /**
+       * Not null iff the response is 202.
+       */
+      public final retrofit2.Response<T> accepted;
```
NightlyNexus: gross to reuse this type for both responses. unsure if it's worth having a marker interface or something else.

samples/src/main/java/com/example/retrofit/MultipleDeserializedTypes.java
```diff
@@ -0,0 +1,279 @@
...
+
+    boolean isExecuted();
+
+    void cancel();
+
+    boolean isCanceled();
+
+    Call<R, T> clone();
+
+    Request request();
```
NightlyNexus: unsure if these methods should be snipped from the example to tighten its focus.

Issue #2123, opened 2016-12-12
Add sample showing polymorphic converter.


karlkar:
So that the runtime type is serialized and not the declared type.

*Original issue below*

----

I have a very strange problem with Retrofit, in fact it may be a bug of it. I want to send to server a POST request with some data. My data is in a subclass of Request class.

```
public abstract class Request {}
public abstract class Measurement {}

public class StepMeasurement extends Measurement {
    ...
    public int value;
}
public class SendStepsRequest extends Request {
    public List<StepMeasurement> steps;
}
````

example is of course very simplified. I've prepared an interface for sending it:
```

public interface EndpointInterface {

@Headers({"Content-Type: application/json"})
@POST("/patients/{id}/{dataType}")
Call<ResponseBody> postRequest(
       @Path("id") int patientId,
        @Path("dataType") String dataType,
        @Header("X-XSRF-TOKEN") String token,
        @Header("Cookie") String cookie,
        @Body Request req);
}
```

And I have a retrofit ready to go:

```
mEndpointInterface = new Retrofit.Builder()
            .baseUrl(BASE_URL)
            .client(getUnsafeOkHttpClient())
            .addConverterFactory(GsonConverterFactory.create())
            .build().create(EndpointInterface.class);
```

I think that now calling mEndpointInterface.postRequest(1, "steps", "token", "cookie", new SendStepsRequest(5)); should send a JSON looking like that:

> {"steps":[{"value":5}]}

However all that is sent is

> {}

For me unlogical. I guess that Gson doesn't know how to go down in a class hierarchy, so he doesn't know that SendStepsRequest is a subtype of Request. I've already prepared RuntimeTypeAdapterFactory and it started to work then, but that's strange for me. Shouldn't GSON know about inheritance? So I started to experiment.

And now some interesting stuff. I've created a class RequestWrapper:

```
public class RequestWrapper {
    Request mRequest;
}
```

and changed my interface to accept RequestWrapper instead of Request and guess what! The result was:

> {"mRequest":{"steps":[{"value":5}]}}

so he found out how my classes are hierarchized, but couldn't do it without RequestWrapper.

I've also tried a different thing - I've changed my Interface to accept SendStepsRequest objects and changed List< StepMeasurement > to List< Measurement >. What Retrofit send to the server?

> {"steps":[{"value":5}]}

so again, he detected the subtype of Measurement class and used it as he should.

[...]

PR #2310, opened 2017-05-03
Add polymorphic converter example.

NightlyNexus:
A proposal for #2123.

### DIFF
--- a/samples/src/main/java/com/example/retrofit/PolymorphicConverter.java
+++ b/samples/src/main/java/com/example/retrofit/PolymorphicConverter.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright (C) 2017 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.example.retrofit;
+
+import java.io.IOException;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import java.util.LinkedHashMap;
+import java.util.Map;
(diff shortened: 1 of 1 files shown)


samples/src/main/java/com/example/retrofit/PolymorphicConverter.java
```diff
@@ -0,0 +1,88 @@
...
+import okhttp3.RequestBody;
+import okio.Buffer;
+import retrofit2.Call;
+import retrofit2.Converter;
+import retrofit2.Retrofit;
+import retrofit2.converter.moshi.MoshiConverterFactory;
+import retrofit2.http.Body;
+import retrofit2.http.POST;
+
+final class PolymorphicConverter {
```
JakeWharton: I would like documentation on this describing why you would want something like this and also warning of the performance implications of doing so.

samples/src/main/java/com/example/retrofit/PolymorphicConverter.java
```diff
@@ -0,0 +1,108 @@
...
+import retrofit2.Call;
+import retrofit2.Converter;
+import retrofit2.Retrofit;
+import retrofit2.converter.moshi.MoshiConverterFactory;
+import retrofit2.http.Body;
+import retrofit2.http.POST;
+
+/**
+ * A converter for polymorphic request bodies. This may be useful for libraries and places where the
+ * runtime types are unknown.
```
NightlyNexus: honestly, i dunno if anybody has a legitimate use case for this, though...

samples/src/main/java/com/example/retrofit/PolymorphicConverter.java
```diff
@@ -0,0 +1,108 @@
...
+ * For each service method, Retrofit caches converters for request bodies.
+ * The following requires loading from another cache for each invocation of the service method.
+ */
+final class PolymorphicConverter {
+  public static final class PolymorphicRequestBodyConverter<T>
+      implements Converter<T, RequestBody> {
+    public static final Factory FACTORY = new Factory() {
+      @Override public Converter<?, RequestBody> requestBodyConverter(Type type,
+          Annotation[] parameterAnnotations, Annotation[] methodAnnotations, Retrofit retrofit) {
+        return new PolymorphicRequestBodyConverter<>(this, parameterAnnotations, methodAnnotations,
```
NightlyNexus: could have done something around `if (getRawType(type) != Animal.class) return null` to discourage using this everywhere, but that's just on CallAdapter.Factory.

PR #2315, opened 2017-05-08
Add JSR 305 nullability annotations.

JakeWharton:


retrofit/src/main/java/retrofit2/RequestBuilder.java
```diff
@@ -71,7 +71,6 @@
   }
 
   void setRelativeUrl(Object relativeUrl) {
-    checkNotNull(relativeUrl, "@Url parameter is null.");
```
JakeWharton: Moved to call site

pom.xml
```diff
@@ -156,7 +156,11 @@
         <artifactId>moshi</artifactId>
         <version>${moshi.version}</version>
       </dependency>
-
+      <dependency>
+        <groupId>com.google.code.findbugs</groupId>
+        <artifactId>jsr305</artifactId>
+        <version>3.0.2</version>
```
swankjesse: I've been putting `provided` at this level. Unnecessary?

retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/Result.java
```diff
@@ -16,24 +16,27 @@
...
 
 /** The result of executing an HTTP request. */
 public final class Result<T> {
+  @SuppressWarnings("ConstantConditions") // Guarding public API nullability.
   public static <T> Result<T> error(Throwable error) {
     if (error == null) throw new NullPointerException("error == null");
     return new Result<>(null, error);
   }
 
+  @SuppressWarnings("ConstantConditions") // Guarding public API nullability.
```
swankjesse: I didn't bother with these in OkHttp etc.
Should do.

swankjesse: Or we could ask IntelliJ to special case this kind of check.

retrofit/src/main/java/retrofit2/OkHttpCall.java
```diff
@@ -28,16 +30,18 @@
...
   private final ServiceMethod<T, ?> serviceMethod;
-  private final Object[] args;
+  private final @Nullable Object[] args;
 
   private volatile boolean canceled;
 
-  // All guarded by this.
-  private okhttp3.Call rawCall;
-  private Throwable creationFailure; // Either a RuntimeException or IOException.
+  @GuardedBy("this")
```
swankjesse: Neat

pom.xml
JakeWharton: Putting it here forces all submodule consumers to use it as provided. The same would be true if you put 'test' then only submodules could use it as a test artifact. As a rule I generally put only the dependency pointer in the root and then let the submodules decide how they want to actually use it.

But it really doesn't matter...

Jawnnypoo:
Nice stuff! Curious what convinced you guys to settle on JSR305? 

Also, I don't know if you guys have already considered this, but this will be a potentially breaking change for those using Kotlin + Retrofit, as it will change the signature of calls from `Any!` to `Any?` right?

JakeWharton:
Yep. The whole point is to break consumers in a source-compatible way such
that they handle these cases properly.

joshfriend:
@Jawnnypoo @swankjesse offered an explanation of the JSR305 choice over on okio: https://github.com/square/okio/pull/307#issuecomment-298484987

Jawnnypoo:
@JakeWharton Good, good, just wanted to make sure it was on the radar. I have changed projects from no annotation to `@Nullable` annotated and was surprised by the compilation errors when upgrading a Kotlin project. I am sure something in the release notes will help others find out what is going on. 

@joshfriend Thanks for pointing that out!

PR #2322, opened 2017-05-11
Explicitly handle wrapped, re-thrown exceptions from onComplete, onError.

### LINKED ISSUE #2285: Retrofit + RxJava: Exception in onError crashes without stack trace
If exception is thrown in onError  (not intentionally) app crashes but stack trace is missing in logcat.

```
githubService.contributors("square", "retrofit")
        .subscribeOn(Schedulers.io())
        .observeOn(AndroidSchedulers.mainThread())
        .subscribe(new Subscriber<List<SimpleService.Contributor>>() {
            @Override
            public void onCompleted() {
            }

            @Override
            public void onError(Throwable e) {
                // Force java.lang.IndexOutOfBoundsException
                // App will crash, exception will be swollen (not visible in Logcat)
                int[] ints = new int[2];
                int intValue = ints[10];
            }

            @Override
            public void onNext(List<SimpleService.Contributor> contributors) {
            }
        });
```
While the same exception in plain RxJava leaves a stack trace in logcat.
```
Observable.error(new RuntimeException())
        .subscribeOn(Schedulers.io())
        .observeOn(AndroidSchedulers.mainThread())
        .subscribe(new Subscriber<Object>() {
            @Override
            public void onCompleted() {
            }

            @Override
            public void onError(Throwable throwable) {
                // Force java.lang.IndexOutOfBoundsException
                // App will crash, exception is visible in Logcat
                int[] ints = new int[2];
                int intValue = ints[10];
            }

            @Override
            public void onNext(Object o) {
            }
        });
```
Full examples available [here](https://github.com/AndrazP/onErrorExceptionSample). I'm not sure if this is a Retrofit issue or RxJava/RxAndroid, but would be thankful if you can help me out. It's nearly impossible to debug such unintentional exceptions without stack trace.

JakeWharton:
Because `Exceptions.throwIfFatal` treats these as fatal, prior to this they would bubble up to the caller. This is either the thread uncaught exception handler which is fine, but in the case of async sources it would go to the library code invoking the async callback (like OkHttp's `Call`) which might swallow them.

Tests are just copy/paste from RxJava 2 adapter.

Closes #2285.

### DIFF
--- a/retrofit-adapters/rxjava/src/test/java/retrofit2/adapter/rxjava/CompletableThrowingSafeSubscriberTest.java
+++ b/retrofit-adapters/rxjava/src/test/java/retrofit2/adapter/rxjava/CompletableThrowingSafeSubscriberTest.java
@@ -1,0 +1,129 @@
+/*
+ * Copyright (C) 2016 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package retrofit2.adapter.rxjava;
+
+import java.util.concurrent.atomic.AtomicReference;
+import okhttp3.mockwebserver.MockResponse;
+import okhttp3.mockwebserver.MockWebServer;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TestRule;
+import retrofit2.Retrofit;
+import retrofit2.http.GET;
+import rx.Completable;
+import rx.CompletableSubscriber;
+import rx.Subscription;
+import rx.exceptions.CompositeException;
+import rx.exceptions.Exceptions;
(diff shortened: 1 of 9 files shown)


retrofit-adapters/rxjava/src/test/java/retrofit2/adapter/rxjava/CompletableThrowingSafeSubscriberTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+  @Test public void throwingInOnCompleteDeliveredToPlugin() {
+    server.enqueue(new MockResponse());
+
+    final AtomicReference<Throwable> pluginRef = new AtomicReference<>();
+    RxJavaPlugins.getInstance().registerErrorHandler(new RxJavaErrorHandler() {
+      @Override public void handleError(Throwable throwable) {
+        if (throwable instanceof OnCompletedFailedException) {
+          if (!pluginRef.compareAndSet(null, throwable)) {
+            throw Exceptions.propagate(throwable); // Don't swallow secondary errors!
```
swankjesse: neat

retrofit-adapters/rxjava/src/test/java/retrofit2/adapter/rxjava/SingleThrowingTest.java
```diff
@@ -0,0 +1,234 @@
...
+    service.result().subscribe(new ForwardingObserver<Result<String>>(observer) {
+      @Override public void onSuccess(Result<String> value) {
+        throw e;
+      }
+    });
+
+    assertThat(pluginRef.get()).isSameAs(e);
+  }
+
+  @Ignore("Single's contract is onNext|onError so we have no way of triggering this case")
```
swankjesse: tricky

PR #2323, opened 2017-05-11
Allowing skipping of params if the converters converted to a null

JakeWharton:
Closes #2243.


retrofit/src/test/java/retrofit2/RequestBuilderTest.java
```diff
@@ -2578,6 +2579,84 @@
     assertThat(readBody.indexOf("secondParam")).isLessThan(readBody.indexOf("thirdParam"));
   }
 
+  @Test public void queryParamsSkippedIfConvertedToNull() throws Exception {
+    class Example {
+      @GET("/query") Call<ResponseBody> queryPath(@Query("a") Object a) {
+        return null;
```
NightlyNexus: these could probably throw AssertionErrors, but all the other tests return null, too, so 👍 consistency.

Issue #2242, opened 2017-03-07
RxJava 2 Adapter: null may slip into RxJava


akarnokd:
There was [a question on Stackoverflow](http://stackoverflow.com/questions/42624420/retrofit-2-with-rxjava2-and-plain-null-responses-error) recently which reported an unexpected `NullPointerException` caught by the one of the internal safeguards of RxJava 2.

Looks like `retrofit2.Response.body()` may return `null` which is handed over to RxJava 2 without validation in [BodyObservable](https://github.com/square/retrofit/blob/master/retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/BodyObservable.java#L51).

```java
@Test
public void responseBodyNull() throws Exception {
    
    Response<Integer> response = Response.success(null);
    
    Assert.assertNull(response.body());
    
    Class<?> clazz = Class.forName("retrofit2.adapter.rxjava2.BodyObservable");
    Constructor<?> c = clazz.getDeclaredConstructor(Observable.class);
    c.setAccessible(true);
    Observable<Integer> o = (Observable)c.newInstance(Observable.just(response));

    o.test()
    .assertNever(r -> r == null)
    .assertTerminated();
}
```

I don't know what the reasonably response to a `null` body could be:

  - complete the `Observable` without value or
  - signal a `NoSuchElementException`.

naturalwarren:
I think it makes sense to complete the Observable. It's not really an error and if you've written a custom converter to return null when you've received an empty body (as JW described in #1554) this seems like the right behavior.

I could help out with a PR here. I've had trouble with backends returning empty response bodies and currently use a delegating converter to return null. This is going to cause troubles now that I'm moving to RxJava 2.

JakeWharton:
I don't think there's any action to take here. A `null` body in the success (HTTP 2xx) case should warrant the use of `Observable<Response<Integer>>` in the Retrofit interface signature. 

Completing the stream without a value is definitely the wrong solution as it means the successful-but-empty response will go unprocessed due to the lack of actionable notification downstream onto which operators can be applied.

This is most strongly exemplified by the `Single<Integer>` case which more accurately represents an HTTP request. `Observable<Integer>` is just a variant of this where `onSuccess` is turned into the pair of `onNext`/`onComplete` calls (despite the implementation details actually being the reverse of this).

So there's three ways to handle this case:
 * Use `Observable<BodyType>` which indicates you expect to never receive `null` in the success case.
[...]

PR #2324, opened 2017-05-11
Add converters which facilitate the use of Optional.

JakeWharton:
Both Java 8 and Guava are supported, just like in the adapters.

This is especially useful in the use of RxJava 2 as it does not allow null values in the stream. This means a return type declaration of `Single<BodyType>` might result in an NPE should a converter deserialize the response body into null. By adding one of these delegating converters before your normal converter, it allows you to use `Single<Optional<BodyType>>` to support this case.

Refs #2242

### DIFF
--- a/retrofit-converters/guava/README.md
+++ b/retrofit-converters/guava/README.md
@@ -1,0 +1,30 @@
+Guava Converter
+===============
+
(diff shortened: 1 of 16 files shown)


swankjesse:
Nice. I love how straightforward this composition is.

PR #2325, opened 2017-05-11
clarify random network failure message

dtanner:
Small change, but this one bit us when someone copy/pasted an example usage into our project and we weren't aware of the new behavior.  Having a less ambiguous error message will help track down the problem more quickly IMO.

### DIFF
--- a/retrofit-mock/src/main/java/retrofit2/mock/NetworkBehavior.java
+++ b/retrofit-mock/src/main/java/retrofit2/mock/NetworkBehavior.java
@@ -76,7 +76,7 @@ public final class NetworkBehavior {
   private NetworkBehavior(Random random) {
     this.random = random;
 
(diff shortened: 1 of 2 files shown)


JakeWharton:
Would a custom exception type help more? Like `MockRetrofitIOException`?

dtanner:
@JakeWharton - yeah that's a more strictly typed solution, although unnecessary in our situation because I don't see a reason to code for that specific type.  The verbiage change would have been enough for us as we scanned the stack trace to indicate it was an intentional error, and not some unknown error.

I don't have a strong opinion either way, as long as one of those is done.  If you'd like I can issue a new PR with a new type like you suggested.

JakeWharton:
I think I'd prefer a custom exception type because otherwise the lack of a trace prevents you from figuring out where this comes from. I don't really see the exception message change as clarifying much.

PR #2327, opened 2017-05-12
Use custom exception type when triggering mock failure.

JakeWharton:
Since we omit the stack trace, this more clearly indicates the source being from Retrofit's mock behavior.

Closes #2325.

### DIFF
--- a/retrofit-mock/src/main/java/retrofit2/mock/MockRetrofitIOException.java
+++ b/retrofit-mock/src/main/java/retrofit2/mock/MockRetrofitIOException.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright (C) 2017 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


retrofit-mock/src/main/java/retrofit2/mock/MockRetrofitIOException.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package retrofit2.mock;
+
+import java.io.IOException;
+
+final class MockRetrofitIOException extends IOException {
+  MockRetrofitIOException() {
+    super("Failure triggered by MockRetrofit's NetworkBehavior");
```
NightlyNexus: Period?

JakeWharton: We don't have a policy around this. I think we only tend to put periods where there is more than one sentence in the message.

NightlyNexus: 👍 glad I ask the silly questions. good to know, thx!

dtanner:
Thanks @JakeWharton !

PR #2330, opened 2017-05-13
updating copyright notices

tristansokol:
I saw that the copyright on on the [html page](http://square.github.io/retrofit/) was from 2013, and that seemed wrong. Then I realized that there were a _ton_ of copyright notices that were from all over the place. 

I spoke with some friends at legal@ and it seems like a decent strategy around keeping the copyright dates accurate is to include both the original date, as well as the most recent date that the files were updated. So I wrote a bash script to do all that. 

I did some spot checking, and it should avoid things like external libraries pretty well and the like. Let me know what you think!

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Retrofit requires at minimum Java 7 or Android 2.3.
 License
 =======
 
-    Copyright 2013 Square, Inc.
+    Copyright 2013 - 2017 Square, Inc.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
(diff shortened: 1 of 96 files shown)


swankjesse:
I think the cost of being precise here is high, and the utility is low. Just the noise in the version history of every file is worth avoiding.

For example, in projects that do this, the “last modified” date in the GitHub UI is broken. Got a regression and looking for something to blame? Everything is recently updated!

tristansokol:
I can pick up what you are putting down. What about a pre-commit hook?

swankjesse:
It's still pretty invasive. I don't think the utility justifies the cost. Maybe we remove the years altogether and rely on Git to tell us when things changed?

Issue #2335, opened 2017-05-16
Error while generating signed APK


krishnameena:
I have updated retrofit 2.3.0 and okhttp 3.8.0 and now while generating signed APK I am getting following error 

Warning:retrofit2.OkHttpCall: can't find referenced class javax.annotation.concurrent.GuardedBy
Warning:there were 3 unresolved references to classes or interfaces.
Warning:Exception while processing task java.io.IOException: Please correct the above warnings first.
Error:Execution failed for task ':app:transformClassesAndResourcesWithProguardForRelease'.
> Job failed, see logs for details

Following are the proguard rules which I am using for Retrofit and OKHTTP

```
# Platform calls Class.forName on types which do not exist on Android to determine platform.
-dontnote retrofit2.Platform
# Platform used when running on Java 8 VMs. Will not be used at runtime.
-dontwarn retrofit2.Platform$Java8
# Retain generic type information for use by reflection by converters and adapters.
-keepattributes Signature
# Retain declared checked exceptions for use by a Proxy instance.
-keepattributes Exceptions
```

```
-dontwarn org.xmlpull.v1.**
-dontwarn okhttp3.**
-keep class okhttp3.** { *; }
-dontwarn okio.**
-dontwarn javax.annotation.Nullable
-dontwarn javax.annotation.ParametersAreNonnullByDefault
```

NightlyNexus:
I think you just need
`-dontwarn javax.annotation.GuardedBy`

JakeWharton:
or `-dontwarn javax.annotation.**`

JakeWharton:
Also please don't use `-keep class okhttp3.** { *; }`. There should be no need for that.

krishnameena:
@JakeWharton  Thanks it worked

gjsalot:
Someone should update the docs to include `-dontwarn javax.annotation.**`

KhadkaRabindra:
@JakeWharton I am getting follwoing erros after I add -dontwarn javax.annotation.**
`Error:The string "--" is not permitted within comments.
Warning:Configuration 'compile' is obsolete and has been replaced with 'implementation'.
Information:Gradle tasks [:app:assembleDebug]`

JakeWharton:
That link is a horrid recommendation. If you ever feel the urge to do that
you should just be disabling ProGuard. You're using it wrong.

alitele:
For me the fix was just adding this line in proguard file.

`-ignorewarnings`

PR #2336, opened 2017-05-17
Add ProGuard rules

PromanSEW:
These rules were copied from OkHttp README. If there are Retrofit specific settings, please note.
Also see #2335 for additional information


swankjesse:
Thanks!

PR #2338, opened 2017-05-18
Additional rules for ProGuard

### LINKED ISSUE #2129: Enhancement: retrofit proguard rules
For now the progurad rules in http://square.github.io/retrofit/ won't work correctly if you are using `proguard-android-optimize`. 
We need add this to make it work.
```
-keepclasseswithmembers class * {
    @retrofit2.http.* <methods>;
}
```
For more information please see http://stackoverflow.com/questions/35506111/retrofit-2-not-sending-data-when-proguard-is-enabled.

PromanSEW:
See http://square.github.io/retrofit/
Are these rules still actual?
Also fixes #2129

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,15 +30,20 @@ Retrofit requires at minimum Java 7 or Android 2.3.
 ProGuard
 --------
 
-If you are using ProGuard you might need to add the following options:
+If you are using ProGuard you need to add the following options:
 ```
+# Retain generic type information for use by reflection by converters and adapters.
+-keepattributes Signature
+# Retain service method parameters.
+-keepclassmembernames,allowobfuscation interface * {
+    @retrofit2.http.* <methods>;
+}
+# Ignore annotation used for build tooling.
 -dontnote retrofit2.Platform
 -dontwarn retrofit2.Platform$Java8
 -dontwarn okio.**
 -dontwarn javax.annotation.**
--keepclasseswithmembers class * {
-    @retrofit2.http.* <methods>;
-}
+-dontwarn org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement
 ```
 
 
--- a/pom.xml
+++ b/pom.xml
@@ -49,21 +49,23 @@
 
     <!-- Dependencies -->
     <android.version>4.1.1.4</android.version>
-    <okhttp.version>3.8.0</okhttp.version>
+    <okhttp.version>3.9.1</okhttp.version>
     <animal.sniffer.version>1.14</animal.sniffer.version>
 
     <!-- Adapter Dependencies -->
     <rxjava.version>1.3.0</rxjava.version>
     <rxjava2.version>2.0.0</rxjava2.version>
     <guava.version>19.0</guava.version>
+    <scala.version>2.12.4</scala.version>
 
     <!-- Converter Dependencies -->
-    <gson.version>2.7</gson.version>
+    <gson.version>2.8.2</gson.version>
     <protobuf.version>3.0.0</protobuf.version>
-    <jackson.version>2.7.2</jackson.version>
+    <jackson.version>2.8.0</jackson.version>
     <wire.version>2.2.0</wire.version>
     <simplexml.version>2.7.1</simplexml.version>
-    <moshi.version>1.4.0</moshi.version>
+    <moshi.version>1.5.0</moshi.version>
+    <jaxb.version>2.2.12</jaxb.version><!-- 2.3.0 breaks due to https://github.com/mojohaus/animal-sniffer/issues/29 -->
 
     <!-- Sample Dependencies -->
     <jsoup.version>1.7.3</jsoup.version>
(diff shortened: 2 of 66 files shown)


PromanSEW:
@swankjesse please review

README.md
```diff
@@ -32,8 +32,15 @@ ProGuard
 
 If you are using ProGuard you might need to add the following options:
 ```
+-dontnote retrofit2.Platform
+-dontwarn retrofit2.Platform$Java8
 -dontwarn okio.**
 -dontwarn javax.annotation.**
+-keepattributes Signature
+-keepattributes Exceptions
+-keepclasseswithmembers class * {
```
JakeWharton: This isn't needed. If you're not referencing the interface it should be removed.

README.md
```diff
@@ -32,8 +32,15 @@ ProGuard
 
 If you are using ProGuard you might need to add the following options:
 ```
+-dontnote retrofit2.Platform
+-dontwarn retrofit2.Platform$Java8
 -dontwarn okio.**
 -dontwarn javax.annotation.**
+-keepattributes Signature
+-keepattributes Exceptions
```
JakeWharton: Why is this needed? We don't throw checked exceptions from inside the proxy.

PromanSEW: See http://square.github.io/retrofit/
ProGuard section. I added lines from these recommendations

JakeWharton: We can remove it from both then.

README.md
PromanSEW: If I use these interfaces this block is needed. Maybe note about adding this block is needed here

JakeWharton: We don't use this rule with ProGuard so I'm not convinced it's needed until someone proves otherwise

PromanSEW: OK, I chec again and provide the stacktrace today or tomorrow

PromanSEW: @JakeWharton see provided link in #2338:
https://stackoverflow.com/questions/35506111/retrofit-2-not-sending-data-when-proguard-is-enabled

PromanSEW:
Here is stacktrace without
```
-keepclasseswithmembers class * {
    @retrofit2.http.* <methods>;
}
```
```
HttpCodeException{code=400, response=ApiError{description='Error validating field "path": This field is required.', error='FieldValidationError'}}
at com.yandex.disk.rest.b.b.a(Unknown Source)
at com.yandex.disk.rest.b.b.a(Unknown Source)
at com.yandex.disk.rest.j.a(Unknown Source)
at com.yandex.disk.rest.j.a(Unknown Source)
...
```
Here custom `HttpCodeException` throws when request failed.
```java
private <T> T processResponse(Response<T> response) throws HttpCodeException {
      return response.isSuccessful() ? response.body() : throwHttpCodeException(response);
}
```
I see implementation and here is method from interface:
```java
@GET("/v1/disk/resources/download")
Call<Link> getDownloadLink(@Query("path") String path) throws IOException, ServerIOException;
```
You can see than method has `@Query("path") ` parameter, which is obfuscated while ProGuard processing.
When I added
```
-keepclasseswithmembers class * {
     @retrofit2.http.* <methods>;
}
```
to ProGuard rules, all works fine.

JakeWharton:
Obfuscating the annotation is fine since the reference in Retrofit will
also be updated. The string is a constant so it won't be touched. If the
annotation was removed Retrofit would have thrown an error during
validation.

I'll have to try to reproduce. We don't have this problem internally.

priyanshi-gupta:
With updated JSR-305 annotations, following proguard rules need to be added in ReadMe file:

```
-dontwarn javax.annotation.Nullable
-dontwarn javax.annotation.ParametersAreNonnullByDefault
```

NightlyNexus:
Nullable and ParametersAreNonnullByDefault are already included as `javax.annotation.**`

stevestock:
from #2596...

Please consider keeping [github.com/square/retrofit#proguard](https://github.com/square/retrofit#proguard) and [square.github.io/retrofit](http://square.github.io/retrofit/) ([website code](https://github.com/square/retrofit/blob/master/website/index.html#L176-L189)) in sync:

recommend removing this from the current PR
```
-dontwarn okio.**
-dontwarn javax.annotation.**
```

and adding

> Retrofit uses OKHttp under the hood, so you may want to look at [its ProGuard rules](https://github.com/square/okhttp#proguard) as well.

and then changing the [website code](https://github.com/square/retrofit/blob/master/website/index.html#L176-L189) to match, something like:

> If you are using ProGuard in your project add the following lines to your configuration:
> 
> ```
> # Platform calls Class.forName on types which do not exist on Android to determine platform.
> -dontnote retrofit2.Platform
> # Platform used when running on Java 8 VMs. Will not be used at runtime.
> -dontwarn retrofit2.Platform$Java8
> # ______
>-keepclasseswithmembers class * {
>    @retrofit2.http.* <methods>;
>}
> ```
> Retrofit uses OKHttp under the hood, so you may want to look at [its ProGuard rules](https://github.com/square/okhttp#proguard) as well.

PR #2346, opened 2017-05-25
Add SimpleJSON body support

leleliu008:
We always Post some simple JSON to server,  simple JSON means its fields only has raw type, for example {'age':18}、 {'username':'xxxx', 'password':'yyyyy'}. If we use @Body annotation, we need New a Java class，it seems not simple, you might suggest me to design my API with Restful API, but some old programmers not like Restful API，they only use POST method. so, I add this support.

the usage as follow:
-----------------------------------------------------
public interface HttpAPI {
        @POST("/user-login")
        @SimpleJSON
        Observable<ResponseEntity> changeHeight(@SimpleJSONField("username") String username, @SimpleJSONField("password") String password);
}
-----------------------------------------------------
the usage is similar to @FormUrlEncoded and @Field
-----------------------------------------------------
HTTP Request like as follows:
-----------------------------------------------------
POST /user-change-gender HTTP/1.1
Host: http://192.168.1.100
Content-Type: application/json;charset=UTF-8
Content-Length: 16
{'username':'fpliu','password':'SHA256--------'}
-----------------------------------------------------
add this support, we don't need to new a class，if we have a large number of similar HTTP interfaces，it will save us much time.

### DIFF
--- a/retrofit/src/main/java/retrofit2/ParameterHandler.java
+++ b/retrofit/src/main/java/retrofit2/ParameterHandler.java
@@ -360,4 +360,23 @@ abstract class ParameterHandler<T> {
       builder.setBody(body);
     }
   }
+
+  static final class SimpleJSONField<T> extends ParameterHandler<T> {
+    private final String name;
+    private final Converter<T, String> valueConverter;
+
+    SimpleJSONField(String name, Converter<T, String> valueConverter) {
+      this.name = checkNotNull(name, "name == null");
+      this.valueConverter = valueConverter;
+    }
+
+    @Override void apply(RequestBuilder builder, @Nullable T value) throws IOException {
+      if (value == null) return; // Skip null values.
+
+      String strValue = valueConverter.convert(value);
+      if (strValue == null) return; // Skip converted but null values
+
+      builder.addJSONField(name, strValue);
+    }
+  }
 }
--- a/retrofit/src/main/java/retrofit2/RequestBuilder.java
+++ b/retrofit/src/main/java/retrofit2/RequestBuilder.java
@@ -17,6 +17,8 @@ package retrofit2;
 
 import java.io.IOException;
 import javax.annotation.Nullable;
+
+import android.util.Log;
 import okhttp3.FormBody;
 import okhttp3.Headers;
 import okhttp3.HttpUrl;
@@ -26,11 +28,16 @@ import okhttp3.Request;
 import okhttp3.RequestBody;
 import okio.Buffer;
 import okio.BufferedSink;
+import org.json.JSONException;
+import org.json.JSONObject;
 
 final class RequestBuilder {
   private static final char[] HEX_DIGITS =
       { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F' };
   private static final String PATH_SEGMENT_ALWAYS_ENCODE_SET = " \"<>^`{}|\\?#";
+
+  private static final MediaType CONTENT_TYPE_JSON =
+          MediaType.parse("application/json;charset=UTF-8");
 
   private final String method;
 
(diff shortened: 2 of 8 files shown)


NightlyNexus:
If you absolutely have to use the `org.json.*` map APIs, you could use a custom [Converter](https://github.com/square/retrofit/blob/9168d0d6873ccbb95bd69f13fd1f4fdcd0a68dfd/retrofit/src/main/java/retrofit2/Converter.java).

leleliu008:
I use map insdead of org.json.JSONObject。Although it is not very appropriate，but, SimpleJSON usually contains only a few simple characters.

swankjesse:
Implementing partial JSON encoding in Retrofit isn’t appropriate. Anyone who uses this will be unsatisfied by its limitations, and by being included by default it imposes an implicit preference that I don’t think is appropriate.

As @NightlyNexus suggests, if you want this you should build it as a standalone adapter.

leleliu008:
Thank you for your suggests! but, I don't think a custom [Converter](https://github.com/square/retrofit/blob/9168d0d6873ccbb95bd69f13fd1f4fdcd0a68dfd/retrofit/src/main/java/retrofit2/Converter.java) can meet my requirement. I can not use my own annotations, I have to use like Body、Part、PartMap annotations，they are not convenient for me！

in my opinion, SimpleJSON is similar to x-www-form-urlencoded and multipart, all of them are particular! now, SimpleJSON is More widely used in HTTP request, so add this support can save us a lot of time!

leleliu008:
examples:
`
public interface HttpAPI {

    @POST("/user-register")
    @SimpleJSON
    @Headers("X-User-Agent:com.fpliu.newton")
    Observable<ResponseEntity> register(@SimpleJSONField("username") String username,
                                        @SimpleJSONField("password") String password,
                                        @SimpleJSONField("verifyCode") String verifyCode);

    @POST("/user-login")
    @SimpleJSON
    @Headers("X-User-Agent:com.fpliu.newton")
    Observable<ResponseEntity<UserInfo>> login(@SimpleJSONField("username") String username,
                                               @SimpleJSONField("password") String password);

    @POST("/user-logout")
    @Headers("X-User-Agent:com.fpliu.newton")
    Observable<ResponseEntity> logout();

    @POST("/user-info")
    @Headers("X-User-Agent:com.fpliu.newton")
    Observable<ResponseEntity<UserInfo>> getUserInfo();

    @POST("/user-change-password")
    @SimpleJSON
    @Headers("X-User-Agent:com.fpliu.newton")
    Observable<ResponseEntity> changePassword(@SimpleJSONField("oldPassword") String oldPassword,
                                              @SimpleJSONField("newPassword") String newPassword);

    @POST("/user-reset-password")
    @SimpleJSON
    @Headers("X-User-Agent:com.fpliu.newton")
    Observable<ResponseEntity> resetPassword(@SimpleJSONField("username") String username,
                                             @SimpleJSONField("verifyCode") String verifyCode,
                                             @SimpleJSONField("newPassword") String newPassword);

    @POST("/user-change-nick-name")
    @SimpleJSON
    @Headers("X-User-Agent:com.fpliu.newton")
    Observable<ResponseEntity> changeNickName(@SimpleJSONField("nickName") String nickName);

    @POST("/user-change-phone-number")
    @SimpleJSON
    @Headers("X-User-Agent:com.fpliu.newton")
    Observable<ResponseEntity> changePhoneNumber(@SimpleJSONField("phoneNumber") String phoneNumber);

    @POST("/user-change-gender")
    @SimpleJSON
    @Headers("X-User-Agent:com.fpliu.newton")
    Observable<ResponseEntity> changeGender(@SimpleJSONField("gender") String gender);

    @POST("/user-change-mood")
    @SimpleJSON
    @Headers("X-User-Agent:com.fpliu.newton")
    Observable<ResponseEntity> changeMood(@SimpleJSONField("mood") String mood);
}
`

PR #2360, opened 2017-06-08
fix when Status Codes is 204 NO CONTENT, RxJava throws Exception onNe…

iflove:
fix when Status Codes is 204 NO CONTENT, RxJava throws Exception onNext called with null. Null values are generally not allowed in 2.x operators and sources


JakeWharton:
See https://github.com/square/retrofit/issues/2242#issuecomment-300850574 for details on how to fix this.

JakeWharton:
Also for reasoning why completing as an empty stream is not the correct solution.

iflove:
thanks

PR #2371, opened 2017-06-20
Update README.md - Include Kotlin documentation

leandroBorgesFerreira:
The default instance o Moshi created by MoshiConverterFactory.create() doesn't work with Kotlin's data classes. It took me a while to find out how to make it work. I think it is a good help to include this in the documentation.

### DIFF
--- a/retrofit-converters/moshi/README.md
+++ b/retrofit-converters/moshi/README.md
@@ -1,3 +1,6 @@
+
+
+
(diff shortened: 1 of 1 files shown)


Jawnnypoo:
This is covered in the Moshi README now for the most part 👍

leandroBorgesFerreira:
Ok!

PR #2403, opened 2017-07-11
Add Thrifty converter

reisub:
Adds a converter for [Thrifty](https://github.com/Microsoft/thrifty), modeled after the Wire converter.

It supports binary and compact protocol.


JakeWharton:
We don't ship a regular Thrift converter and we're not Thrift users so I'm not sure we're the best person to be maintaining this. Would you consider releasing this as a separate open source library that we could link to? There's already examples of this ([LoganSquare][1], [fastjson][2]) that we link to from the Wiki, but we could move these links into the `converters/README.md` as well to improve discoverability.

 [1]: https://github.com/aurae/retrofit-logansquare
 [2]: https://github.com/ligboy/retrofit-converter-fastjson

reisub:
@JakeWharton that's a fair point. I figured it would make most sense to have it live with the majority of other converters, and it's really similar to the wire converter anyways.

I'll check with the Thrifty development team first and if they are not interested then we figure something else out.

Thanks for taking the time to review it.

JakeWharton:
If we were already maintaining a Thrift converter there may be a precedent to include this. Since we don't, however, and none of us use Thrift, I don't think it appropriate to include this. I would strongly recommend posting this as a separate library and we can move the third-party converter list from the wiki into the README of `retrofit-converters/` to improve their discoverability.

reisub:
Okay, that makes sense.

The converter is now published as a separate library (we're just waiting for it to be linked to jcenter):
https://github.com/infinum/thrifty-retrofit-converter

I've also updated the wiki here:
https://github.com/square/retrofit/wiki/Converters

JakeWharton:
Thanks!

PR #2404, opened 2017-07-12
Create ResponseObservable

JSpiner:
Until now, to use rxjava-retrofit-adapter to get status code or response information, we had to mapping object twice. like `Observable<Response<ItemModel>>`.
Status codes are always needed in most api. So we define `Response` again and again...
But this makes the source code hard to see. Nobody wants it.

So I created a class `ResponseObservable`. Its `ResponseObservable.ResponseObserver` allows you to subscribe to a Response object. Like below.

```java
//before you know ResponseObservable
  interface Service {
    ...
    // Observable<Response<List<blahblah>> is too complicated...
    @GET("/api1") Observable<Response<String>> getRepository();
    @GET("/api2") Observable<Response<ItemModel>> getRepository2();
    @GET("/api3") Observable<Response<List<ItemModel>>> getRepository3();
    ...
  }
  
  void request(){
    service.getRepository().subscribe(new Observer<Response<String>>() {
      ...
      @Override
      public void onNext(Response<String> response) {
        int code = response.code();
        ...
      }
      ...
    });
  }

```

```java
//after you know ResponseObservable
  interface Service {
    ...
    // No need to be complicated anymore. Nice!
    @GET("/api1") ResponseObservable<String> getRepository();
    @GET("/api2") ResponseObservable<ItemModel> getRepository2();
    @GET("/api3") ResponseObservable<List<ItemModel>> getRepository3();
    ...
  }
  
  // same usage
  void request(){
    service.getRepository().subscribe(new Observer<Response<String>>() {
      ...
      @Override
      public void onNext(Response<String> response) {
        int code = response.code();
        ...
      }
      ...
    });
  }
```

This is my first pull request, so there might be a mistake. Let me know and I'll fix it. Thank you.

### DIFF
--- a/retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/ResponseObservable.java
+++ b/retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/ResponseObservable.java
@@ -1,0 +1,72 @@
+package retrofit2.adapter.rxjava2;
+
+import io.reactivex.Observable;
+import io.reactivex.Observer;
+import io.reactivex.disposables.Disposable;
+import io.reactivex.exceptions.CompositeException;
+import io.reactivex.exceptions.Exceptions;
+import io.reactivex.plugins.RxJavaPlugins;
+import retrofit2.Response;
+
+
+/**
+ * Created by jspiner on 2017. 7. 12..
+ */
+public class ResponseObservable<T> extends Observable<T> {
+    private final Observable<Response<T>> upstream;
+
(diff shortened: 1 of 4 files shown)


JakeWharton:
This is a lot of code for very little gain. You're saving two characters in the return type without any actual functional change in the consumer behavior.

Retrofit provides the CallAdapter abstraction to allow you to use your own return types in the interface declaration while delegating to the normal types for behavior (see the samples). Something like this is appropriate using that mechanism in your code base, but not for being baked into Retrofit itself.

PR #2434, opened 2017-08-01
(refactor) some opportunities to use lambda functions

desmorto:


JakeWharton:
This library targets Java 6 so this isn't feasible

PR #2442, opened 2017-08-06
Optimize Rx2Adapter - Don't clone the Call if not executed

digitalbuddha:
Currently, a call is always cloned when a new subscriber used the `RxJava2Adapter`. This pr first checks if a call is executed prior to cloning.  

I needed to pull `retrofitMock` in as a dependency to the adapters module to ease in testing. Let me know if overkill and I can remove the test or switch the  `FakeCall` to an anonymous implementation of a `Call<T>`


retrofit-adapters/rxjava2/pom.xml
```diff
@@ -49,5 +49,10 @@
       <artifactId>mockwebserver</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.squareup.retrofit2</groupId>
```
JakeWharton: This should be `${project.groupId}`

retrofit-adapters/rxjava2/pom.xml
```diff
@@ -49,5 +49,10 @@
...
+      <artifactId>retrofit-mock</artifactId>
+      <version>2.4.0-SNAPSHOT</version>
```
JakeWharton: This should be `${project.version}`

retrofit-adapters/rxjava2/pom.xml
```diff
@@ -49,5 +49,10 @@
...
+    </dependency>
```
JakeWharton: This needs to be in the "test" scope, otherwise you leak the dependency to consumers.

retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/CallExecuteObservable.java
```diff
@@ -61,6 +61,13 @@
     }
   }
 
+  Call<T> cloneCallIfAlreadyExecuted(Call<T> call) {
```
JakeWharton: This isn't worth a method–especially not a virtual one. Let's just do a ternary.
```java
Call<T> call = originalCall.isExecuted()
    ? originalCall.clone()
    : originalCall;
```

retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/CallExecuteObservable.java
```diff
@@ -32,8 +32,8 @@
   }
 
   @Override protected void subscribeActual(Observer<? super Response<T>> observer) {
-    // Since Call is a one-shot type, clone it for each new observer.
```
JakeWharton: Don't lose this comment, just update it to reflect the new behavior.

retrofit-adapters/rxjava2/src/test/java/retrofit2/adapter/rxjava2/CallExecuteObservableTest.java
```diff
@@ -0,0 +1,55 @@
...
+package retrofit2.adapter.rxjava2;
+
+import org.junit.Before;
+import org.junit.Test;
+import retrofit2.Call;
+import retrofit2.mock.Calls;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class CallExecuteObservableTest {
```
JakeWharton: nit: final

retrofit-adapters/rxjava2/src/test/java/retrofit2/adapter/rxjava2/CallExecuteObservableTest.java
```diff
@@ -0,0 +1,55 @@
...
+  private Call<String> executedCall;
+  private Call<String> nonExecutedCall;
+
+
+  @Before
+  public void setUp() throws Exception {
+    executedCall = Calls.response("executed");
+    executedCall.execute();
+    nonExecutedCall = Calls.response("nonExecuted");
+
```
JakeWharton: nit: remove empty line

retrofit-adapters/rxjava2/src/test/java/retrofit2/adapter/rxjava2/CallExecuteObservableTest.java
```diff
@@ -0,0 +1,55 @@
...
+import retrofit2.mock.Calls;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class CallExecuteObservableTest {
+
+  private Call<String> executedCall;
+  private Call<String> nonExecutedCall;
+
+
```
JakeWharton: nit: remove extra empty line

retrofit-adapters/rxjava2/src/test/java/retrofit2/adapter/rxjava2/CallExecuteObservableTest.java
```diff
@@ -0,0 +1,55 @@
...
+  @Before
+  public void setUp() throws Exception {
+    executedCall = Calls.response("executed");
+    executedCall.execute();
+    nonExecutedCall = Calls.response("nonExecuted");
+
+  }
+
+
```
JakeWharton: nit: remove extra empty line

digitalbuddha:
comments addressed, @JakeWharton inlining the `clone` check made the test less specfc. Best I can test for is  whether the `CallExecuteObservable` can successfully be subscribed to twice. Let me know if test is valid or should be dropped. Thank you for review

retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/CallExecuteObservable.java
```diff
@@ -32,8 +32,10 @@
...
-    Call<T> call = originalCall.clone();
+    // Since Call is a one-shot type, clone it for each new observer when already executed.
+    Call<T> call = originalCall.isExecuted()
```
swankjesse: I think this is racy

swankjesse:
I think this new code is racy and I don’t see a good way to make it not racy. In theory here two concurrent calls could skip the clone, which would be problematic.

swankjesse:
... you could do `synchronized (call) { ... }` around the ternary. That’d be safe.

digitalbuddha:
As much as I want my first pr to land is it worth the cost of synchronization to negate the cost of cloning? As in should I close or continue?

JakeWharton:
```
$ java -cp jol-cli-0.8-full.jar:retrofit-2.3.0.jar:okhttp-3.8.1.jar:okio-1.13.0.jar org.openjdk.jol.Main internals retrofit2.OkHttpCall
objc[96986]: Class JavaLaunchHelper is implemented in both /Library/Java/JavaVirtualMachines/jdk1.8.0_131.jdk/Contents/Home/bin/java (0x101bc34c0) and /Library/Java/JavaVirtualMachines/jdk1.8.0_131.jdk/Contents/Home/jre/lib/libinstrument.dylib (0x11f4854e0). One of the two will be used. Which one is undefined.
# WARNING: Unable to attach Serviceability Agent. You can try again with escalated privileges. Two options: a) use -Djol.tryWithSudo=true to try with sudo; b) echo 0 | sudo tee /proc/sys/kernel/yama/ptrace_scope
# Running 64-bit HotSpot VM.
# Using compressed oop with 3-bit shift.
# Using compressed klass with 3-bit shift.
# WARNING | Compressed references base/shifts are guessed by the experiment!
# WARNING | Therefore, computed addresses are just guesses, and ARE NOT RELIABLE.
# WARNING | Make sure to attach Serviceability Agent to get the reliable addresses.
# Objects are 8 bytes aligned.
# Field sizes by type: 4, 1, 1, 2, 2, 4, 4, 8, 8 [bytes]
# Array element sizes: 4, 1, 1, 2, 2, 4, 4, 8, 8 [bytes]

Instantiated the sample instance via retrofit2.OkHttpCall(retrofit2.ServiceMethod,java.lang.Object[])

retrofit2.OkHttpCall object internals:
 OFFSET  SIZE                      TYPE DESCRIPTION                               VALUE
      0     4                           (object header)                           05 00 00 00 (00000101 00000000 00000000 00000000) (5)
      4     4                           (object header)                           00 00 00 00 (00000000 00000000 00000000 00000000) (0)
      8     4                           (object header)                           e0 21 01 f8 (11100000 00100001 00000001 11111000) (-134143520)
     12     1                   boolean OkHttpCall.canceled                       false
     13     1                   boolean OkHttpCall.executed                       false
     14     2                           (alignment/padding gap)
     16     4   retrofit2.ServiceMethod OkHttpCall.serviceMethod                  null
     20     4        java.lang.Object[] OkHttpCall.args                           null
     24     4              okhttp3.Call OkHttpCall.rawCall                        null
     28     4       java.lang.Throwable OkHttpCall.creationFailure                null
Instance size: 32 bytes
Space losses: 2 bytes internal + 0 bytes external = 2 bytes total
```
Doesn't seem worth it to save 32 bytes...

JakeWharton:
Sorry this was my bad. I didn't think through the stuff Jesse ended up uncovering when I said "go for it".

digitalbuddha:
No worries. Swing and a miss ☺️ enjoy your night folks.

naturalwarren:
This PR was super instructive as I was looking at doing something similar in a custom adapter. Thanks for exploring this.

PR #2458, opened 2017-08-18
Use "." as default String parameter for all HTTP method annotations.

julioromano:
This will make using `@GET` (or similar) without `@Url` possible and will call the base url.
N.B. Calling the base URL without any further path paremeters is still generally not a good idea, but at least now it won't crash at runtime.


julioromano:
Nevermind. What I was trying to achieve is not feasible as there is no way to tell a `@GET` with default "." String parameter from an explicit `@GET(".")`.

PR #2462, opened 2017-08-23
provide more flexibility

Stuazt:
Allow developers to do something with the Observable. Like this:

```
public abstract class HttpObservable<T> extends Observable<T> {

    public final void smartSubscribe(Observer<T> observer)
    {
        subscribeOn(Schedulers.io()).observeOn(AndroidSchedulers.mainThread()).subscribe(observer);
    }

}
```


JakeWharton:
We have no way of knowing how to instantiate an `HttpObservable` so this will not work. You can write your own `CallAdapter` factory to support creating `Observable` subclasses or follow the RxJava threading example in the samples to show how to always add threading operators to the returned `Observable` instances.

Stuazt:
Thanks a lot. So excited to get a response from YOU! You're right. I should dig deeper into the source code.

PR #2475, opened 2017-09-08
duplicate the license in the leaf pom.xml

martinbonnin:
this allows google oss-licenses (https://developers.google.com/android/guides/opensource) to pick them up


JakeWharton:
I filed an issue internally about this. Going to wait to see what they say. Needing to do this for some random plugin isn't something I want to do.

martinbonnin:
Makes sense. Thanks for reporting this internally. 
Maybe the plugin could fallback to the github licenses which might be more reliable and/or, as a last resort, add an api to manually append undetected licenses (but having to do this for a lot of projects kind of defeats the purpose of the plugin...).

Issue #2471, opened 2017-09-06
API Proposal - Retrofit.Builder.converterFactories()


naturalwarren:
I have a `Retrofit` instance that's configured with a bunch of `CallAdapter.Factory` instances. I want to be able to call `newBuilder()` and swap in a new `GsonConverterFactory` built with a new `Gson` instance but `Retrofit.Builder` doesn't expose `List<Converter.Factory>`.

Would a PR be welcomed that adds such an API? Also, would we want a similar API for `List<CallAdapter.Factory>`?

baristaze:
That would be a great addition.

vidhithakrar:
Retrofit.Builder should not expose List<Converter.Factory>. Retrofit already exposes it via converterFactories() method.

naturalwarren:
Once it's exposed as part of a Retrofit instance the list should be immutable and not allow for further configuration though right? As part of a Builder I'd like to modify the list in a similar fashion to how OkHttpClient.Builder let's you access a mutable list of interceptors.

JakeWharton:
Feel free to send a PR to make the list public.

There's a decent amount of subtlety that has to be handled and tested. The
default converters which must be placed first must be moved into the
build() method so they're not exposed in the list. It also needs removed
when copying the existing set into the builder in newBuilder().

naturalwarren:
Sounds good. Thanks!

naturalwarren:
Reopened to track adding a sister API for call adapter factories.

PR #2476, opened 2017-09-10
Add Retrofit.Builder.converterFactories().

naturalwarren:
Addresses #2471.

The default converters are now added in `Builder.build()` and removed when copying the existing set into the builder in `newBuilder()`.

### DIFF
--- a/retrofit/src/main/java/retrofit2/Retrofit.java
+++ b/retrofit/src/main/java/retrofit2/Retrofit.java
@@ -582,11 +582,12 @@ public final class Retrofit {
       adapterFactories.add(platform.defaultCallAdapterFactory(callbackExecutor));
 
       // Make a defensive copy of the converters.
-      List<Converter.Factory> converterFactories = new ArrayList<>(this.converterFactories);
+      List<Converter.Factory> converterFactories = new ArrayList<>();
 
       // Add the built-in converter factory first. This prevents overriding its behavior but also
       // ensures correct behavior when using converters that consume all types.
-      converterFactories.add(0, new BuiltInConverters());
+      converterFactories.add(new BuiltInConverters());
+      converterFactories.addAll(this.converterFactories);
 
       return new Retrofit(callFactory, baseUrl, converterFactories, adapterFactories,
           callbackExecutor, validateEagerly);
--- a/retrofit/src/main/java/retrofit2/Retrofit.java
+++ b/retrofit/src/main/java/retrofit2/Retrofit.java
@@ -242,7 +242,7 @@ public final class Retrofit {
   }
 
   /**
-   * Returns a list of the factories tried when creating a
+   * Returns an unmodifiable list of the factories tried when creating a
    * {@linkplain #requestBodyConverter(Type, Annotation[], Annotation[]) request body converter}, a
    * {@linkplain #responseBodyConverter(Type, Annotation[]) response body converter}, or a
    * {@linkplain #stringConverter(Type, Annotation[]) string converter}.
@@ -402,9 +402,6 @@ public final class Retrofit {
 
     Builder(Platform platform) {
       this.platform = platform;
-      // Add the built-in converter factory first. This prevents overriding its behavior but also
-      // ensures correct behavior when using converters that consume all types.
-      converterFactories.add(new BuiltInConverters());
     }
 
     public Builder() {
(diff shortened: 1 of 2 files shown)


naturalwarren:
@FabianTerhorst is there something specific you don't like in this PR? We'd find it useful to reconfigure Retrofit instances with different converters while hanging onto our adapters. The converter API added is only present in the builder.

JakeWharton:
I'll try to get to this tomorrow.

retrofit/src/main/java/retrofit2/Retrofit.java
```diff
@@ -580,6 +584,10 @@ public Retrofit build() {
       // Make a defensive copy of the converters.
       List<Converter.Factory> converterFactories = new ArrayList<>(this.converterFactories);
```
swankjesse: nit: if you create a new ArrayList, and add `BuiltInConverters` first, then use `addAll()`, you can avoid shifting all _N_ converter factories. Very small performance improvement, but the type of thing I like to keep a clean house

retrofit/src/test/java/retrofit2/RetrofitTest.java
```diff
@@ -171,6 +171,37 @@
     assertSame(callFactory, two.callFactory());
   }
 
+  @Test public void builtInConvertersAbsentInCloneBuilder() {
+    Retrofit retrofit = new Retrofit.Builder()
+        .baseUrl(server.url("/"))
+        .build();
+
+    assertEquals(retrofit.converterFactories().size() - 1,
```
swankjesse: should this just assert that it’s empty? Otherwise if we split the built in ones across two factories then this test is wrong

retrofit/src/test/java/retrofit2/RetrofitTest.java
```diff
@@ -171,6 +171,37 @@
...
+        retrofit.newBuilder().converterFactories().size());
+  }
+
+  @Test public void builtInConvertersAddedOnBuild() {
```
swankjesse: This feels like it’s testing an implementation detail?

swankjesse: Basically anywhere you have `+ 1` tests something we don’t want to enforce. For example Moshi uses 5 built-in factories.

retrofit/src/test/java/retrofit2/RetrofitTest.java
```diff
@@ -171,6 +171,37 @@
...
+    Retrofit.Builder builder = new Retrofit.Builder();
+    Retrofit retrofit = builder
+        .baseUrl(server.url("/"))
+        .build();
+
+    assertEquals(builder.converterFactories().size() + 1, retrofit.converterFactories().size());
+  }
+
+  @Test public void builtInConvertersRemainFirstInClone() {
```
swankjesse: I think the test you really want is that the converterFactories on the builder is mutable and initially empty

retrofit/src/test/java/retrofit2/RetrofitTest.java
JakeWharton: This is probably implicitly tested wherever we test `Call<Void>`. I guess the right way to black box test that we don't regress is to install a custom converter for `Void` and validate that it's _not_ called.

retrofit/src/test/java/retrofit2/RetrofitTest.java
naturalwarren: That sounds like a much more robust test. b2b3b841

retrofit/src/test/java/retrofit2/RetrofitTest.java
naturalwarren: Agreed that this is testing an implementation detail. Removed this test in 4d4cee4c. 

Added Jake's suggested black box test in e5d30fc7.

retrofit/src/test/java/retrofit2/RetrofitTest.java
naturalwarren: b2b3b841 has got the second part of that covered. 

Mutability seems like something that's an implementation detail. Thoughts?

retrofit/src/main/java/retrofit2/Retrofit.java
naturalwarren: Great catch. 5edfa32c

swankjesse:
This is great! Last step is the CLA. I don’t think we have you on it yet.
https://docs.google.com/forms/d/e/1FAIpQLSeRVQ35-gq2vdSxD1kdh7CJwRdjmUA0EZ9gRXaWYoUeKPZEQQ/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1

naturalwarren:
I signed it!

JakeWharton:
Want to do converters as well?

naturalwarren:
Not a bad idea, will do. Thanks @JakeWharton.

PR #2486, opened 2017-09-18
Add Retrofit.Builder.adapterFactories().

naturalwarren:
Adds a sister API to `Retrofit.Builder.converterFactories()` for adapter factories.

Follow-up to #2471.

### DIFF
--- a/retrofit/src/main/java/retrofit2/Retrofit.java
+++ b/retrofit/src/main/java/retrofit2/Retrofit.java
@@ -542,6 +542,11 @@ public final class Retrofit {
       return this;
     }
 
+    /** Returns a modifiable list of call adapter factories. */
+    public List<CallAdapter.Factory> adapterFactories() {
+      return this.adapterFactories;
+    }
+
     /** Returns a modifiable list of converter factories. */
     public List<Converter.Factory> converterFactories() {
       return this.converterFactories;
--- a/retrofit/src/main/java/retrofit2/Retrofit.java
+++ b/retrofit/src/main/java/retrofit2/Retrofit.java
@@ -543,7 +543,7 @@ public final class Retrofit {
     }
 
     /** Returns a modifiable list of call adapter factories. */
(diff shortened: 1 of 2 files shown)


retrofit/src/main/java/retrofit2/Retrofit.java
```diff
@@ -542,6 +542,11 @@ public Builder callbackExecutor(Executor executor) {
       return this;
     }
 
+    /** Returns a modifiable list of call adapter factories. */
```
NightlyNexus: For both `converterFactories()` and `adapterFactories()`, is "modifiable" sufficiently descriptive?
That is, it's very different from returning `new ArrayList<>(adapterFactories)` which is also mutable.
This exposes the builder's list to modify the builder's state.

naturalwarren: This language is taken from `OkHttpClient.interceptors()` and `OkHttpClient.networkInterceptors()`: 

"Returns a modifiable list of interceptors ..."

NightlyNexus: Ah, cool, precedence. Sgtm!

retrofit/src/main/java/retrofit2/Retrofit.java
```diff
@@ -542,6 +542,11 @@ public Builder callbackExecutor(Executor executor) {
...
+    public List<CallAdapter.Factory> adapterFactories() {
```
JakeWharton: This should be called `callAdapterFactories()`

naturalwarren:
The default adapter is already added in `Retrofit.Builder.build()` and removed in `Builder(Retrofit retrofit)` (unlike the converters) so there wasn't much to do here.

retrofit/src/main/java/retrofit2/Retrofit.java
naturalwarren: Renamed in 57a39e68.

PR #2490, opened 2017-09-20
Add initial capacity for converter factory list.

NightlyNexus:
Sorry for the low-effort changes. Getting caught back up.

Dunno if this is worth it or too awkward of a change...


JakeWharton:
I was waiting for CI to go green but 3 hours is too much

PR #2511, opened 2017-10-05
Update Readme to warn about interface inheritance

vitorOta:
according to this issue https://github.com/square/retrofit/issues/504, I believe that is good to let this well clear


swankjesse:
This is not how we’ve organized our documentation. What other documentation did you read?

PR #2528, opened 2017-10-17
 add fruit converter for retrofit

ghuiii:
add [fruit](https://github.com/ghuiii/fruit) converter for retrofit
> Fruit is a Java library which can be used to convert html to Java Objects  follow a certain rule. So, in short, Fruit for Html just like Gson for Json.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
     <wire.version>2.2.0</wire.version>
     <simplexml.version>2.7.1</simplexml.version>
     <moshi.version>1.5.0</moshi.version>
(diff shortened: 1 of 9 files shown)


JakeWharton:
We are not adding new converters for random serialization libraries. In fact, we'd like to even get rid of some of the first-party ones in a future major version. Feel free to release your converter as a standalone artifact and then [add it to the converter wiki page][1] with the other third-party ones.

 [1]: https://github.com/square/retrofit/wiki/Converters

ghuiii:
@JakeWharton Thanks JW!

PR #2546, opened 2017-11-06
Add implementation in gradle at website for android studio 3.0

jgabrielfreitas:
website changes


JakeWharton:
I don't think this is an improvement. It doesn't address api vs. implementation, this is controlled by the Android Gradle Plugin version not Android Studio version, and the library isn't Android-specific so mentioning AS only adds confusion for non-Android users.

Issue #2329, opened 2017-05-12
Add Kotlin Unit Converter


consp1racy:
As per https://medium.com/square-corner-blog/square-open-source-loves-kotlin-c57c21710a17 please add support for `kotlin.Unit` type so consumers don't have to return `Call<Void>` in Kotlin code. Cheers!

JakeWharton:
This is really easy. Can do.

JakeWharton:
Only downside is we have to check for `kotlin.Unit` using its qualified name. This means if you use ProGuard or jarjar to rewrite the Kotlin stdlib it won't work.

The alternative to that is a separate dependency and explicit converter you have to install.

consp1racy:
Dammit, didn't account for rewrites.

Can we slap more things together to a Kotlin artifact to make it worthwhile?

JakeWharton:
Hmm like what?

consp1racy:
Phew...

`retrofit.create<MyService>()` instead of `retrofit.create<MyService>(MyService::class.java)`

that's all I could find right now. I guess other candidates for extension functions could be worth exploring.

JakeWharton:
Yep, good one. I'll think about it more.

swankjesse:
Will it be compiled to `Void` in the classfile?

JakeWharton:
No it stays as Unit. Unlike Void which can only be null, Unit can only be
Unit.INSTANCE.

consp1racy:
`Nothing?` can be `null` but that's counterintuitive to use.

JakeWharton:
BTW this is trivial to add yourself if you don't want to wait for official support:
```kotlin
object UnitConverterFactory : Converter.Factory() {
  override fun responseBodyConverter(type: Type, annotations: Array<out Annotation>,
      retrofit: Retrofit): Converter<ResponseBody, *>? {
    return if (getRawType(type) == Unit::class.java) UnitConverter else null
  }

  private object UnitConverter : Converter<ResponseBody, Unit> {
    override fun convert(value: ResponseBody) {
      value.close()
    }
  }
}
```

jaredsburrows:
@JakeWharton About the proguad comment, it looks like `Class.forName` calls might work: https://www.guardsquare.com/en/proguard/faq:

> Does ProGuard handle Class.forName calls?
[...]

PR #2550, opened 2017-11-08
Add a Kotlin coroutine adapter.

JakeWharton:
This supports returning a `Deferred` from your service methods allowing an easy call to `await()` inside a suspending function.

Kotlin coroutines are experimental. This does not make them API unstable, however, which is essential for including them inside Retrofit. When non-experimental coroutines are released we will add a second adapter for them and mark this one as deprecated for removal in the next major version. For more information see https://stackoverflow.com/a/46240340/132047.

Refs #2329.

### DIFF
--- a/retrofit-adapters/kotlin-coroutines-experimental/src/main/java/retrofit/adapter/kotlin/coroutines/experimental/CoroutineCallAdapterFactory.kt
+++ b/retrofit-adapters/kotlin-coroutines-experimental/src/main/java/retrofit/adapter/kotlin/coroutines/experimental/CoroutineCallAdapterFactory.kt
@@ -1,0 +1,141 @@
+/*
+ * Copyright (C) 2016 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package retrofit.adapter.kotlin.coroutines.experimental
+
+import kotlinx.coroutines.experimental.CompletableDeferred
+import kotlinx.coroutines.experimental.Deferred
+import retrofit2.Call
+import retrofit2.CallAdapter
+import retrofit2.Callback
+import retrofit2.HttpException
+import retrofit2.Response
+import retrofit2.Retrofit
+import java.lang.reflect.ParameterizedType
+import java.lang.reflect.Type
+
+/**
+ * A [CallAdapter.Factory] for use with Kotlin coroutines.
+ *
+ * Adding this class to [Retrofit] allows you to return [Deferred] from
+ * service methods.
+ *
+ *     interface MyService {
+ *       &#64;GET("user/me")
+ *       Deferred&lt;User&gt; getUser()
+ *     }
+ *
+ * There are two configurations supported for the [Deferred] type parameter:
+ *
+ * * Direct body (e.g., `Deferred<User>`) returns the deserialized body for 2XX responses, throws
+ * [HttpException] errors for non-2XX responses, and throws [IOException][java.io.IOException] for
+ * network errors.
+ * * Response wrapped body (e.g., `Deferred<Response<User>>`) returns a [Response] object for all
+ * HTTP responses and throws [IOException][java.io.IOException] for network errors
+ */
+class CoroutineCallAdapterFactory private constructor() : CallAdapter.Factory() {
+  companion object {
+    @JvmStatic @JvmName("create")
+    operator fun invoke() = CoroutineCallAdapterFactory()
+  }
+
+  override fun get(
+      returnType: Type,
+      annotations: Array<out Annotation>,
+      retrofit: Retrofit
+  ): CallAdapter<*, *>? {
+    if (Deferred::class.java != getRawType(returnType)) {
+      return null
+    }
+    if (returnType !is ParameterizedType) {
+      throw IllegalStateException(
+          "Deferred return type must be parameterized as Deferred<Foo> or Deferred<out Foo>")
+    }
+    val responseType = getParameterUpperBound(0, returnType)
+
+    val rawDeferredType = getRawType(responseType)
+    return if (rawDeferredType == Response::class.java) {
+      if (responseType !is ParameterizedType) {
+        throw IllegalStateException(
+            "Response must be parameterized as Response<Foo> or Response<out Foo>")
+      }
+      ResponseCallAdapter<Any>(getParameterUpperBound(0, responseType))
+    } else {
+      BodyCallAdapter<Any>(responseType)
+    }
+  }
+
+  private class BodyCallAdapter<T>(
+      private val responseType: Type
+  ) : CallAdapter<T, Deferred<T>> {
+
+    override fun responseType() = responseType
+
+    override fun adapt(call: Call<T>): Deferred<T> {
+      val deferred = CompletableDeferred<T>()
+
+      deferred.invokeOnCompletion {
+        if (deferred.isCancelled) {
+          call.cancel()
+        }
+      }
+
+      call.enqueue(object : Callback<T> {
+        override fun onFailure(call: Call<T>, t: Throwable) {
+          deferred.completeExceptionally(t)
+        }
+
+        override fun onResponse(call: Call<T>, response: Response<T>) {
+          if (response.isSuccessful) {
+            deferred.complete(response.body()!!)
+          } else {
+            deferred.completeExceptionally(HttpException(response))
+          }
+        }
+      })
+
+      return deferred
+    }
+  }
+
+  private class ResponseCallAdapter<T>(
+      private val responseType: Type
+  ) : CallAdapter<T, Deferred<Response<T>>> {
+
+    override fun responseType() = responseType
+
+    override fun adapt(call: Call<T>): Deferred<Response<T>> {
+      val deferred = CompletableDeferred<Response<T>>()
+
+      deferred.invokeOnCompletion {
+        if (deferred.isCancelled) {
+          call.cancel()
+        }
+      }
+
+      call.enqueue(object : Callback<T> {
+        override fun onFailure(call: Call<T>, t: Throwable) {
+          deferred.completeExceptionally(t)
+        }
+
+        override fun onResponse(call: Call<T>, response: Response<T>) {
+          deferred.complete(response)
+        }
+      })
+
+      return deferred
+    }
+  }
+}
--- a/retrofit-adapters/kotlin-coroutines-experimental/src/test/java/retrofit/adapter/kotlin/coroutines/experimental/DeferredTest.kt
+++ b/retrofit-adapters/kotlin-coroutines-experimental/src/test/java/retrofit/adapter/kotlin/coroutines/experimental/DeferredTest.kt
@@ -1,0 +1,111 @@
+/*
+ * Copyright (C) 2016 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package retrofit.adapter.kotlin.coroutines.experimental
+
+import kotlinx.coroutines.experimental.Deferred
+import kotlinx.coroutines.experimental.runBlocking
+import okhttp3.mockwebserver.MockResponse
+import okhttp3.mockwebserver.MockWebServer
+import okhttp3.mockwebserver.SocketPolicy.DISCONNECT_AFTER_REQUEST
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Assert.fail
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import retrofit2.HttpException
+import retrofit2.Response
+import retrofit2.Retrofit
+import retrofit2.http.GET
+import java.io.IOException
+
+class DeferredTest {
+  @get:Rule val server = MockWebServer()
+
+  private lateinit var service: Service
+
+  interface Service {
+    @GET("/") fun body(): Deferred<String>
+    @GET("/") fun response(): Deferred<Response<String>>
+  }
+
+  @Before fun setUp() {
+    val retrofit = Retrofit.Builder()
+        .baseUrl(server.url("/"))
+        .addConverterFactory(StringConverterFactory())
+        .addCallAdapterFactory(CoroutineCallAdapterFactory())
+        .build()
+    service = retrofit.create(Service::class.java)
+  }
+
+  @Test fun bodySuccess200() = runBlocking<Unit> {
+    server.enqueue(MockResponse().setBody("Hi"))
+
+    val deferred = service.body()
+    assertThat(deferred.await()).isEqualTo("Hi")
+  }
+
+  @Test fun bodySuccess404() = runBlocking<Unit> {
+    server.enqueue(MockResponse().setResponseCode(404))
+
+    val deferred = service.body()
+    try {
+      deferred.await()
+      fail()
+    } catch (e: HttpException) {
+      assertThat(e).hasMessage("HTTP 404 Client Error")
+    }
+  }
+
+  @Test fun bodyFailure() = runBlocking {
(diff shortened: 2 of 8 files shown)


JakeWharton:
@elizarov would love a review if you have the time. I'm intending to make the Retrofit example from your talk obsolete!

jaredsburrows:
Great work! I look forward to using this soon.

retrofit-adapters/kotlin-coroutines-experimental/src/main/java/retrofit/adapter/kotlin/coroutines/experimental/CoroutineCallAdapterFactory.kt
```diff
@@ -0,0 +1,137 @@
...
+      }
+
+      call.enqueue(object : Callback<T> {
+        override fun onFailure(call: Call<T>, t: Throwable) {
+          deferred.completeExceptionally(t)
+        }
+
+        override fun onResponse(call: Call<T>, response: Response<T>) {
+          if (response.isSuccessful) {
+            deferred.complete(response.body()!!)
```
bhargavms: does this mean in cases where the server responds with no body (or null body) this adapter is going to throw `NullPointerException` ? (Like in `204 No Content`)

retrofit-adapters/kotlin-coroutines-experimental/src/main/java/retrofit/adapter/kotlin/coroutines/experimental/CoroutineCallAdapterFactory.kt
```diff
@@ -0,0 +1,137 @@
...
+      private val responseType: Type
+  ) : CallAdapter<T, Deferred<T>> {
+
+    override fun responseType() = responseType
+
+    override fun adapt(call: Call<T>): Deferred<T> {
+      val deferred = CompletableDeferred<T>()
+
+      deferred.invokeOnCompletion {
+        call.cancel()
```
artem-zinnatullin: nit: check `if (deferred.isCancelled) {}`? 

same below

JakeWharton: Nice, will do. It's a no-op on completed calls, but it saves taking a lock and a bunch of dispatch.

retrofit-adapters/kotlin-coroutines-experimental/src/main/java/retrofit/adapter/kotlin/coroutines/experimental/CoroutineCallAdapterFactory.kt
JakeWharton: I'll look at it. We don't have the ability to track declared nullability of the response type so we might have to categorically ban nulls. For Rx you have to use the null-object pattern (e.g., `Unit`) or the `Optional` converter.

retrofit-adapters/kotlin-coroutines-experimental/src/main/java/retrofit/adapter/kotlin/coroutines/experimental/CoroutineCallAdapterFactory.kt
artem-zinnatullin: yep, plus a clearer intention

JakeWharton: Yeah I wanted to do a custom JobSupport subclass (which has an explicit cancel callback) but it's API is unstable. Maybe sometime in the future.

artem-zinnatullin: Yeah, I'd just add an inline extension function `invokeOnCancellation()` for now.

This is completely off-topic, but I find Coroutines API pretty badly designed.

Mainly naming of the functions and types and discoverability of API parts:

- `resume()` and `resumeWithException()`, why not `resumeWithValue()`…
- `suspend` keyword just always make me think that function is gonna be slow
- `Deferred` and `Job` while the API is called "Coroutines"… idk
- `async()`, `launch()`, `run()` just doesn't click, maybe only `async`. `run()` unfortunately collides with `run()` extension from stdlib

Not that I'm going to use coroutines anytime soon, but if it's going to be major part of what Kotlin offers, it feels like API should be reworked before GA release

elizarov: @artem-zinnatullin You are welcome to comment on Kotlin coroutines libraries design at https://github.com/Kotlin/kotlinx.coroutines/issues It would be really helpful before they get finalized.

retrofit-adapters/kotlin-coroutines-experimental/src/main/java/retrofit/adapter/kotlin/coroutines/experimental/CoroutineCallAdapterFactory.kt
```diff
@@ -0,0 +1,137 @@
...
+      returnType: Type,
+      annotations: Array<out Annotation>,
+      retrofit: Retrofit
+  ): CallAdapter<*, *>? {
+    if (Deferred::class.java != getRawType(returnType)) {
+      return null
+    }
+    if (returnType !is ParameterizedType) {
+      throw IllegalStateException(
+          "Deferred return type must be parameterized as Deferred<Foo> or Deferred<? extends Foo>")
```
elizarov: `Deferred<out Foo>` maybe, since the code is most likely to be used from Kotlin. Same for the response message a few lines below.

elizarov:
@JakeWharton All in all, looks good to me. It would be really great if, in addition to supporting functions like this:

```
fun doSomething(): Deferred<T>
```

Retrofit would also directly support suspending functions like this:

```
suspend fun doSomething(): T
```

This would likely be not as easy to support in the existing call adapter architecture of Retrofit, since it is not just adapting the result type, but taking care of an additional continuation (callback) parameter. However, if supported, this would be immensely useful, as it would allow to write code in a natural and direct sequential style without any need to call `await` and without any risk to forget it.

retrofit-adapters/kotlin-coroutines-experimental/src/main/java/retrofit/adapter/kotlin/coroutines/experimental/CoroutineCallAdapterFactory.kt
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * There are two configurations supported for the [Deferred] type parameter:
+ *
+ * * Direct body (e.g., `Deferred<User>`) returns the deserialized body for 2XX responses, throws
+ * [HttpException] errors for non-2XX responses, and throws [IOException][java.io.IOException] for
+ * network errors.
+ * * Response wrapped body (e.g., `Deferred<Response<User>>`) returns a [Response] object for all
+ * HTTP responses and throws [IOException][java.io.IOException] for network errors
+ */
+class CoroutineCallAdapterFactory private constructor() : CallAdapter.Factory() {
```
swankjesse: one interesting challenge here is that this class is documented with KDoc, and the rest of Retrofit is documented with Javadoc.

retrofit-adapters/kotlin-coroutines-experimental/src/main/java/retrofit/adapter/kotlin/coroutines/experimental/CoroutineCallAdapterFactory.kt
```diff
@@ -0,0 +1,137 @@
...
+      throw IllegalStateException(
+          "Deferred return type must be parameterized as Deferred<Foo> or Deferred<? extends Foo>")
+    }
+    val responseType = getParameterUpperBound(0, returnType)
+
+    val rawDeferredType = getRawType(responseType)
+    return if (rawDeferredType == Response::class.java) {
+      if (responseType !is ParameterizedType) {
+        throw IllegalStateException(
+            "Response must be parameterized as Response<Foo> or Response<? extends Foo>")
```
swankjesse: `Response<out Foo>` ?

retrofit-adapters/kotlin-coroutines-experimental/src/main/java/retrofit/adapter/kotlin/coroutines/experimental/CoroutineCallAdapterFactory.kt
LouisCAD: What if the endpoint returns a nullable body with a 2XX response and that is ok for me?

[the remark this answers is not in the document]
LouisCAD: @jaredsburrows It's a `KotlinNullPointerException`. Unfortunately, this prevents nullable bodies. This will probably change before it gets merged though

LouisCAD: From my tests, I had to deal with a server which always specified a Content-Type (maybe also an empty String) even when replying with 204 No Content, and it made your `UnitConverterFactory` fail when using gildor's library, so I needed to use Call<Void> plus a modified extension function I called `awaitSuccess()` that would allow it in Kotlin. Hope this will not happen again with this

retrofit-adapters/kotlin-coroutines-experimental/src/test/java/retrofit/adapter/kotlin/coroutines/experimental/DeferredTest.kt
```diff
@@ -0,0 +1,114 @@
...
+  @Test fun bodySuccess404() = runBlocking<Unit> {
+    server.enqueue(MockResponse().setResponseCode(404))
+
+    val deferred = service.body()
+    try {
+      deferred.await()
+      fail()
+    } catch (e: HttpException) {
+      assertThat(e)
+          .isInstanceOf(HttpException::class.java) // Required for backwards compatibility.
```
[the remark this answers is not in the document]
mhdtouban: I though assertThat shouldn't be used in production.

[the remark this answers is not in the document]
LouisCAD: @mhdtouban It is in test sources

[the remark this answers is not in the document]
bhargavms: @jaredsburrows I think its there for backward compatibility for now.

retrofit-adapters/kotlin-coroutines-experimental/src/main/java/retrofit/adapter/kotlin/coroutines/experimental/CoroutineCallAdapterFactory.kt
JakeWharton: Fixed.

retrofit-adapters/kotlin-coroutines-experimental/src/main/java/retrofit/adapter/kotlin/coroutines/experimental/CoroutineCallAdapterFactory.kt
JakeWharton: Fixed.

retrofit-adapters/kotlin-coroutines-experimental/src/main/java/retrofit/adapter/kotlin/coroutines/experimental/CoroutineCallAdapterFactory.kt
JakeWharton: Fixed

retrofit-adapters/kotlin-coroutines-experimental/src/main/java/retrofit/adapter/kotlin/coroutines/experimental/CoroutineCallAdapterFactory.kt
```diff
@@ -0,0 +1,137 @@
+/*
+ * Copyright (C) 2016 Square, Inc.
```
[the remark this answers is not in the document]
JakeWharton: It's derived from old code, hence the old copyright.

retrofit-adapters/kotlin-coroutines-experimental/src/test/java/retrofit/adapter/kotlin/coroutines/experimental/DeferredTest.kt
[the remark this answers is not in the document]
JakeWharton: Removed duplicated type check.

retrofit-adapters/kotlin-coroutines-experimental/src/test/java/retrofit/adapter/kotlin/coroutines/experimental/StringConverterFactory.java
```diff
@@ -0,0 +1,46 @@
...
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import okhttp3.MediaType;
+import okhttp3.RequestBody;
+import okhttp3.ResponseBody;
+import retrofit2.Converter;
+import retrofit2.Retrofit;
+
+final class StringConverterFactory extends Converter.Factory {
+  @Override
```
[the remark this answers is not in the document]
JakeWharton: It never changes so I don't see any value in deduplicating.

retrofit-adapters/kotlin-coroutines-experimental/src/main/java/retrofit/adapter/kotlin/coroutines/experimental/CoroutineCallAdapterFactory.kt
JakeWharton: As expected, there's no access to this information from a call adapter.

If Kotlin used type annotations then on JDK 8 or newer we could access the information, but Android doesn't retain type annotations at runtime so it would be a JVM-only feature even if supported.

retrofit-adapters/kotlin-coroutines-experimental/src/main/java/retrofit/adapter/kotlin/coroutines/experimental/CoroutineCallAdapterFactory.kt
```diff
@@ -0,0 +1,141 @@
...
+      private val responseType: Type
+  ) : CallAdapter<T, Deferred<T>> {
+
+    override fun responseType() = responseType
+
+    override fun adapt(call: Call<T>): Deferred<T> {
+      val deferred = CompletableDeferred<T>()
+
+      deferred.invokeOnCompletion {
+        if (deferred.isCancelled) {
```
NightlyNexus: any good way to add a test for this check?
besides saving all the dispatch, checking `isCancelled` fixes canceling on the underlying OkHttp Call which could affect people checking `Call.isCanceled` in an OkHttp layer.

(canceled in OkHttp and Retrofit, cancelled (L x2) in Kotlin coroutines Job. 🙁)

retrofit-adapters/kotlin-coroutines-experimental/src/main/java/retrofit/adapter/kotlin/coroutines/experimental/CoroutineCallAdapterFactory.kt
LouisCAD: Shouldn't an unchecked cast be used instead of a `!!` then? I see no value in producing a `KotlinNullPointerException` from a code the library users can't edit.

JakeWharton: Definitely not. That would just delay the error rather than squashing it at the point of first detection. Consumers should not be using a converter that returns null.

retrofit-adapters/kotlin-coroutines-experimental/src/main/java/retrofit/adapter/kotlin/coroutines/experimental/CoroutineCallAdapterFactory.kt
JakeWharton: I'll add a test. It'll be bit too white-box for my taste since we need to create the Call and manually invoke the adapter, but I can see the value.

NightlyNexus: Yeah, I think a grayer test of using a okhttp3.Call.Factory would be too much machinery.

Thanks. I didn't mean to add more work, but I appreciate it!

JakeWharton:
I moved this to https://github.com/JakeWharton/retrofit2-kotlin-coroutines-adapter until coroutines become stable.

retrofit-adapters/kotlin-coroutines-experimental/src/main/java/retrofit/adapter/kotlin/coroutines/experimental/CoroutineCallAdapterFactory.kt
JakeWharton: Added in the separate project.

PR #2556, opened 2017-11-15
Handle Errors in OkHttpCall.

### LINKED ISSUE #2507: Cast of OutOfMemoryError to RuntimeException
https://github.com/square/okhttp/issues/3637

```
Fatal Exception: java.lang.ClassCastException: java.lang.OutOfMemoryError cannot be cast to java.lang.RuntimeException
       at retrofit2.OkHttpCall.request(OkHttpCall.java:63)
       at retrofit2.ExecutorCallAdapterFactory$ExecutorCallbackCall.request(ExecutorCallAdapterFactory.java:108)
       at com.abelkit.android.api.AbelCallback.onFailure(AbelCallback.java:47)
       at retrofit2.ExecutorCallAdapterFactory$ExecutorCallbackCall$1$2.run(ExecutorCallAdapterFactory.java:79)
       at android.os.Handler.handleCallback(Handler.java:751)
       at android.os.Handler.dispatchMessage(Handler.java:95)
       at android.os.Looper.loop(Looper.java:154)
       at android.app.ActivityThread.main(ActivityThread.java:6321)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776)
```

NightlyNexus:
If an Error is thrown while creating a request, handle it the same as a RuntimeException.

Closes #2507.

### DIFF
--- a/retrofit/src/test/java/retrofit2/CallTest.java
+++ b/retrofit/src/test/java/retrofit2/CallTest.java
@@ -768,6 +768,41 @@ public final class CallTest {
     assertThat(writeCount.get()).isEqualTo(1);
   }
 
+  @Test public void requestThrowingErrorBeforeExecuteFailsExecute() throws IOException {
+    Retrofit retrofit = new Retrofit.Builder()
+        .baseUrl(server.url("/"))
+        .addConverterFactory(new ToStringConverterFactory())
+        .build();
+    Service service = retrofit.create(Service.class);
+
+    server.enqueue(new MockResponse());
+
+    final AtomicInteger writeCount = new AtomicInteger();
+    Object a = new Object() {
+      @Override public String toString() {
+        writeCount.incrementAndGet();
+        throw new OutOfMemoryError("Broken!");
+      }
+    };
+    Call<String> call = service.postRequestBody(a);
+
+    try {
+      call.request();
+      fail();
+    } catch (OutOfMemoryError e) {
+      assertThat(e).hasMessage("Broken!");
+    }
+    assertThat(writeCount.get()).isEqualTo(1);
+
+    try {
+      call.execute();
+      fail();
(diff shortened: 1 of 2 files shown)


retrofit/src/test/java/retrofit2/CallTest.java
```diff
@@ -828,6 +863,41 @@
...
+    } catch (OutOfMemoryError e) {
+      assertThat(e).hasMessage("Broken!");
+    }
+    assertThat(writeCount.get()).isEqualTo(1);
+
+    try {
+      call.request();
+      fail();
+    } catch (OutOfMemoryError e) {
+      assertThat(e).hasMessage("Broken!");
```
NightlyNexus: this feels weird in to wait to throw the OOM error and then to throw it again when calling request. makes sense for other Errors, though.

swankjesse:
Tough one. The spec for `Error` really hurts here:

> An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch.

Here it’s definitely an improvement because casting an error is just a bug. But in general we really don’t want to be recovering from OutOfMemoryErrors; those leave the runtime in an inconsistent state.

NightlyNexus:
👍
Btw, the alternative bug fix would have been to catch only RuntimeExceptions and IOExceptions instead of all Throwables [here](https://github.com/square/retrofit/blob/ea1e8fb59b704ebc6ddee0b56c0216c542f25015/retrofit/src/main/java/retrofit2/OkHttpCall.java#L94).

brinkkemper:
Any news on when this will be released?

PR #2566, opened 2017-11-21
Don't catch fatal errors.

NightlyNexus:


retrofit/src/main/java/retrofit2/OkHttpCall.java
```diff
@@ -37,7 +38,7 @@
   @GuardedBy("this")
   private @Nullable okhttp3.Call rawCall;
   @GuardedBy("this")
-  private @Nullable Throwable creationFailure; // Either a RuntimeException, Error, or IOException.
```
artem-zinnatullin: Isn't the comment still valid? Or the problem is that it doesn't really add any value?

retrofit/src/test/java/retrofit2/CallTest.java
```diff
@@ -781,24 +781,24 @@
     Object a = new Object() {
       @Override public String toString() {
         writeCount.incrementAndGet();
-        throw new OutOfMemoryError("Broken!");
+        throw new AssertionError("Broken!");
```
artem-zinnatullin: nit: `AssertionError` doesn't sound like "recoverable" :D

retrofit/src/test/java/retrofit2/CallTest.java
```diff
@@ -1125,9 +1125,49 @@
...
+    };
+    Call<String> call = service.postRequestBody(a);
+
+    try {
+      call.enqueue(new Callback<String>() {
+        @Override public void onResponse(Call<String> call, Response<String> response) {
+        }
+
+        @Override public void onFailure(Call<String> call, Throwable t) {
+        }
```
artem-zinnatullin: add a `fail()` inside `onFailure()`? Just to verify that error is not delivered to a listener before throw

retrofit/src/main/java/retrofit2/OkHttpCall.java
NightlyNexus: it could be any custom throwable, now/ (hopefully nobody's doing that, though.)

retrofit/src/test/java/retrofit2/CallTest.java
NightlyNexus: yeah, i'd normally do that, but these callbacks are hit asynchronously, so you'd never be able to test that...

retrofit/src/test/java/retrofit2/CallTest.java
NightlyNexus: yeah, maybe rename the tests to "nonFatal" and "fatal" instead of "recoverable" and "unrecoverable"?

retrofit/src/main/java/retrofit2/Utils.java
```diff
@@ -494,4 +494,15 @@ static Type getCallResponseType(Type returnType) {
       return "? extends " + typeToString(upperBound);
     }
   }
+
+  // github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/exceptions/Exceptions.java#L66
```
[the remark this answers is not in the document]
NightlyNexus: 👍 is the 2.x tag not a good permalink?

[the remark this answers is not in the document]
NightlyNexus: oh, i just realized that's their master branch, not any tag. yup, will fix.

retrofit/src/test/java/retrofit2/CallTest.java
NightlyNexus: i could add a check to make sure the callback isn't hit synchronously, but that might confuse the test...

artem-zinnatullin: Yep, that was the point, check that error is not synchronously delivered before throw, it's a nit and up2u of course :)

retrofit/src/test/java/retrofit2/CallTest.java
artem-zinnatullin: yep, I definitely like that it matches the `throwIfFatal` naming!

retrofit/src/test/java/retrofit2/CallTest.java
NightlyNexus: yeah, i liked it after seeing it. done, thanks!

retrofit/src/test/java/retrofit2/CallTest.java
```diff
@@ -781,24 +782,24 @@
...
       }
     };
     Call<String> call = service.postRequestBody(a);
 
     try {
       call.request();
       fail();
-    } catch (OutOfMemoryError e) {
-      assertThat(e).hasMessage("Broken!");
+    } catch (AssertionError expected) {
```
JakeWharton: nit: we tend to only use `expected` when there is no `catch` body.

NightlyNexus: ah, i didn't realize that, thanks. i'll update to `e`.

retrofit/src/test/java/retrofit2/CallTest.java
```diff
@@ -781,23 +782,23 @@
...
+    } catch (AssertionError e) {
       assertThat(e).hasMessage("Broken!");
     }
     assertThat(writeCount.get()).isEqualTo(1);
 
     try {
       call.execute();
       fail();
-    } catch (OutOfMemoryError e) {
+    } catch (AssertionError e) {
```
artem-zinnatullin: huh, just clicked that `fail()` also throws exception that is instance of `AssertionError`, exception message will catch that but still looks scary, maybe just throw `Error()` or define a custom one?

Same applies to tests below

NightlyNexus: yeah, i thought about that but liked AssertionError because it looked like something you might actually use in real code.
plain Error seems alright, though.

artem-zinnatullin: Ah, I see your reasoning, but `AssertionError` is basically a standard for testing frameworks/libraries, would be nice to avoid accidental clashing with that :)

swankjesse:
Nice!

Issue #2277, opened 2017-04-11
Add JAXB Converter


sblommers:
Some of us still use jaxb2 maven plugins to generate JAXB (javax.xml.bind.*) annotated java using jaxb bindings and that sort of stuff. Simplexml does not have a mavenized generator. I was hoping I could somehow easily convert from jaxb to simplexml (bindings) but that ultimately means hand-changing the generated code to switch the annotations. I made this _quick'n'dirty_ implementation that can convert JAXB xml annotated pojos and be used as a replacement convertor.

Maybe someone finds this useful or we can add this to the convertor plugins?

**Usage**:
````java
.....
.addConverterFactory(JAXBConverterFactory.create())
....
````

**JAXBConverterFactory**
````java
package retrofit2.converter.xml;

import java.lang.annotation.Annotation;
import java.lang.reflect.Type;

import okhttp3.RequestBody;
import okhttp3.ResponseBody;
import retrofit2.Converter;
import retrofit2.Retrofit;

public final class JAXBConverterFactory extends Converter.Factory {

	  public static JAXBConverterFactory create() {
		  return new JAXBConverterFactory();
	  }

	  @Override
	  public Converter<ResponseBody, ?> responseBodyConverter(Type type, Annotation[] annotations,
	      Retrofit retrofit) {
	    try {
			return new JAXBResponseBodyConverter<>(type);
		} catch (ClassNotFoundException e) {
			e.printStackTrace();
		}
	    return null;
	  }

	  @Override
	  public Converter<?, RequestBody> requestBodyConverter(Type type,
	      Annotation[] parameterAnnotations, Annotation[] methodAnnotations, Retrofit retrofit) {
	    return new JAXBRequestBodyConverter<>();
	  }
}
````
**JAXBRequestBodyConverter**
````java
package retrofit2.converter.xml;

import java.io.IOException;
import java.io.StringWriter;

import javax.xml.bind.JAXBContext;
import javax.xml.bind.JAXBException;
import javax.xml.bind.Marshaller;

import okhttp3.MediaType;
import okhttp3.RequestBody;
import retrofit2.Converter;

final class JAXBRequestBodyConverter<T> implements Converter<T, RequestBody> {
	private static final MediaType MEDIA_TYPE = MediaType.parse("application/xml; charset=UTF-8");

	@Override
	public RequestBody convert(T value) throws IOException {

		StringWriter writer = new StringWriter();
        try {
[...]

PR #2609, opened 2018-01-11
Initial JAXB converter factory.

swankjesse:
We probably want to add options to make it strict, or to change the defaults
and make it strict by default and optionally lenient.

We can do that in follow-up. This works.

https://github.com/square/retrofit/issues/2277

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,6 +65,7 @@
     <wire.version>2.2.0</wire.version>
     <simplexml.version>2.7.1</simplexml.version>
     <moshi.version>1.5.0</moshi.version>
+    <jaxb.version>2.2.12</jaxb.version><!-- 2.3.0 breaks due to https://github.com/mojohaus/animal-sniffer/issues/29 -->
 
     <!-- Sample Dependencies -->
     <jsoup.version>1.7.3</jsoup.version>
@@ -156,6 +157,11 @@
         <groupId>com.squareup.moshi</groupId>
         <artifactId>moshi</artifactId>
         <version>${moshi.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>javax.xml.bind</groupId>
+        <artifactId>jaxb-api</artifactId>
+        <version>${jaxb.version}</version>
       </dependency>
       <dependency>
         <groupId>org.scala-lang</groupId>
--- a/retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbConverterFactory.java
+++ b/retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbConverterFactory.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright (C) 2018 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package retrofit2.converter.jaxb;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import javax.annotation.Nullable;
+import javax.xml.bind.JAXBContext;
+import javax.xml.bind.JAXBException;
+import javax.xml.bind.annotation.XmlRootElement;
+import okhttp3.MediaType;
+import okhttp3.RequestBody;
+import okhttp3.ResponseBody;
+import retrofit2.Converter;
+import retrofit2.Retrofit;
+
+/**
+ * A {@linkplain Converter.Factory converter} which uses JAXB for XML. All validation events are
+ * ignored.
+ */
+public final class JaxbConverterFactory extends Converter.Factory {
+  static final MediaType XML = MediaType.parse("application/xml; charset=utf-8");
+
+  /** Create an instance using a default {@link JAXBContext} instance for conversion. */
+  public static JaxbConverterFactory create() {
+    return new JaxbConverterFactory(null);
+  }
+
+  /** Create an instance using {@code context} for conversion. */
+  @SuppressWarnings("ConstantConditions") // Guarding public API nullability.
+  public static JaxbConverterFactory create(JAXBContext context) {
+    if (context == null) throw new NullPointerException("context == null");
+    return new JaxbConverterFactory(context);
+  }
+
+  /** If null, a new JAXB context will be created for each type to be converted. */
+  private final @Nullable JAXBContext context;
+
+  private JaxbConverterFactory(@Nullable JAXBContext context) {
+    this.context = context;
+  }
+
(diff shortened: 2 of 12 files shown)


JakeWharton:
I wrote https://github.com/square/retrofit/compare/jakew/jaxb/2018-01-10 yesterday

pom.xml
```diff
@@ -65,6 +65,7 @@
     <wire.version>2.2.0</wire.version>
     <simplexml.version>2.7.1</simplexml.version>
     <moshi.version>1.5.0</moshi.version>
+    <jaxb.version>2.2.12</jaxb.version><!-- 2.3.0 breaks due to https://github.com/mojohaus/animal-sniffer/issues/29 -->
```
swankjesse: ick

retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbRequestConverter.java
```diff
@@ -0,0 +1,62 @@
...
+      }
+
+      @Override public void writeTo(BufferedSink sink) throws IOException {
+        try {
+          Marshaller marshaller = context.createMarshaller();
+          marshaller.setEventHandler(IGNORE_UNRECOGNIZED_FIELDS);
+
+          XMLStreamWriter xmlWriter = xmlOutputFactory.createXMLStreamWriter(
+              sink.outputStream(), JaxbConverterFactory.XML.charset().name());
+          marshaller.marshal(value, xmlWriter);
```
NightlyNexus: What happens if somebody mutates the value while it's being written out?

NightlyNexus: https://github.com/square/retrofit/compare/jakew/jaxb/2018-01-10#diff-b645eed3acd195172d51a4e5b51a9d7fR49 seems consistent with the other request body converters.

JakeWharton: That's already a problem between `enqueue` and when the converter is called.

That being said, I don't think any other converter streams serialization currently. Is there a reason to prefer this over writing to a `Buffer` and capturing a `ByteString`? It seems like we should be consistent.

swankjesse: Consistent works for me. Lemme fix.

NightlyNexus: >That's already a problem between enqueue and when the converter is called.

Good point! Is there any obstacle to making all of the converters encode directly to the stream, then?

swankjesse: Fixed.

retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbConverterFactory.java
```diff
@@ -0,0 +1,87 @@
...
+    if (type instanceof Class && ((Class<?>) type).isAnnotationPresent(XmlRootElement.class)) {
+      return new JaxbResponseConverter<>(contextForType((Class<?>) type), (Class<?>) type);
+    }
+    return null;
+  }
+
+  private JAXBContext contextForType(Class<?> type) {
+    try {
+      return context != null ? context : JAXBContext.newInstance(type);
+    } catch (JAXBException e) {
```
JakeWharton: JAXB is a perfect example of how _not_ to design your exceptions. Everything is funneled into this single type and it's checked. So gross.

retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/XmlDataException.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package retrofit2.converter.jaxb;
+
+final class XmlDataException extends RuntimeException {
```
JakeWharton: I'm skeptical of the value of this as client code is unable to use it.

retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbConverterFactory.java
```diff
@@ -0,0 +1,87 @@
...
+import okhttp3.ResponseBody;
+import retrofit2.Converter;
+import retrofit2.Retrofit;
+
+/**
+ * A {@linkplain Converter.Factory converter} which uses JAXB for XML. All validation events are
+ * ignored.
+ */
+public final class JaxbConverterFactory extends Converter.Factory {
+  static final ValidationEventHandler IGNORE_UNRECOGNIZED_FIELDS = new ValidationEventHandler() {
```
JakeWharton: Is this really needed? In my tests I saw no adverse effects from including unrecognized fields in either the XML.

retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbRequestConverter.java
```diff
@@ -0,0 +1,55 @@
...
+  JaxbRequestConverter(JAXBContext context, Class<T> type) {
+    this.context = context;
+    this.type = type;
+  }
+
+  @Override public RequestBody convert(final T value) throws IOException {
+    Buffer buffer = new Buffer();
+    try {
+      Marshaller marshaller = context.createMarshaller();
+      marshaller.setEventHandler(IGNORE_UNRECOGNIZED_FIELDS);
```
JakeWharton: This can likely be deleted, right? If we're going from T to RequestBody how can there be unrecognized fields?

retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbConverterFactory.java
swankjesse: I got rid of it. I’ll try to figure out why it was there in the sample I started with. If I find anything, we can expose a lenient mode in follow-up.

retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbConverterFactory.java
swankjesse: Don‘t forget this gem!
```
    /**
     * Vendor specific error code
     *
     */
    private String errorCode;
```

retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbRequestConverter.java
swankjesse: Deleted!

retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/XmlDataException.java
swankjesse: Deleted!

JakeWharton:
👍

PR #2610, opened 2018-01-11
Deprecate our SimpleXmlConverterFactory

swankjesse:


retrofit-converters/simplexml/src/main/java/retrofit2/converter/simplexml/SimpleXmlConverterFactory.java
```diff
@@ -29,6 +29,8 @@
  * <p>
  * This converter only applies for class types. Parameterized types (e.g., {@code List<Foo>}) are
  * not handled.
+ *
+ * @deprecated we recommend switching to the JAXB converter.
  */
 public final class SimpleXmlConverterFactory extends Converter.Factory {
```
JakeWharton: Need `@Deprecated` or the bit won't be set in the class file and classfile-only consumers (almost everyone) won't see it as deprecated.

swankjesse: Untrue.

swankjesse: https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.7.15

swankjesse: Fixed regardless!

JakeWharton: That link refers to the bit. I don't see anything about whether the Javadoc tag sets it.

JakeWharton: Tests show that it does, though.

swankjesse: Yeah. Java had `@deprecated` long before it had annotations. Having both it and`@Deprecated` is redundant (but common!).

PR #2614, opened 2018-01-18
Remove old ProGuard nonsense.

JakeWharton:
As of API 26 the android.jar contains the classes checked by Platform and used by Platform.Java8. And we don't encourage throwing checked exceptions from the service method but instead routing them through your call adapter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,10 +30,16 @@ Retrofit requires at minimum Java 7 or Android 2.3.
 ProGuard
 --------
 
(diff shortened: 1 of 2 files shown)


technoir3:
Without this:
```
-dontwarn retrofit2.Platform$Java8
```
build fails with
```
Warning: retrofit2.Platform$Java8: can't find referenced class org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement
```

PR #2621, opened 2018-01-24
Completable has no body (Void.class), and thus will produce a null in onNext.

lassem:
While running with the akarnokd/RxJava2Extensions that verifies the observable contract; we noticed that Retrofit services that returned `Completable` and got no body from the server would fail.


swankjesse:
The PR prevents a crash, but I don't think anyone would do this? Why would you use a Completable with Void?

JakeWharton:
This is definitely the wrong fix. Can you PR just a failing test case?

JakeWharton:
If you have an optional body then you should be using an abstraction to represent that such as an `Optional` via our Guava or Java 8 converter (or your own). That test isn't a failure of Retrofit but of your service configuration.

lassem:
I can't remember the details off the top of my head, but the body wasn't optional in that sense.
IIRC the service definition was a simple:
```
interface Api {
        @DELETE("/")
        Completable delete();
    }
```
But when running through he akarnokd/RxJava2Extensions it would fail due to calling `onNext(null)`

PR #2627, opened 2018-02-01
Add http response codes

jeziellago:


JakeWharton:
Thanks for the contribution, but we don't want to maintain this list. While HTTP codes aren't added or removed at any unreasonable frequency, they're most often referred to as raw integers and it's awkward doing comparisons against constants for ranges since you don't actually care what the named code is.

PR #2636, opened 2018-02-06
Define Automatic-Module-Name for Retrofit modules

swankjesse:


retrofit-adapters/scala/pom.xml
```diff
@@ -69,6 +69,17 @@
           </signature>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <archive>
+            <manifestEntries>
+              <Automatic-Module-Name>retrofit2.converter.scala</Automatic-Module-Name>
```
swankjesse: I’m continuing a problem: the Scala adapter uses this package: `retrofit2.converter.scala`.

swankjesse: Worth the grind to move it to the right package in follow up?

JakeWharton: It's unreleased, so yes.

swankjesse: https://github.com/square/retrofit/pull/2639

PR #2639, opened 2018-02-06
Rename the Scala adapter package to retrofit2.adapter.scala.

swankjesse:
Previously it was retrofit2.converter.scala.


JakeWharton:
#2637...

PR #2645, opened 2018-02-09
Create the ResponseBody converter lazily.

NightlyNexus:
On an implementation note, I tried adding a separate cache and loading it separately in OkHttpCall, but it was more lines of code changed.
Adding more state to ServiceMethod seems okay.
Leaving the same abstraction internally allows for reinstating early creation/validation if that still needs to be an option.

Downsides to the concept:
  - Inconsistent with creating string and RequestBody converters.
  - Removes early validation. If you're missing a ResponseBody converter, you won't find out until it's time to use one. (See two changed tests.)

Feel free to tell me this is a terrible idea!


NightlyNexus:
Another note: maybe this should be an example instead for those who need it instead of a feature of Retrofit. Anybody could add a deferred converter.

NightlyNexus:
this doesn't need anything from Retrofit, and i don't think everyone wants this.

PR #2658, opened 2018-02-13
Bump Jackson version a bit.

### LINKED ISSUE #2657: Currently used version of jackson-databind contains a security vulnerability
What kind of issue is this?

 - [X] Feature Request. Start by telling us what problem you’re trying to solve. Often a solution

In our exploration of your project we found that it is currently using version 2.7.2 of jackson-databind which is vulnerable to Deserialization of Untrusted Data.  An attacker may exploit this issue by sending a maliciously crafted input to the readValue method of the ObjectMapper.

Recommendation:

   Upgrade the version of jackson-databind in the pom.xml to version 2.7.9.3 or higher to stay in the 2.7 branch or switch to 2.8 or 2.9 branches at 2.8.11.1 or 2.9.4.

For additional details on this vulnerability you can visit the following websites:
Snyk: https://snyk.io/vuln/SNYK-JAVA-COMFASTERXMLJACKSONCORE-31507
Common Vulnerabilities and Exposures (CVE): https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=jackson-databind

JakeWharton:
This avoids a vulnerability on older 2.7 versions.

Figured a minor bump was easier than some arbitrary patch version on 2.7. 2.9 is latest but I don't see much reason to "force" it.

Closes #2657.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,7 +61,7 @@
     <!-- Converter Dependencies -->
     <gson.version>2.8.2</gson.version>
     <protobuf.version>3.0.0</protobuf.version>
-    <jackson.version>2.7.2</jackson.version>
+    <jackson.version>2.8.0</jackson.version>
     <wire.version>2.2.0</wire.version>
     <simplexml.version>2.7.1</simplexml.version>
     <moshi.version>1.5.0</moshi.version>


fredsjones:
I think I need to work on my wording next time the 2.8.0 that you switched to has the same issue, but if you change the 2.8.0 to 2.8.11.1 then you would be good to go.  Sorry for the confusion.

JakeWharton:
Got it. Will fix. We should probably just bump to latest 2.9 then.

PR #2659, opened 2018-02-13
Don't conflate downstream disposal with upstream cancel.

### LINKED ISSUE #2656: RxJava2CallAdapter not propagating request cancel when called on client
I'm trying to cancel all requests on an OkHttpClient while using it in a Retrofit + RxJava2CallAdapter configuration.
After cancelling the request, wrapped as RxJava Single, I don't receive any event in the stream.
Here's a failing test that reproduces the issue: https://gist.github.com/Bxttx/c6e8b7474464e5e7722ca692b4885c72

I guess there is an unhandled case in `CallExecuteObservable`:
```
@Override protected void subscribeActual(Observer<? super Response<T>> observer) {
    // Since Call is a one-shot type, clone it for each new observer.
    Call<T> call = originalCall.clone();
    observer.onSubscribe(new CallDisposable(call));

    boolean terminated = false;
    try {
      Response<T> response = call.execute();
      if (!call.isCanceled()) {
        observer.onNext(response);
      }
      if (!call.isCanceled()) {
        terminated = true;
        observer.onComplete();
      }
      // <----------- else, throw error when canceled ?
    } catch (Throwable t) {
      Exceptions.throwIfFatal(t);
      if (terminated) {
        RxJavaPlugins.onError(t);
      } else if (!call.isCanceled()) {
        try {
          observer.onError(t);
        } catch (Throwable inner) {
          Exceptions.throwIfFatal(inner);
          RxJavaPlugins.onError(new CompositeException(t, inner));
        }
      }
    }
  }
```

JakeWharton:
Since Call instances can be canceled externally, re-using that state for disposed means that some results were not delivered.

Closes #2656.

### DIFF
--- a/retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/CallArbiter.java
+++ b/retrofit-adapters/rxjava/src/main/java/retrofit2/adapter/rxjava/CallArbiter.java
@@ -37,6 +37,7 @@ final class CallArbiter<T> extends AtomicInteger implements Subscription, Produc
   private final Call<T> call;
   private final Subscriber<? super Response<T>> subscriber;
 
+  private volatile boolean unsubscribed;
   private volatile Response<T> response;
 
   CallArbiter(Call<T> call, Subscriber<? super Response<T>> subscriber) {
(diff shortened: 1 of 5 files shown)


JakeWharton:
I'll try to write tests later.

JakeWharton:
With tests now

PR #2660, opened 2018-02-14
Assert full JSON document consumption with Moshi.

NightlyNexus:


retrofit-converters/moshi/src/test/java/retrofit2/converter/moshi/MoshiConverterFactoryTest.java
```diff
@@ -269,4 +277,15 @@
     } catch (IOException expected) {
     }
   }
+
+  @Test public void requireFullConsumption() throws Exception {
+    server.enqueue(new MockResponse().setBody("{}[]"));
```
NightlyNexus: Gson will normally fail already for this, since it does extra buffering, so I don't think Gson's converter needs an option for this.

NightlyNexus:
Open to shorter and more descriptive names for the factory method.

retrofit-converters/moshi/src/main/java/retrofit2/converter/moshi/MoshiResponseBodyConverter.java
```diff
@@ -39,7 +42,12 @@
       if (source.rangeEquals(0, UTF8_BOM)) {
         source.skip(UTF8_BOM.size());
       }
-      return adapter.fromJson(source);
+      JsonReader reader = JsonReader.of(source);
+      T result = adapter.fromJson(reader);
+      if (requireFullConsumption && reader.peek() == JsonReader.Token.END_DOCUMENT) {
+        throw new RuntimeException("JSON document was not fully consumed.");
```
JakeWharton: Hmm should this be a `JsonDataException`?

retrofit-converters/moshi/src/main/java/retrofit2/converter/moshi/MoshiConverterFactory.java
```diff
@@ -52,37 +52,49 @@ public static MoshiConverterFactory create() {
...
     if (moshi == null) throw new NullPointerException("moshi == null");
-    return new MoshiConverterFactory(moshi, false, false, false);
+    return new MoshiConverterFactory(moshi, false, false, false, false);
   }
 
   private final Moshi moshi;
   private final boolean lenient;
   private final boolean failOnUnknown;
   private final boolean serializeNulls;
+  private final boolean requireFullResponseConsumption;
```
JakeWharton: I'm not sure we need a knob for this...

retrofit-converters/moshi/src/main/java/retrofit2/converter/moshi/MoshiResponseBodyConverter.java
NightlyNexus: yes.

retrofit-converters/moshi/src/main/java/retrofit2/converter/moshi/MoshiConverterFactory.java
NightlyNexus: yeah. seems unlikely to break any current users.

swankjesse:
I think we also want support on JsonAdapter for when the input is a String.

swankjesse:
(in Moshi itself!)

PR #2664, opened 2018-02-17
More accurately name call adapter list field.

JakeWharton:


retrofit/src/main/java/retrofit2/Retrofit.java
```diff
@@ -583,7 +586,7 @@ public Retrofit build() {
       }
 
       // Make a defensive copy of the adapters and add the default Call adapter.
-      List<CallAdapter.Factory> adapterFactories = new ArrayList<>(this.adapterFactories);
+      List<CallAdapter.Factory> adapterFactories = new ArrayList<>(this.callAdapterFactories);
```
NightlyNexus: Want to rename this, too?

JakeWharton: #2667

PR #2685, opened 2018-03-07
Small tweaks while doing other work.

JakeWharton:


retrofit/src/main/java/retrofit2/Utils.java
```diff
@@ -82,7 +82,9 @@ static boolean equals(Type a, Type b) {
       if (!(b instanceof ParameterizedType)) return false;
       ParameterizedType pa = (ParameterizedType) a;
       ParameterizedType pb = (ParameterizedType) b;
-      return equal(pa.getOwnerType(), pb.getOwnerType())
+      Object ownerA = pa.getOwnerType();
+      Object ownerB = pb.getOwnerType();
+      return (ownerA == ownerB || (ownerA != null && ownerA.equals(ownerB)))
```
NightlyNexus: could be `equals(pa.getOwnerType(), pb.getOwnerType())` (one character change).
also avoids Type.equals usage, though i highly doubt there would ever be real case of a broken owner type.

retrofit/src/main/java/retrofit2/OkHttpCall.java
```diff
@@ -110,16 +110,20 @@
...
         } catch (Throwable e) {
           callFailure(e);
           return;
         }
-        callSuccess(response);
+
+        try {
+          callback.onResponse(OkHttpCall.this, response);
+        } catch (Throwable t) {
+          t.printStackTrace();
```
swankjesse: this doesn’t seem right

swankjesse: oh, but it’s what we’ve always done? Sigh

JakeWharton: Same as OkHttp. When the callback fails we have nowhere to go.

PR #2715, opened 2018-03-26
Ability to add Http status code to Response.success

joshkleven09:
When writing tests using Retrofit, it would be great if there were a way to set the "success" Http status code in order to verify functionality of other status codes, like 204 no content.

### DIFF
--- a/retrofit/src/main/java/retrofit2/Response.java
+++ b/retrofit/src/main/java/retrofit2/Response.java
@@ -44,11 +44,11 @@ public final class Response<T> {
       throw new IllegalArgumentException("code < 200 or >= 300: " + code);
     }
     return success(body, new okhttp3.Response.Builder() //
-            .code(code)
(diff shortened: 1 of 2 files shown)


retrofit/src/main/java/retrofit2/Response.java
```diff
@@ -35,6 +35,20 @@
         .build());
   }
 
+  /**
+   * Create a synthetic successful response with an HTTP status code of {@code code} and
+   * {@code body} as the deserialized body.
+   */
+  public static <T> Response<T> success(int code, @Nullable T body) {
+    if (code < 200 || code >= 300) throw new IllegalArgumentException("code < 200 or >= 300: " + code);
```
JakeWharton: Line too long. Please add `{` and `}` when wrapping.

JakeWharton:
Thanks!

PR #2717, opened 2018-03-27
Cache default method's MethodHandler to avoid large LambdaForm loading.

kojilin:
We met lots of class loading/unloading problem when using default method. After each young gc, jvm  unload LambdaForm and re-load new one when we call method again. 
When we met mix gc, it needs some time to unloading classes.

### DIFF
--- a/retrofit/src/main/java/retrofit2/Platform.java
+++ b/retrofit/src/main/java/retrofit2/Platform.java
@@ -18,9 +18,12 @@ package retrofit2;
 import android.os.Build;
 import android.os.Handler;
 import android.os.Looper;
+import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles.Lookup;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
+import java.util.Map;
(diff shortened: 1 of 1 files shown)


retrofit/src/main/java/retrofit2/Platform.java
```diff
@@ -70,20 +73,74 @@ boolean isDefaultMethod(Method method) {
...
         @Nullable Object... args) throws Throwable {
-      // Because the service interface might not be public, we need to use a MethodHandle lookup
-      // that ignores the visibility of the declaringClass.
-      Constructor<Lookup> constructor = Lookup.class.getDeclaredConstructor(Class.class, int.class);
-      constructor.setAccessible(true);
-      return constructor.newInstance(declaringClass, -1 /* trusted */)
-          .unreflectSpecial(method, declaringClass)
-          .bindTo(object)
-          .invokeWithArguments(args);
+      CacheKey key = new CacheKey(method, declaringClass, object);
```
kojilin: maybe just method is enough for key?

kojilin:
ah, let me create issue first.

PR #2721, opened 2018-03-28
Makes sure observable assembly hooks are called

### LINKED ISSUE #2716: Facilitate trace propagation with RxJava2
What kind of issue is this?

I'm trying to help get tracing working for retrofit. For example, continuing an incoming trace during an retrofit outbound request (including but not limited to SLF4J propagation).

I looked at similar tech https://github.com/akaita/RxJava2Debug and assembly hooks look quite good for this. I started testing and noticed everything works nicely, except when the result type is Observable. That's because observable types are returned as-is:

```java
final class RxJava2CallAdapter<R> implements CallAdapter<R, Object> {
 --snip--
    if (isCompletable) {
      return observable.ignoreElements(); // .ignoreElements indirectly calls RxJavaPlugins.onAssembly
    }
    return observable; // returned with no hooks
  }
```

I understand that a custom adapter could do this, but wondering if we could make it default to use the hook on plain observable.

ex.
```java
final class RxJava2CallAdapter<R> implements CallAdapter<R, Object> {
 --snip--
    if (isCompletable) {
      return observable.ignoreElements();
    }
    return RxJavaPlugins.onAssembly(observable);
  }
```

Open to other ideas to make this drop-in

adriancole:
This allows things like error and trace tracking to work naturally.

Fixes #2716

### DIFF
--- a/retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/RxJava2CallAdapter.java
+++ b/retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/RxJava2CallAdapter.java
@@ -18,6 +18,7 @@ package retrofit2.adapter.rxjava2;
 import io.reactivex.BackpressureStrategy;
 import io.reactivex.Observable;
 import io.reactivex.Scheduler;
+import io.reactivex.plugins.RxJavaPlugins;
 import java.lang.reflect.Type;
 import javax.annotation.Nullable;
 import retrofit2.Call;
@@ -83,6 +84,6 @@ final class RxJava2CallAdapter<R> implements CallAdapter<R, Object> {
     if (isCompletable) {
       return observable.ignoreElements();
     }
-    return observable;
+    return RxJavaPlugins.onAssembly(observable);
   }
 }
--- a/retrofit-adapters/rxjava2/src/test/java/retrofit2/adapter/rxjava2/ObservableTest.java
+++ b/retrofit-adapters/rxjava2/src/test/java/retrofit2/adapter/rxjava2/ObservableTest.java
@@ -16,6 +16,8 @@
 package retrofit2.adapter.rxjava2;
 
 import io.reactivex.Observable;
+import io.reactivex.functions.Function;
+import io.reactivex.plugins.RxJavaPlugins;
 import java.io.IOException;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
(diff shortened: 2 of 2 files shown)


retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/RxJava2CallAdapter.java
```diff
@@ -83,6 +84,6 @@
     if (isCompletable) {
       return observable.ignoreElements();
     }
-    return observable;
+    return RxJavaPlugins.onAssembly(observable);
```
artem-zinnatullin: Hm, I thought only custom observable instantiation will be wrapped:

```java
Observable<Response<R>> responseObservable = RxJavaPlugins.onAssembly(isAsync
        ? new CallEnqueueObservable<>(call)
        : new CallExecuteObservable<>(call));
```

```java
if (isResult) {
  observable = RxJavaPlugins.onAssembly(new ResultObservable<>(responseObservable));
} else if (isBody) {
  observable = RxJavaPlugins.onAssembly(new BodyObservable<>(responseObservable));
} else {
  observable = responseObservable;
}
```

Current implementation is not bad, but:

- As far as I see stactrace won't point to specific lines where `CallEnqueueObservable`/`CallExecuteObservable`/`ResultObservable`/`BodyObservable` got intantiated but will only point to line 87 where it can be hard to figure which branch of code was executed and what is the original observable type if `subscribeOn` was applied
- Wrapping happens after potential `subscribeOn`. Retrofit Observable types are package-private which means that reflecting on their internals is not part of public api provided by Retrofit, however it seems legal to check for `package` of the Observable type to apply some custom behavior. Since wrapping happens after `subscribeOn`, type information gets hidden.

artem-zinnatullin: I've made a branch with more fine-grained wrapping that we can turn into PR if you want 

https://github.com/square/retrofit/compare/master...artem-zinnatullin:az/rxjava-adapter-onAssembly-fix?expand=1

JakeWharton:
I don't see any reason to wrap at that granularity. We can wrap once at
line 70 I think.

artem-zinnatullin:
My only objection against wrapping on line 70 is to get more precise stacktrace where it'll be clear which code path in the adapter was executed for say crash reporting

JakeWharton:
I'm hesitant about calling it twice because it's an implementation detail
and I don't want anyone to rely on how we call out.

artem-zinnatullin:
I guess it might be helpful for dealing with users' issues on SO or here if stacktraces will be more precise

Issue #2693, opened 2018-03-15
Support for retrofit2.mock.Calls.failure() to accept exceptions types other than IOException


cpeppas:
- [x] Feature Request. 

I have implemented unit tests for my api using the MockRetrofit and BehaviorDelegate

So far I am able to assert success 200, Http error (400,500) and Network (IoException) error cases.

I would like also to be able to assert error cases such as NullPointerException (mainly parsing issues that can be possible thrown) or any other Exception other than IOException

Example

```
    private class MockApi implements MyApi {
        final BehaviorDelegate<MyApi> delegate;

        MockApi(BehaviorDelegate<MyApi> delegate) {
            this.delegate = delegate;
        }

        @Override public Single<MyResponse> retrieveStuff(String param) {
            return delegate.returning(Calls.failure(new MyException())).retrieveStuff(param);
        }
    }
```

JakeWharton:
PR welcome!

naturalwarren:
I'd love to help here if there's any interest. @cpeppas Is this something you're already working on?

cpeppas:
Thanks @naturalwarren I haven't started anything yet so feel free 👍

JakeWharton:
Merged as ce787ed5.

PR #2725, opened 2018-04-01
Add support for creating failed test call instances with any exception.

naturalwarren:
When using `Calls.failure()` to create a failed `Call` object for test you can now pass in any `Throwable`.

If the `Throwable` is a `RuntimeException`, `Error` or `IOException` subtype it is thrown with a cast. Otherwise the `Throwable` is "sneaky thrown".

Addresses #2693

### DIFF
--- a/retrofit-mock/src/main/java/retrofit2/mock/Calls.java
+++ b/retrofit-mock/src/main/java/retrofit2/mock/Calls.java
@@ -47,9 +47,10 @@ public final class Calls {
   }
 
   /**
-   * Creates a failed {@link Call} from the provided {@code failure}. If {@code failure} is a {@link RuntimeException},
-   * {@link Error} or {@link IOException} subtype it is thrown with a cast. Otherwise the {@code failure} is
-   * "sneaky thrown" via {@linkplain Calls#sneakyThrow(Throwable)}.
+   * Creates a failed {@link Call} from the provided {@code failure}. If {@code failure} is a
+   * {@link RuntimeException}, {@link Error} or {@link IOException} subtype it is thrown with a
+   * cast. Otherwise the {@code failure} is "sneaky thrown" via
+   * {@linkplain Calls#sneakyThrow(Throwable)}.
    */
   public static <T> Call<T> failure(Throwable failure) {
     return new FakeCall<>(null, failure);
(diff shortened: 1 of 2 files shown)


retrofit-mock/src/main/java/retrofit2/mock/Calls.java
```diff
@@ -42,7 +42,7 @@
     return new FakeCall<>(response, null);
   }
 
-  public static <T> Call<T> failure(IOException failure) {
+  public static <T> Call<T> failure(Exception failure) {
```
JakeWharton: This is a backwards incompatible change and needs to be introduced as an overload. Also it probably should take `Throwable`.

retrofit-mock/src/main/java/retrofit2/mock/Calls.java
```diff
@@ -74,7 +74,11 @@ private Calls() {
       if (response != null) {
         return response;
       }
-      throw error;
+      if (error instanceof IOException) {
+        throw (IOException) error;
+      } else {
+        throw new IOException(error);
```
JakeWharton: I don't think this is what we want because it changes the semantics of the error. If it's a `RuntimeException` subtype we can throw it directly with a cast, if it's an `Error` subtype we can throw it directly with a cast`, and if it's not one of those (and not an `IOException`) then we probably need to [sneaky throw][1] it. And then this behavior needs documented on the factory method.

 [1]: http://www.proofbyexample.com/sneakythrow-avoid-checked-exceptions.html

naturalwarren: Whoa! "sneaky throw" that's awesome! Done in a98b705.

retrofit-mock/src/main/java/retrofit2/mock/Calls.java
naturalwarren: Probably nicer as an overload (did this as part of a98b705) but this is still backwards compatible as written right? Tests didn't fail on us.

naturalwarren:
@JakeWharton Would appreciate another look when you can.

JakeWharton:
Merged as ce787ed5. Thanks!

PR #2737, opened 2018-04-20
Remove synthetic accessors.

Apsaliya:
From IJ inspection.


JakeWharton:
Thanks for the contribution, but these are all in test code where we don't care about the accessors as much since they only affect Android applications which consume the library.

PR #2750, opened 2018-05-01
Update README.md

mosamabinomar:
`implementation` instead of `compile`


JakeWharton:
Please update every README if you are going to make this change. There's about 20 of them.

JakeWharton:
Fixed by #2816

PR #2771, opened 2018-05-31
Upgrade RxJava2 to 2.1.14, Upgrade AssertJ to 2.9.1

bednar:
This Pull request also cover the RxJava2 API changes between 2.0.0 and 2.1.14

-  2.0.6: introduces specific exception wrappers to help distinguish and track down what was happening the time of the error = exceptions are wrapped in UndeliverableException - [additional info](https://github.com/ReactiveX/RxJava/wiki/What%27s-different-in-2.0#error-handling)
- 2.0.7: a non-positive Subscription.request() will not stop the current stream but signal an error via RxJavaPlugins.onError  - [additional info](https://github.com/ReactiveX/RxJava/wiki/What%27s-different-in-2.0#reactive-streams-compliance)

Hi @JakeWharton, it is correct change for you?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -54,7 +54,7 @@
 
     <!-- Adapter Dependencies -->
     <rxjava.version>1.3.0</rxjava.version>
-    <rxjava2.version>2.0.0</rxjava2.version>
+    <rxjava2.version>2.1.14</rxjava2.version>
     <guava.version>19.0</guava.version>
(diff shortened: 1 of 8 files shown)


jabson009:
Good one

JakeWharton:
Thanks, I landed a derivative of this change in #3636 which also removes the runtime backpressure support and completely removes the associated test cases.

PR #2797, opened 2018-06-15
Fix pointer to renamed rules file

JakeWharton:


JakeWharton:
CI was file system flake

PR #2799, opened 2018-06-16
Consolidate exception formatters in one place

JakeWharton:


JakeWharton:
@swankjesse Friendly ping!

PR #2805, opened 2018-06-24
RxJava: Skip enqueuing calls if disposed

ansman:
This is a minor optimization but one that can avoid needless network calls.

Consider for example the following code:
```
myService
  .getStuff()
  .toObservable()
  .compose(ReplayingShare.instance())
  .firstOrError()
```

`RxReplaying` caches values and emits them in `onSubscribe`. If there is a cached value this will enqueue or worst case fully perform the network request even though downstream has been disposed.

### DIFF
--- a/retrofit-adapters/rxjava2/src/test/java/retrofit2/adapter/rxjava2/CancelDisposeTestSync.java
+++ b/retrofit-adapters/rxjava2/src/test/java/retrofit2/adapter/rxjava2/CancelDisposeTestSync.java
@@ -17,9 +17,9 @@ package retrofit2.adapter.rxjava2;
 
 import io.reactivex.Observable;
 import io.reactivex.disposables.Disposable;
-import java.util.List;
 import okhttp3.Call;
 import okhttp3.OkHttpClient;
+import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
 import org.junit.Before;
 import org.junit.Rule;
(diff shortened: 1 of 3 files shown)


JakeWharton:
Can we add a test for this behavior?

ansman:
Sure, I’ll add one for both factories

ansman:
Done. Both tests failed before the patch and pass after.

retrofit-adapters/rxjava2/src/test/java/retrofit2/adapter/rxjava2/CancelDisposeTestSync.java
```diff
@@ -0,0 +1,60 @@
...
+        .baseUrl(server.url("/"))
+        .addConverterFactory(new StringConverterFactory())
+        .addCallAdapterFactory(RxJava2CallAdapterFactory.create())
+        .callFactory(client)
+        .build();
+    service = retrofit.create(Service.class);
+  }
+
+  @Test public void disposeBeforeExecuteDoesNotEnqueue() {
+    server.enqueue(new MockResponse().setBody("Hi"));
```
ansman: This is done just so that we don't block the main thread for 10s if the test were to fail.

JakeWharton: I've removed this. I don't think it's worth optimizing for what happens should a test start failing.

retrofit-adapters/rxjava2/src/test/java/retrofit2/adapter/rxjava2/CancelDisposeTestSync.java
```diff
@@ -0,0 +1,60 @@
...
+import okhttp3.mockwebserver.MockWebServer;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import retrofit2.Retrofit;
+import retrofit2.http.GET;
+
+import java.util.List;
+
+import static org.junit.Assert.*;
```
JakeWharton: star imports are not allowed

retrofit-adapters/rxjava2/src/test/java/retrofit2/adapter/rxjava2/CancelDisposeTestSync.java
ansman: Fair enough

PR #2827, opened 2018-07-13
Update to OkHttp 3.11

JakeWharton:
Use get instead of parse for HttpUrl and MediaType where valid input is always required.

### DIFF
--- a/retrofit/src/main/java/retrofit2/RequestBuilder.java
+++ b/retrofit/src/main/java/retrofit2/RequestBuilder.java
@@ -76,11 +76,11 @@ final class RequestBuilder {
 
   void addHeader(String name, String value) {
     if ("Content-Type".equalsIgnoreCase(name)) {
-      MediaType type = MediaType.parse(value);
-      if (type == null) {
-        throw new IllegalArgumentException("Malformed content type: " + value);
+      try {
+        contentType = MediaType.get(value);
+      } catch (IllegalArgumentException e) {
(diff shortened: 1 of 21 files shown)


retrofit/src/main/java/retrofit2/RequestBuilder.java
```diff
@@ -76,11 +76,11 @@ void setRelativeUrl(Object relativeUrl) {
 
   void addHeader(String name, String value) {
     if ("Content-Type".equalsIgnoreCase(name)) {
-      MediaType type = MediaType.parse(value);
-      if (type == null) {
-        throw new IllegalArgumentException("Malformed content type: " + value);
+      try {
+        contentType = MediaType.get(value);
+      } catch (IllegalArgumentException e) {
+        throw new IllegalArgumentException("Malformed content type: " + value, e);
```
swankjesse: Necessary?

JakeWharton: Kinda, yeah. I had removed the try/catch here and around the `HttpUrl.get` initially, but the exception message you get isn't very helpful as it lacks context. It just says something like "Invalid subtype: blah" or similar. It tells you very little about what you did wrong unless you look very closely at the stacktrace and infer what it's doing. Adding these back gives you "Malformed content type: foo/bar; baz" with a cause of "Invalid subtype: baz" which makes a lot more sense.

swankjesse: 👍🏻

PR #2859, opened 2018-08-16
Add built-in support for Kotlin's Unit type

### LINKED ISSUE #2329: Add Kotlin Unit Converter

JakeWharton:
This is built in a way such that its absence from the classpath will not cause failures.

Closes #2329.

### DIFF
--- a/retrofit/src/main/java/retrofit2/BuiltInConverters.java
+++ b/retrofit/src/main/java/retrofit2/BuiltInConverters.java
@@ -18,11 +18,15 @@ package retrofit2;
 import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
+import kotlin.Unit;
 import okhttp3.RequestBody;
 import okhttp3.ResponseBody;
 import retrofit2.http.Streaming;
 
 final class BuiltInConverters extends Converter.Factory {
+  /** Not volatile because we don't mind multiple threads discovering this. */
+  private boolean checkForKotlinUnit = true;
+
   @Override
   public Converter<ResponseBody, ?> responseBodyConverter(Type type, Annotation[] annotations,
       Retrofit retrofit) {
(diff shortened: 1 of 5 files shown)


retrofit/src/test/java/retrofit2/KotlinUnitTest.java
```diff
@@ -0,0 +1,53 @@
...
+    Call<Unit> empty();
+  }
+
+  @Test public void unitOnClasspath() throws IOException {
+    Retrofit retrofit = new Retrofit.Builder()
+        .baseUrl(server.url("/"))
+        .build();
+    Service example = retrofit.create(Service.class);
+
+    server.enqueue(new MockResponse().setBody("Hi"));
```
NightlyNexus: i guess there's no easy way to demonstrate that the body isn't read.

JakeWharton: And technically it _is_ read by OkHttp internally so that the connection can be re-used. But yeah there's no good way to validate it. We don't for Void either.

retrofit/src/main/java/retrofit2/BuiltInConverters.java
```diff
@@ -18,11 +18,15 @@
...
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
+import kotlin.Unit;
 import okhttp3.RequestBody;
 import okhttp3.ResponseBody;
 import retrofit2.http.Streaming;
 
 final class BuiltInConverters extends Converter.Factory {
+  /** Not volatile because we don't mind multiple threads discovering this. */
+  private boolean checkForKotlinUnit = true;
```
artem-zinnatullin: Make it `static`?

retrofit/src/main/java/retrofit2/BuiltInConverters.java
```diff
@@ -34,6 +38,15 @@
     if (type == Void.class) {
       return VoidResponseBodyConverter.INSTANCE;
     }
+    if (checkForKotlinUnit) {
+      try {
+        if (type == Unit.class) {
+          return UnitResponseBodyConverter.INSTANCE;
+        }
+      } catch (NoClassDefFoundError ignored) {
```
naturalwarren: Whoa, love it.

trevjonez:
@JakeWharton need to add an update to the [proguard config](https://github.com/square/retrofit/blob/master/retrofit/src/main/resources/META-INF/proguard/retrofit2.pro) as well right? I am thinking if `kotlin.Unit` isn't on the path downstream it will scream about it.

JakeWharton:
Sure, we can add a `-dontwarn kotlin.Unit`.

PR #2866, opened 2018-08-23
use retrofit2.httpException instead

MohamedISoliman:
remove deprecated HttpException class, use the retrofit2.httpException instead


JakeWharton:
This exception is retained for backwards compatibility at a binary and source level. We cannot remove it until Retrofit 3.0.

PR #2878, opened 2018-09-06
Re-work unresolvable type detection in parameters

JakeWharton:
In order to natively support Kotlin's 'suspend' modifier in the future, we need to allow wildcards to be used for the last parameter. As a result, we only validate the shape of a type once a Retrofit annotation is detected.

### DIFF
--- a/retrofit/src/main/java/retrofit2/RequestFactory.java
+++ b/retrofit/src/main/java/retrofit2/RequestFactory.java
@@ -171,18 +171,7 @@ final class RequestFactory {
       int parameterCount = parameterAnnotationsArray.length;
       parameterHandlers = new ParameterHandler<?>[parameterCount];
       for (int p = 0; p < parameterCount; p++) {
-        Type parameterType = parameterTypes[p];
-        if (Utils.hasUnresolvableType(parameterType)) {
-          throw parameterError(method, p,
-              "Parameter type must not include a type variable or wildcard: %s", parameterType);
-        }
-
-        Annotation[] parameterAnnotations = parameterAnnotationsArray[p];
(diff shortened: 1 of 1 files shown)


retrofit/src/main/java/retrofit2/RequestFactory.java
```diff
@@ -680,6 +683,7 @@ private Headers parseHeaders(String[] headers) {
         return new ParameterHandler.PartMap<>(valueConverter, partMap.encoding());
 
       } else if (annotation instanceof Body) {
+        validateResolvableType(p, type);
```
JakeWharton: Having this in every branch sucks and is error-prone but we need to do the validation before running other code (like calling out to resolve converters)

retrofit/src/main/java/retrofit2/RequestFactory.java
```diff
@@ -171,18 +171,7 @@ RequestFactory build() {
...
       for (int p = 0; p < parameterCount; p++) {
-        Type parameterType = parameterTypes[p];
-        if (Utils.hasUnresolvableType(parameterType)) {
-          throw parameterError(method, p,
-              "Parameter type must not include a type variable or wildcard: %s", parameterType);
-        }
-
-        Annotation[] parameterAnnotations = parameterAnnotationsArray[p];
-        if (parameterAnnotations == null) {
-          throw parameterError(method, p, "No Retrofit annotation found.");
```
NightlyNexus: neat, still handled in `parseParameter` for now. 🎉

JakeWharton:
@swankjesse ping!

PR #2888, opened 2018-09-13
Fix `Unit` converter not being called on a 204 response.

nabrozidhs:
This is probably very wrong but I thought it would be interesting to put it up for comment given I didn't see anyone else running into this issue.

I'm using retrofit + kotlin coroutines and all current coroutines adapters rely on successful responses to not have a `null` body. This becomes an issue when you try to deserialise an empty response. Recently a `Unit` converter was [added](https://github.com/square/retrofit/commit/7cbfa055d7f69317bfd33e5a5b206e60cfff3347) but it fails if the backend response is a `204` (and by digging into retrofit code also a `205` code).

This is just a proposed quick fix with a failing test added on `KotlinUnitTest` but I imagine the implications are quite big and will have a hit on performance.

I guess my argument is that if a succesful response on a given api call would be a `204` I would expect an interface like the following one to work fine.

```kotlin
interface Service {
  @POST("event")
  fun sendEvent(...): Call<Unit>
}
```

### DIFF
--- a/retrofit/src/main/java/retrofit2/OkHttpCall.java
+++ b/retrofit/src/main/java/retrofit2/OkHttpCall.java
@@ -213,11 +213,6 @@ final class OkHttpCall<T> implements Call<T> {
       }
     }
 
-    if (code == 204 || code == 205) {
-      rawBody.close();
-      return Response.success(null, rawResponse);
-    }
-
     ExceptionCatchingResponseBody catchingBody = new ExceptionCatchingResponseBody(rawBody);
     try {
       T body = responseConverter.convert(catchingBody);
--- a/retrofit/src/test/java/retrofit2/CallTest.java
+++ b/retrofit/src/test/java/retrofit2/CallTest.java
@@ -47,8 +47,6 @@ import static okhttp3.mockwebserver.SocketPolicy.DISCONNECT_DURING_RESPONSE_BODY
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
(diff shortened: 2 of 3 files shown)


JakeWharton:
This is a very backwards incompatible change as far as converter implementations are concerned. I'm not sure that I see the problem here, though. A 204 and 205 have no body so there is nothing to convert and as a result `null` is really our only option here. Beyond that, `Response`'s `body()` accessor is marked as `@Nullable` to promote this onto callers.

nabrozidhs:
Don't take me wrong I'm not arguing in favour of this, just trying to understand what kind of retrofit interface would you use for this case.

Before this I was using `Competable` as the return type and that worked fine. I kinda got the wrong assumption by `Call<Unit>`, that's all :sweat_smile:

JakeWharton:
`Call<Unit>` creates a `Response<Unit>` which has a `@Nullable T body()` so `null` is always allowed. In a 204/205, there is no body and so calling the converter is weird. This is in direct contrast to a 200 which has a 0-byte body where we do call the converter. The HTTP responses here have the same 0-byte body but the semantics of the response code alter the behavior. The converter gets to decide if a 0-byte body is valid only in the case of a response that contains a body.

The problem here really occurs when the consumer defines something like `Observable<Unit>` where the body is propagated to a different container where the `Response` is not available. Here the `null` in a 204/205 case needs to be handled in a call adapter, not a converter. This is because the call adapter is mapping different semantics onto the body. In this example, `Observable` can't actually propagate `null` and so it would need to decide what to do. Right now it does nothing and lets the stream error.

For 3.x we probably want to look at this more closely.

PR #2892, opened 2018-09-19
Normalize null args to empty

JakeWharton:
There is no reason to propagate the pain of nullability throughout multiple internal types.


swankjesse:
Ugh, I think I stomped all over this with https://github.com/square/retrofit/pull/2899

If we punt or defer that we should merge this.

swankjesse:
Gonna rebase over this since it’ll make coroutines easier.

Issue #2893, opened 2018-09-19
Move Java8 converter and adapter into the core


JakeWharton:
These should be progressive enhancements based on availability rather than requiring explicit registration and a separate dependency.

PR #2895, opened 2018-09-19
Move CompletableFuture support into Retrofit core.

NightlyNexus:
the adapter part of #2893.

### DIFF
--- a/retrofit/src/main/java/retrofit2/CompletableFutureCallAdapterFactory.java
+++ b/retrofit/src/main/java/retrofit2/CompletableFutureCallAdapterFactory.java
@@ -13,23 +13,16 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package retrofit2.adapter.java8;
+package retrofit2;
 
 import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.concurrent.CompletableFuture;
-import retrofit2.Call;
-import retrofit2.CallAdapter;
-import retrofit2.Callback;
-import retrofit2.Response;
-import retrofit2.Retrofit;
+import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement;
 
 /**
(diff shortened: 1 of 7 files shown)


retrofit/src/main/java/retrofit2/Retrofit.java
```diff
@@ -64,16 +64,19 @@
   final HttpUrl baseUrl;
   final List<Converter.Factory> converterFactories;
   final List<CallAdapter.Factory> callAdapterFactories;
+  final int userAddedCallAdapterFactoryCount;
```
NightlyNexus: ugh, i can't think of a better way to do this.

retrofit/src/main/java/retrofit2/CompletableFutureCallAdapterFactory.java
```diff
@@ -0,0 +1,151 @@
...
+ * errors</li>
+ * </ul>
+ */
+@IgnoreJRERequirement
+final class CompletableFutureCallAdapterFactory extends CallAdapter.Factory {
+  static final CallAdapter.Factory INSTANCE = new CompletableFutureCallAdapterFactory();
+
+  @Override
+  public CallAdapter<?, ?> get(Type returnType, Annotation[] annotations, Retrofit retrofit) {
+    if (getRawType(returnType) != CompletableFuture.class) {
```
swankjesse: Won’t this bomb out on Java 7?

NightlyNexus: the factory is added only on Java 8 (from Platform.defaultCallAdapterFactories).

retrofit/src/main/java/retrofit2/Retrofit.java
swankjesse: maybe have the platform tell you how many callAdapterFactories it produces? Then you keep this state once in Platform rather than in each instance of `Retrofit`

NightlyNexus: yeah 👍

PR #2899, opened 2018-09-21
New 'Invocation' class for tracking metrics on Retrofit calls

swankjesse:


retrofit/src/main/java/retrofit2/Call.java
```diff
@@ -70,4 +70,7 @@
 
   /** The original HTTP request. */
   Request request();
+
+  /** The service method invocation that created this call. */
+  Invocation invocation();
```
JakeWharton: I can't think of a reason to expose this here. In my version I only had it added as a request tag. You can still access it from here by calling `request().tag(Invocation.class)` although I expect no one to do that.

retrofit/src/main/java/retrofit2/Retrofit.java
```diff
@@ -143,7 +145,11 @@
             if (platform.isDefaultMethod(method)) {
               return platform.invokeDefaultMethod(method, service, proxy, args);
             }
-            return loadServiceMethod(method).invoke(args);
+            List<Object> argsList = args != null
+                ? Arrays.asList(args)
+                : Collections.emptyList();
+            Invocation invocation = new Invocation(method, argsList);
```
JakeWharton: We need to push this creation farther in so that we can remove the `Continuation` argument from #2886. I don't think we want to reference that instance from the request since it could leak objects if you retain a response object for longer than we expect.

samples/src/main/java/com/example/retrofit/InvocationMetrics.java
```diff
@@ -0,0 +1,87 @@
...
+
+  static final class InvocationLogger implements Interceptor {
+    @Override public Response intercept(Chain chain) throws IOException {
+      Request request = chain.request();
+      long startNanos = System.nanoTime();
+      Response response = chain.proceed(request);
+      long elapsedNanos = System.nanoTime() - startNanos;
+
+      Invocation invocation = request.tag(Invocation.class);
+      System.out.printf("%s.%s %s HTTP %s (%.0f ms)%n",
```
JakeWharton: Null check? You have it in the Javadoc but this is far more likely to be copy/pasted

retrofit/src/main/java/retrofit2/Call.java
swankjesse: SGTM!

samples/src/main/java/com/example/retrofit/InvocationMetrics.java
swankjesse: Done!

retrofit/src/main/java/retrofit2/Retrofit.java
swankjesse: Yeah, I was anticipating we’d strip that out, but wasn’t sure how/where. Fixed to make that easier when it lands.

PR #2902, opened 2018-09-22
Expose static factory for Invocation

JakeWharton:
Also rework the internals of how an Invocation gets instantiated and set as a tag on the Request. This will make a future change easier to land.

@swankjesse This makes #2886 simpler to integrate.

### DIFF
--- a/retrofit/src/main/java/retrofit2/Invocation.java
+++ b/retrofit/src/main/java/retrofit2/Invocation.java
@@ -50,15 +50,19 @@ import static retrofit2.Utils.checkNotNull;
  * types for parameters!
  */
 public final class Invocation {
+  public static Invocation of(Method method, List<?> arguments) {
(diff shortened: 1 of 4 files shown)


retrofit/src/main/java/retrofit2/Invocation.java
```diff
@@ -50,15 +50,19 @@
...
+  public static Invocation of(Method method, List<?> arguments) {
     checkNotNull(method, "method == null");
     checkNotNull(arguments, "arguments == null");
+    return new Invocation(method, new ArrayList<>(arguments)); // Defensive copy.
+  }
+
+  private final Method method;
+  private final List<?> arguments;
 
+  /** Trusted constructor assumes ownership of {@code arguments}. */
```
swankjesse: Nice

PR #2905, opened 2018-09-24
Move Optional support into Retrofit core.

### LINKED ISSUE #2893: Move Java8 converter and adapter into the core

NightlyNexus:
Closes #2893.

### DIFF
--- a/retrofit/src/main/java/retrofit2/OptionalConverterFactory.java
+++ b/retrofit/src/main/java/retrofit2/OptionalConverterFactory.java
@@ -13,43 +13,47 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package retrofit.converter.java8;
+package retrofit2;
 
+import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.Optional;
 import javax.annotation.Nullable;
 import okhttp3.ResponseBody;
-import retrofit2.Converter;
-import retrofit2.Retrofit;
+import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement;
 
 /**
- * @deprecated Retrofit includes support for Optional. This no longer needs to be added to the
- * Retrofit instance explicitly.
- * <p>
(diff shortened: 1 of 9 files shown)


retrofit/src/test/java/retrofit2/OptionalConverterFactoryTest.java
```diff
@@ -0,0 +1,64 @@
+package retrofit2;
```
swankjesse: nit: copyright

retrofit/src/main/java/retrofit2/OptionalConverterFactory.java
```diff
@@ -0,0 +1,45 @@
+package retrofit2;
```
swankjesse: nit: copyright

retrofit/src/main/java/retrofit2/OptionalConverterFactory.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * A {@linkplain Converter.Factory converter} for {@code Optional<T>} which delegates to another
+ * converter to deserialize {@code T} and then wraps it into {@link Optional}.
+ */
+@IgnoreJRERequirement
+final class OptionalConverterFactory extends Converter.Factory {
+  static final Converter.Factory INSTANCE = new OptionalConverterFactory();
+
+  @Nullable @Override
```
swankjesse: nit: put `@Nullable` immediately preceding the return type, to hint that we interpret it as a type annotation:
```
public @Nullable Converter<ResponseBody, ?> responseBodyConverter(
```

retrofit/src/test/java/retrofit2/OptionalConverterFactoryTest.java
NightlyNexus: Whoops. Thanks for these catches.

retrofit/src/main/java/retrofit2/OptionalConverterFactory.java
NightlyNexus: Agree! Will do.

PR #2912, opened 2018-09-28
Test all Rx types can be subscribed to twice

JakeWharton:


swankjesse:
The consequences of this are interesting. Not subscribing at all will make zero network requests!

JakeWharton:
Yep. Same as a Call though. Only things like coroutines and futures trigger when you invoke the method. The distinction should be documented on the call adapters.

oldergod:
> Not subscribing at all will make zero network requests!

Hasn't that been the case before? Did something change recently?

JakeWharton:
Nope. Always been the case.

PR #2915, opened 2018-10-06
Replace @ParametersAreNonnullByDefault with @EverythingIsNonNull

swankjesse:


retrofit/src/main/java/retrofit2/internal/EverythingIsNonNull.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import javax.annotation.Nonnull;
+import javax.annotation.meta.TypeQualifierDefault;
+
+/**
+ * Extends {@code ParametersAreNonnullByDefault} to also apply to Method results and fields.
```
JakeWharton: This is a weird way to describe the annotation because you're not actually extending any behavior. You're defining your own annotation and its own semantics.

swankjesse:
Should we switch to the JetBrains ones?

PR #2931, opened 2018-10-20
Add support for Multi-value map

louistsaitszho:
See #1324 for the original discussion.

I have added a little bit of extra logic that also check if the 2 type is list, and then pass each of the values into converter first.

### DIFF
--- a/retrofit/src/main/java/retrofit2/ParameterHandler.java
+++ b/retrofit/src/main/java/retrofit2/ParameterHandler.java
@@ -56,8 +56,7 @@ abstract class ParameterHandler<T> {
   }
 
   static final class RelativeUrl extends ParameterHandler<Object> {
-    @Override
-    void apply(RequestBuilder builder, @Nullable Object value) {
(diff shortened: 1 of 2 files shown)


JakeWharton:
This only covers one of three annotations which accept map and it does not include any tests. But it's also a behavior change. Someone previously wanting to supply a list and use a string converter to coalesce the values into, say, a comma-separated list would no longer be able to do so.

This hasn't been implemented because we don't have a good way to do it without breaking people. It may end up requiring new annotations like `@QueryMultimap` to express the intent.

louistsaitszho:
Got it. And sorry for the sloopy PR. 

And in terms of just the issue itself, is there gonna be some kind of timeline on "when will you guys start discussing how to actually implement this"?

JakeWharton:
Not currently, no. Creating a new annotation for this is somewhat unsatisfactory because it won't support things like Guava's `Multimap`. So we're stalled on trying to create a more re-usable mechanism.

PR #2932, opened 2018-10-21
Forbid path traversal ('.' and '..') as @Path parameters.

swankjesse:
They're likely to have the unintended effect. For example, passing ".."
to @DELETE /account/book/{isbn}/ yields @DELETE /account/.

### DIFF
--- a/retrofit/src/main/java/retrofit2/RequestBuilder.java
+++ b/retrofit/src/main/java/retrofit2/RequestBuilder.java
@@ -16,6 +16,7 @@
 package retrofit2;
 
 import java.io.IOException;
+import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 import okhttp3.FormBody;
 import okhttp3.Headers;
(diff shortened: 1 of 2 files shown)


retrofit/src/main/java/retrofit2/RequestBuilder.java
```diff
@@ -91,7 +107,12 @@ void addPathParam(String name, String value, boolean encoded) {
       // The relative URL is cleared when the first query parameter is set.
       throw new AssertionError();
     }
-    relativeUrl = relativeUrl.replace("{" + name + "}", canonicalizeForPath(value, encoded));
+    String replacement = canonicalizeForPath(value, encoded);
+    if (PATH_TRAVERSAL.matcher(replacement).matches()) {
```
JakeWharton: Seems like this should apply to the URL _after_ the path replacement. Otherwise `@GET("foo{bar}baz")` would be marked as invalid when you call it with "." or ".." despite being an otherwise valid URL without path traversal.

What about waiting until the full relative URL is built and then escape `/../`? That's the only problematic thing and we should be able to escape it like we do slashes. Doing so would also require forbidding that in the annotation string.

retrofit/src/test/java/retrofit2/RequestFactoryTest.java
```diff
@@ -2783,4 +2850,12 @@ private static void assertBody(RequestBody body, String expected) {
 
     return buildRequest(cls, retrofitBuilder, args);
   }
+
+  static <T> void assertMalformedRequest(Class<T> cls, Object... args) {
```
JakeWharton: nit: drop `<T>` and use `Class<?>`

retrofit/src/main/java/retrofit2/RequestBuilder.java
swankjesse: I changed the behavior to only check after performing replacement. It’s checking a larger string but I think that it’s otherwise a nice improvement.

swankjesse: ALSO – attempted to escape `..` as `%2E%2E` as the very first thing I did. Unfortunately `resolve()` converts `%2E` into `.` when resolving, and it traverses. I don’t think that’s an OkHttp bug; regrettably I think that’s what’s supposed to happen.

<a href="%2E%2E/%2E%2E/okhttp">link to `%2E%2E/%2E%2E/okhttp`</a>

retrofit/src/test/java/retrofit2/RequestFactoryTest.java
swankjesse: Fixed!

PR #2944, opened 2018-10-30
Add maven BOM (Bill Of Materials)

eddumelendez:


JakeWharton:
Thanks! How do we prevent this from becoming out-of-date with the module list in the root?

eddumelendez:
wondering if `retrofit-mock` should be provided too?

> Thanks! How do we prevent this from becoming out-of-date with the module list in the root?

Not sure how to deal with this

eddumelendez:
@JakeWharton PR updated including `retrofit-mock`

JakeWharton:
According to what happened with OkHttp, this is going to include all of our transitive dependencies because of our use of `<dependencyManagement>`.

bom/pom.xml
```diff
@@ -0,0 +1,113 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
```
swankjesse: I think excluding this might prevent transitive deps from being included.
https://github.com/square/okhttp/pull/4413

swankjesse:
I’d like to hold this until we migrate Retrofit to Gradle, at which point we’ll be able to use [Gradle Platforms](https://docs.gradle.org/current/userguide/platforms.html), which are like BOMs but more capable.

PR #2958, opened 2018-11-15
Allow for use of customer marshaller and unmarshaller properties to be used

djr4488:
## Purpose

Some webservices I consume do not accept valid xml with the header line e.g. `<?xml version=\"1.0\" ?>`.  As such, having the ability to set properties for the marshaller was needed.  For completeness, though not all implementations have unmarshaller properties, I also included the ability to provide those properties as well.  

## Use

```
    Map<String, Object> marshallerProperties = new HashMap<>();
    marshallerProperties.put(Marshaller.JAXB_FRAGMENT, true);
```

What is needed is a map which has the property name for the key, and the setting for the value.  If the map is null or empty, JaxB defaults are used.

### DIFF
--- a/retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbConverterFactory.java
+++ b/retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbConverterFactory.java
@@ -22,8 +22,6 @@ import java.util.Map;
 import javax.annotation.Nullable;
 import javax.xml.bind.JAXBContext;
 import javax.xml.bind.JAXBException;
-import javax.xml.bind.Marshaller;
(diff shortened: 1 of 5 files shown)


swankjesse:
Can you implement `JAXBContext` to do this?

```
class MyJaxbContext(val delegate: JAXBContext) : JAXBContext() {
  fun createMarshaller(): Marshaller {
    val result = delegate.createMarshaller()
    result.setProperty(Marshaller.JAXB_FRAGMENT, true);
    return result
  }
   
  fun createUnmarshaller(): Unmarshaller = delegate.createUnmarshaller()
  fun createValidator(): Validator = delegate.createValidator()
}
```

That way you can do whatever you want to do to the marshaller, not just set properties.

Issue #2974, opened 2018-11-21
Download failed, url @Path encoded while server didn't support decode via http GET!


1step2hell:
# Bug
I developered a SNS share library, if a imageUrl will be shared,
download it with retrofit firstly and save it to local storage, then share it.
Most of the time, it works well.
But for some particular imageUrls, there is something wrong:
* Download failed, http 400 bad request.
* Download successed, but cannot be decode to bitmap.

As a contrast, I write another download method with OkHttp directly, it works well all the time.

Here is my retrofit download code:
```
public static void downloadByRetrofit(String imageUrl, final String dstPath) {
        String baseUrl = "";
        String path = "";
        try {
            URL url = new URL(imageUrl);
            baseUrl = url.getProtocol() + "://" + url.getAuthority() + "/";
            path = url.getPath();
        } catch (MalformedURLException e) {
            return;
        }
        OkHttpClient client = new OkHttpClient.Builder()
                .readTimeout(20000, TimeUnit.MILLISECONDS)
                .writeTimeout(20000, TimeUnit.MILLISECONDS)
                .addNetworkInterceptor(new HttpLoggingInterceptor().setLevel(HttpLoggingInterceptor.Level.BODY))
                .build();
        new Retrofit.Builder()
                .baseUrl(baseUrl)
                .client(client)
                .build()
                .create(ServiceApi.class)
                .requestByRetrofit(path)
                .enqueue(new Callback<ResponseBody>() {
                    @Override
                    public void onResponse(Call<ResponseBody> call, Response<ResponseBody> response) {
                        saveRetrofitFile(response, dstPath);
                    }

                    @Override
                    public void onFailure(Call<ResponseBody> call, Throwable t) {
                        Log.e(TAG, "downloadByRetrofit onFailure:" + t.getMessage());
                    }
                });
    }

interface ServiceApi {
    @GET("{path}")
    Call<ResponseBody> requestByRetrofit(@Path("path") String path);
}

```

And the OkHttp log:

1. Download test imageUrl="https://www.baidu.com/img/bd_logo1.png";
```
D/OkHttp: --> GET https://www.baidu.com/%2Fimg%2Fbd_logo1.png http/1.1
D/OkHttp: Host: www.baidu.com
D/OkHttp: Connection: Keep-Alive
D/OkHttp: Accept-Encoding: gzip
D/OkHttp: User-Agent: okhttp/3.12.0
D/OkHttp: --> END GET
[...]

PR #2975, opened 2018-11-21
Fix issue #2974

1step2hell:


swankjesse:
I do not think this is an appropriate fix. Could you create a separate pull request with a failing test case only? As is I’m missing exactly what the problem is.

1step2hell:
Sorry for sending a pull request directly and my poor English.
I update the issue, attach more detail log, please read [Issue#2974](https://github.com/square/retrofit/issues/2974) and spend several minutes to run [Android Sample](https://github.com/1step2hell/RetrofitUrlEncodeBug).
Although it might not be a retrofit bug, it's a problem definitely.

PR #2981, opened 2018-12-03
Account for static interface methods

### LINKED ISSUE #2032: Support for Java8 interface static methods
Retrofit 2.1.0 currently throws an exception building a client implementation proxy if the interface has static methods.

ZacSweers:
Resolves #2032

Not able to add a test though since it's java 7

### DIFF
--- a/retrofit/src/main/java/retrofit2/Retrofit.java
+++ b/retrofit/src/main/java/retrofit2/Retrofit.java
@@ -18,6 +18,7 @@ package retrofit2;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 import java.lang.reflect.Proxy;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
@@ -143,6 +144,8 @@ public final class Retrofit {
             }
             if (platform.isDefaultMethod(method)) {
               return platform.invokeDefaultMethod(method, service, proxy, args);
+            } else if (Modifier.isStatic(method.getModifiers())) {
+              return method.invoke(null, args);
             }
             return loadServiceMethod(method).invoke(args != null ? args : emptyArgs);
           }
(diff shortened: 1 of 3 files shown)


retrofit/src/main/java/retrofit2/Retrofit.java
```diff
@@ -143,6 +144,8 @@
             }
             if (platform.isDefaultMethod(method)) {
               return platform.invokeDefaultMethod(method, service, proxy, args);
+            } else if (Modifier.isStatic(method.getModifiers())) {
```
hzsweers: Let me know if you want this removed. It's weird to be calling a static method on an instance, but technically legal. This at least keeps that behavior

swankjesse: Is this line reachable in test cases? I assume it won’t be dispatched this way

JakeWharton:
Please duplicate `Java8DefaultMethodsTest` and update for this functionality. You can temporarily set the project to compile with 8 to run it and then comment it out once finished.

retrofit/src/main/java/retrofit2/Retrofit.java
JakeWharton: Yeah I don't think this can happen. The only case I can think of is if a virtual method was made static without recompiling the caller. But even then I would expect the VM to intervene.

ZacSweers: done f1fbbda

ZacSweers:
Updated the test in f6da87c. Also fixed an issue with it redeclaring a `response` variable along the way

retrofit/src/test/java/retrofit2/Java8DefaultMethodsTest.java
```diff
@@ -34,12 +38,13 @@
   //  Retrofit retrofit = new Retrofit.Builder()
   //      .baseUrl(server.url("/"))
   //      .addConverterFactory(new ToStringConverterFactory())
+  //      .validateEagerly(true)
```
JakeWharton: Can you move this to a separate test class? I don't want to conflate the two things being tested here together.

retrofit/src/main/java/retrofit2/Retrofit.java
swankjesse: Is the condition reachable?

swankjesse: Ahh, this happens during reflection, not invocation. Works for me.

Issue #2867, opened 2018-08-23
When response code is 204 (or 205), convert is not used, so RxJava2CallAdapterFactory throws the NoSuchElementException


NeverwinterMoon:
I stumbled upon this: https://github.com/square/retrofit/issues/1554 and added a converter because I wanted to handle the empty response body and not pass it as null to Observable. That thread is talking about code 200 and empty body though.

Then I found out that in the current Retrofit version no converter is called at all for 204 and 205:

```
    if (code == 204 || code == 205) {
      rawBody.close();
      return Response.success(null, rawResponse);
    }
```

This forces me to use `Single<Response<SomeModel>>` instead of `Single<SomeModel>` I was hoping to use...

I understand that 204 stands for no content, so it's only logical to pass the response as null. But for my specific case, I would prefer to handle 204 in `onError`, so I hoped to handle the empty body in the converter and just throw a custom error there instead.

Is there any chance that approach will be changed? I am relatively new to Retrofit but I am used to Alamofire on iOS and there it's possible to specify during configuration what is treated as a successful response code:

` .validate(statusCode: 200)`

So every code that is not on that list would be handled as an error.

artem-zinnatullin:
You should be able to add an [OkHttp Interceptor](https://github.com/square/okhttp/wiki/Interceptors) that can convert original response code from `204` to `200` before it gets passed to Retrofit

NeverwinterMoon:
@artem-zinnatullin Aha, thanks for your advice! I just tried the throwing-in-interceptor approach out, and it seems to satisfy my needs.

Nonetheless, I am curious if Retrofit would ever allow to configure custom validation (similar to Alamofire) for response codes or if it's completely against the Retrofit's design ideas.

villesinisalo:
```
    if (code == 204 || code == 205) {
      return Response.success(null, rawResponse);
    }
```

Isn't this violating the `Response` class contract? Reading from `Response.java`:

```
  /** Returns true if {@link #code()} is in the range [200..300). */
  public boolean isSuccessful() {
    return rawResponse.isSuccessful();
  }

  /** The deserialized response body of a {@linkplain #isSuccessful() successful} response. */
  public @Nullable T body() {
    return body;
  }
```

The way I read it, `response.body()` is always non-null if `isSuccesful()` returns `true`. But in fact, if the server returns 204, we get a `null` regardless. 

[...]

PR #3003, opened 2019-01-14
Handle 204/205 responses in RxJava 2 types

ZacSweers:
This is a proposal implementation to address #2867.

The issue is two-fold

1 - Retrofit 2's RxJava 2 `BodyObservable` will, on 204/205 status codes, still try to send the `null` body into `Observer#onNext`, even though RxJava 2 does not allow nulls in the stream
2 - RxJava 2's `Single#fromObservable()` implementation does not null-check the onNext value, and instead just silently skips it. When an `onComplete` event comes from `BodyObserver` later, it then throws a `NoSuchElementException`

This tries to solve it by being aware of 204/205 status codes in `BodyObservable` and sending a new `NoContentException` through `onError` in these events instead of the obscured `null` into RxJava. `NoContentException` is basically identical to `HttpException` except that it's only for 204/205 codes and used to indicate that a 204/205 event happened in a non-`Completable` environment.

Other considerations made:
- Converter - not feasible because converters are not given a chance at 204/205 responses
- Forwarding call adapter - can work, but requires a lot of knowledge of how the underlying RxJava 2 adapter behaves and would (IMO) be something that's always required. Even if consumers before this had custom adapters to handle this, they can still lean on Retrofit and now catch a more concrete exception if they were trying to guess from an `onErrorResumeNext` or something similar.

Open questions: Unclear how to handle other `null`-body scenarios, such as a misbehaving converter.

### DIFF
--- a/retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/NoContentException.java
+++ b/retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/NoContentException.java
@@ -16,6 +16,7 @@
 package retrofit2.adapter.rxjava2;
 
 import io.reactivex.Completable;
+import javax.annotation.Nullable;
 import retrofit2.Response;
 
 /** Exception for an unexpected, 204/205 HTTP response when not used with {@link Completable}. */
@@ -51,6 +52,7 @@ public final class NoContentException extends RuntimeException {
   /**
    * The full HTTP response. This may be null if the exception was serialized.
    */
+  @Nullable
   public Response<?> response() {
     return response;
   }
--- a/retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/NoContentException.java
+++ b/retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/NoContentException.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright (C) 2016 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package retrofit2.adapter.rxjava2;
+
+import io.reactivex.Completable;
+import retrofit2.Response;
+
+/** Exception for an unexpected, 204/205 HTTP response when not used with {@link Completable}. */
+public final class NoContentException extends RuntimeException {
+  private static String getMessage(Response<?> response) {
+    if (response == null) {
+      throw new NullPointerException("response == null");
+    }
+    return "HTTP " + response.code() + " " + response.message();
+  }
+
+  private final int code;
+  private final String message;
+  private final transient Response<?> response;
+
+  public NoContentException(Response<?> response) {
+    super(getMessage(response));
+    this.code = response.code();
+    this.message = response.message();
+    this.response = response;
+  }
+
+  /** HTTP status code. */
+  public int code() {
+    return code;
+  }
+
+  /** HTTP status message. */
+  public String message() {
+    return message;
+  }
+
+  /**
+   * The full HTTP response. This may be null if the exception was serialized.
+   */
+  public Response<?> response() {
+    return response;
+  }
+}
--- a/retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/RxJava2CallAdapter.java
+++ b/retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/RxJava2CallAdapter.java
@@ -63,7 +63,7 @@ final class RxJava2CallAdapter<R> implements CallAdapter<R, Object> {
     if (isResult) {
       observable = new ResultObservable<>(responseObservable);
     } else if (isBody) {
-      observable = new BodyObservable<>(responseObservable, !isCompletable);
+      observable = new BodyObservable<>(responseObservable, isSingle);
     } else {
       observable = responseObservable;
     }
--- a/retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/RxJava2CallAdapter.java
+++ b/retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/RxJava2CallAdapter.java
@@ -63,7 +63,7 @@ final class RxJava2CallAdapter<R> implements CallAdapter<R, Object> {
     if (isResult) {
       observable = new ResultObservable<>(responseObservable);
     } else if (isBody) {
-      observable = new BodyObservable<>(responseObservable);
+      observable = new BodyObservable<>(responseObservable, !isCompletable);
     } else {
       observable = responseObservable;
     }
--- a/retrofit-adapters/rxjava2/src/test/java/retrofit2/adapter/rxjava2/CompletableTest.java
+++ b/retrofit-adapters/rxjava2/src/test/java/retrofit2/adapter/rxjava2/CompletableTest.java
@@ -54,6 +54,22 @@ public final class CompletableTest {
     observer.assertComplete();
   }
 
+  @Test public void completableSuccess204() {
+    server.enqueue(new MockResponse().setResponseCode(204));
+
+    RecordingCompletableObserver observer = observerRule.create();
+    service.completable().subscribe(observer);
+    observer.assertComplete();
+  }
+
+  @Test public void completableSuccess205() {
+    server.enqueue(new MockResponse().setResponseCode(205));
+
+    RecordingCompletableObserver observer = observerRule.create();
+    service.completable().subscribe(observer);
+    observer.assertComplete();
+  }
+
   @Test public void completableSuccess404() {
     server.enqueue(new MockResponse().setResponseCode(404));
 
--- a/retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/BodyObservable.java
+++ b/retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/BodyObservable.java
@@ -25,21 +25,25 @@ import retrofit2.Response;
 
 final class BodyObservable<T> extends Observable<T> {
   private final Observable<Response<T>> upstream;
+  private final boolean errorOnNoContent;
 
-  BodyObservable(Observable<Response<T>> upstream) {
+  BodyObservable(Observable<Response<T>> upstream, boolean errorOnNoContent) {
     this.upstream = upstream;
+    this.errorOnNoContent = errorOnNoContent;
   }
 
   @Override protected void subscribeActual(Observer<? super T> observer) {
-    upstream.subscribe(new BodyObserver<T>(observer));
+    upstream.subscribe(new BodyObserver<T>(observer, errorOnNoContent));
   }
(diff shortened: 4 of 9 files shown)


allidoiswarren:
I couldn't comment on this from my main account, @naturalwarren so posting from an alt:

Should this be checking 201s as well?

ZacSweers:
As far as I understand, converters are given a chance at 201 responses, and thus you should be able to model that response directly

retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/RxJava2CallAdapter.java
```diff
@@ -63,7 +63,7 @@
     if (isResult) {
       observable = new ResultObservable<>(responseObservable);
     } else if (isBody) {
-      observable = new BodyObservable<>(responseObservable);
+      observable = new BodyObservable<>(responseObservable, !isCompletable);
```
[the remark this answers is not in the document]
ZacSweers: Possibly! So should Observable and Flowale then. 

When my colleague and I were looking at this, our mental model was basically that if you have an endpoint that tries to coerce everything through onSuccess as a sealed type of some kind, it would be weird if a couple types of status codes basically didn't route that way and would be confusing for consumers that otherwise would never check onComplete. With Completable at least it explicitly doesn't allow for that ambiguity. We could err on the side of flexibility with this though and leave it up to the developer.

swankjesse:
Could you provide a concrete example? Does the endpoint ever return a response? What’s the drawback of using `Response<Void>` ?

ZacSweers:
The endpoint in this case is basically an endpoint for holding dynamic strings for our client. When there are strings available, then it returns a successful response with the strings. If there are none, then it returns a 204 status code to signal that. It is weirdly a POST request despite basically being a GET, but I suspect that endpoint writers didn't want to use query parameters for the contents of the request body.

Since a successful response returns real data, we can't just use `Response<Void>`

Endpoint roughly looks something like this

```kotlin
data class GetStringsResponse(val strings: Map<String, String>)

interface StringsApi {
  fun getStrings(request: GetStringsRequest): Single<GetStringsResponse>
}
```

It can also return 400, 401, 429, and 500 responses. Under the hood we have basically a wrapper that does an `onErrorResumeNext()` to map all of these exception types + the successful response type to be something like a sealed type

```kotlin
sealed class GetStringsResult {
  data class GetStringsResponse(val strings: Map<String, String>) : GetStringsResult()
  object NoContent : GetStringsResult()
  data class BadRequest(...) : GetStringsResult()
  data class Unauthorized(...) : GetStringsResult()
  data class RateLimited(...) : GetStringsResult()
  data class ServerError(...) : GetStringsResult()
}

class StringsClient {
  fun getStrings(request: GetStringsRequest): Single<GetStringsResult> {
    return api.getStrings(request)
        .onErrorResumeNext {
          // If httpexception, coerce to one of the 4xx/500 types from raw response body
          // Would like to do the same with NoContent. Could catch NoSuchElementException, but that doesn't give access to the raw response body. Also not very explicit
        }
  }
}
```

swankjesse:
That's a curious API. I'm asking because I want to know whether this proposed behaviour would be useful or surprising in other contexts where the server returns 204.

ZacSweers:
I think it'd be less surprising than the current NoSuchElementException :). Also applies to 205 status codes

My preference would probably be to just allow converters to model empty bodies, but I'm not sure if that'd constitute a breaking change

retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/NoContentException.java
```diff
@@ -0,0 +1,57 @@
...
+
+  /** HTTP status message. */
+  public String message() {
+    return message;
+  }
+
+  /**
+   * The full HTTP response. This may be null if the exception was serialized.
+   */
+  public Response<?> response() {
```
JakeWharton: Needs `@Nullable`

retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/NoContentException.java
```diff
@@ -0,0 +1,57 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package retrofit2.adapter.rxjava2;
+
+import io.reactivex.Completable;
+import retrofit2.Response;
+
+/** Exception for an unexpected, 204/205 HTTP response when not used with {@link Completable}. */
+public final class NoContentException extends RuntimeException {
```
JakeWharton: Does this type need to exist? Seems like a normal `HttpException` could be used.

retrofit-adapters/rxjava2/src/test/java/retrofit2/adapter/rxjava2/CompletableTest.java
```diff
@@ -54,6 +54,22 @@
     observer.assertComplete();
   }
 
+  @Test public void completableSuccess204() {
+    server.enqueue(new MockResponse().setResponseCode(204));
+
+    RecordingCompletableObserver observer = observerRule.create();
+    service.completable().subscribe(observer);
+    observer.assertComplete();
```
JakeWharton: Is this a behavior change? What happened before the change?

retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/NoContentException.java
ZacSweers: We could! The doc for `HttpException` currently says it's only for `non-2xx HTTP response`, so I'm not sure if this would constitute an undesirable behavior change

retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/NoContentException.java
ZacSweers: Should `HttpException` (where this is copied from) be updated too?

retrofit-adapters/rxjava2/src/test/java/retrofit2/adapter/rxjava2/CompletableTest.java
ZacSweers: I don't believe so. This is the only case that sort of "safely" worked before because any onNext event in the underlying observable was knowingly ignored even if it was null (semantics of `ignoreElements()`). This worked before the change too, this is just to have a test for that behavior being explicit while other tests show it throws in non-Completable types now for 204/205 responses.

Other behaviors:
* Single and Maybe impls both silently just ignore nulls in onNext (Single throws NSEE, Maybe completes).
* Flowable and Observable both defer to the given observer, which may or may no nullcheck inputs (though presumably they'd be unexpected since RxJava 2 prohibits null items in the stream). They could complete if the onNext consumer handles/ignores nulls, or a stricter observer intermediary could fail early

retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/NoContentException.java
JakeWharton: `HttpException` was updated and that test ensures you can't supply `null`, not that it can't return `null`. Note the Javadoc which details when you will receive a `null` value.

retrofit-adapters/rxjava2/src/test/java/retrofit2/adapter/rxjava2/CompletableTest.java
JakeWharton: Right. I vaguely remember this being the observed behavior. Just wanted to make sure we don't deviate from it.

retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/NoContentException.java
JakeWharton: Yeah I suspect it might. We wouldn't even want to subclass either because an `instanceof` check still might be happening. Let's leave it separate.

retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/NoContentException.java
ZacSweers: Ahh I missed the `transient` modifier. Got it d90cbe4d8f0ec5e88ce56d1de28ca7b4483da0d2

retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/RxJava2CallAdapter.java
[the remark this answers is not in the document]
ZacSweers: Thinking on this more, I think this could work if we wanted to make it entirely silent and route through onComplete if possible _only_ if we're fine with not exposing the exception. That is to say - `Single` users would get the `NoContentException`, but everyone else would basically have to use some sort of continuation (`switchIfEmpty()`, `defaultIfEmpty()`, etc) if they need to handle it otherwise. They wouldn't have access to the underlying response, but maybe we make the call that that's ok? It's kind of sneaky/silent behavior, but I also suspect that'd be the common case and not making people always manually handle the `NoContentException` in the common case seems possibly better. Also makes it more consistent, rather than making `Completable` the exception even though others have `onComplete`.

Thoughts?

ZacSweers:
Another thought on this - should Retrofit's native understanding of `Unit` and `Void` be wired in here? I.e. not throw this if that's the response type

egor-n:
> Another thought on this - should Retrofit's native understanding of `Unit` and `Void` be wired in here? I.e. not throw this if that's the response type

This sounds reasonable to me. A `Response<Unit>` should have `Unit` as the return value of `body()` if the response code is 204. This is not the case currently:

```java
interface Service {
  @GET("/")
  Call<Unit> empty();
}

@Test public void noNull204() throws IOException {
      .baseUrl(server.url("/"))
      .build();
  Service example = retrofit.create(Service.class);

  server.enqueue(new MockResponse().setResponseCode(204));

  Response<Unit> response = example.empty().execute();
  assertThat(response.isSuccessful()).isTrue();
  assertThat(response.body()).isSameAs(Unit.INSTANCE); // AssertionError: Expecting: <kotlin.Unit> and actual: <null>
}

retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/RxJava2CallAdapter.java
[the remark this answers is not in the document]
ZacSweers: I demoed this in 86f387d, but can revert if we don't want that behavior

ZacSweers:
So Unit would be nice, but I don't see how we could add it in a way that doesn't duplicate a bunch of code that's already there for the mainline retrofit. If 204/205 types could be coerced to Unit at the converter layer, then no changes are needed for call adapters like this one to work. I think at the very least Unit support's probably worth punting to a separate issue/PR

JakeWharton:
I'm not sure this is a satisfactory solution since it solely lives in the RxJava 2 adapter. #3081 just highlighted that this is causing problems for users of `CompletableFuture` as well.

oleg-voronovich:
The same problem with `ReactorCallAdapter`.

RaviH:
> I'm not sure this is a satisfactory solution since it solely lives in the RxJava 2 adapter. #3081 just highlighted that this is causing problems for users of `CompletableFuture` as well.

I have a valid scenario where an API does a find based on a criteria. If something is found in the DB, then it's a 200 response with data, else it's 204. And this is causing an issue in @JakeWharton's converter library for reactor / Mono. Any recommendations on how to avoid it?

DFreds:
Is this planning on being merged soon? Just ran into this issue myself

DFreds:
Guess not. lol

ZacSweers:
Sorry, I'm not personally affected by this anymore since changing jobs and as mentioned above this will likely require a different solution to work across different call adapters.

Issue #2773, opened 2018-05-31
Feature Request: "IllegalArgumentException: Path parameter XXX value must not be null." should include url


tir38:
I was not properly setting a `Path` parameter and saw:

`Caused by: java.lang.IllegalArgumentException: Path parameter "user_id" value must not be null.`

The problem is that I have ~20 web requests all with the `user_id` path parameter. I have no easy way to diagnose which one I'm calling incorrectly. It would be great if that error message included the url.

I think it should be pretty easy to update `Path` to:

```
static final class Path<T> extends ParameterHandler<T> {
    @Override
    void apply(RequestBuilder builder, @Nullable T value) throws IOException {
        if (value == null) {
            throw new IllegalArgumentException(
                    "Path parameter \"" + name + "\" value must not be null in request "
                            + builder.baseUrl + builder.relativeUrl);
        }
        builder.addPathParam(name, valueConverter.convert(value), encoded);
    }
}
```

Probably worth updating all `ParameterHandler`s that throw when value is null.

For the record, calling a service method inside RxJava stream results in a stack trace that isn't very helpful:

```
   io.reactivex.exceptions.OnErrorNotImplementedException: Path parameter "user_id" value must not be null.
        at io.reactivex.internal.functions.Functions$OnErrorMissingConsumer.accept(Functions.java:704)
        at io.reactivex.internal.functions.Functions$OnErrorMissingConsumer.accept(Functions.java:701)
        at io.reactivex.internal.observers.ConsumerSingleObserver.onError(ConsumerSingleObserver.java:47)
        at io.reactivex.internal.operators.single.SingleObserveOn$ObserveOnSingleObserver.run(SingleObserveOn.java:79)
        at io.reactivex.android.schedulers.HandlerScheduler$ScheduledRunnable.run(HandlerScheduler.java:109)
        at android.os.Handler.handleCallback(Handler.java:751)
        at android.os.Handler.dispatchMessage(Handler.java:95)
        at android.os.Looper.loop(Looper.java:154)
        at android.app.ActivityThread.main(ActivityThread.java:6121)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:889)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:779)
     Caused by: java.lang.IllegalArgumentException: Path parameter "user_id" value must not be null.
[...]

PR #3007, opened 2019-01-21
Replace directly thrown exceptions with Utils.parameterError

vitorhugods:
Proposal of feature implementation regarding #2773.

It does not include the URL per say, but says exactly which defined method, from which class was the one which passed parameter was invalid. Which might be even more informative for issue mitigation.

### DIFF
--- a/retrofit/src/main/java/retrofit2/ParameterHandler.java
+++ b/retrofit/src/main/java/retrofit2/ParameterHandler.java
@@ -65,9 +65,9 @@ abstract class ParameterHandler<T> {
     }
 
     @Override void apply(RequestBuilder builder, @Nullable Object value) {
-      try{
+      try {
         checkNotNull(value, "@Url parameter is null.");
-      }catch (NullPointerException npe){
+      } catch (NullPointerException npe) {
         throw Utils.parameterError(method, npe, p, "@Url parameter is null.");
       }
       builder.setRelativeUrl(value);
@@ -111,7 +111,8 @@ abstract class ParameterHandler<T> {
 
     @Override void apply(RequestBuilder builder, @Nullable T value) throws IOException {
       if (value == null) {
-        throw Utils.parameterError(method, p, "Path parameter \"" + name + "\" value must not be null.");
+        throw Utils.parameterError(method, p,
+                "Path parameter \"" + name + "\" value must not be null.");
       }
       builder.addPathParam(name, valueConverter.convert(value), encoded);
     }
@@ -197,7 +198,7 @@ abstract class ParameterHandler<T> {
       }
     }
 
-    private void throwParameterError(String message){
+    private void throwParameterError(String message) {
       throw Utils.parameterError(method, p, message);
     }
   }
@@ -233,7 +234,7 @@ abstract class ParameterHandler<T> {
       }
     }
 
-    private void throwParameterError(String message){
+    private void throwParameterError(String message) {
       throw Utils.parameterError(method, p, message);
     }
   }
@@ -304,7 +305,7 @@ abstract class ParameterHandler<T> {
       }
     }
 
-    private void throwParameterError(String message){
+    private void throwParameterError(String message) {
       throw Utils.parameterError(method, p, message);
     }
   }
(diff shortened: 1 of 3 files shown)


vitorhugods:
There is one change that could affect people using the current version:
RelativeUrl.apply method was the only one that could throw NullPointerException, when all the other ParameterHandlers would throw IllegalArgumentException when dealing with null values.

I think it makes the behaviour more consistent, but anyone catching this NullPointerException should change it.

retrofit/src/main/java/retrofit2/ParameterHandler.java
```diff
@@ -55,8 +56,20 @@
...
+    RelativeUrl(Method method, int p) {
+      this.method = method;
+      this.p = p;
+    }
+
     @Override void apply(RequestBuilder builder, @Nullable Object value) {
-      checkNotNull(value, "@Url parameter is null.");
+      try {
+        checkNotNull(value, "@Url parameter is null.");
+      } catch (NullPointerException npe) {
```
JakeWharton: There's no need to do this. We can just check for null locally and pass null as the cause to the `Utils` method.

retrofit/src/main/java/retrofit2/ParameterHandler.java
```diff
@@ -176,34 +197,46 @@
         builder.addQueryParam(entryKey, convertedEntryValue, encoded);
       }
     }
+
+    private void throwParameterError(String message) {
```
JakeWharton: These helper methods should either be inlined or you should move it to the abstract class and pass the `Method` and parameter index up

retrofit/src/main/java/retrofit2/ParameterHandler.java
```diff
@@ -267,15 +304,23 @@
...
     private final Converter<T, RequestBody> converter;
+    private final Method method;
+    private final int p;
 
-    Part(Headers headers, Converter<T, RequestBody> converter) {
+    Part(Method method, int p, Headers headers, Converter<T, RequestBody> converter) {
       this.headers = headers;
       this.converter = converter;
+      this.method = method;
+      this.p = p;
```
JakeWharton: Can we make the field order and statement order match the constructor argument order?

retrofit/src/main/java/retrofit2/ParameterHandler.java
```diff
@@ -337,26 +387,35 @@ private RawPart() {
...
       RequestBody body;
       try {
         body = converter.convert(value);
       } catch (IOException e) {
-        throw new RuntimeException("Unable to convert " + value + " to RequestBody", e);
+        throw Utils.parameterError(method, e, p, "Unable to convert " + value + " to RequestBody");
       }
       builder.setBody(body);
     }
+
```
JakeWharton: nit: remove

retrofit/src/main/java/retrofit2/ParameterHandler.java
vitorhugods: Yeaaah, I was going to change this. Realized this when commenting before and reviewing the code. Just didn't access to my computer then. I'll work on this and the other changes right now.

retrofit/src/main/java/retrofit2/ParameterHandler.java
```diff
@@ -84,17 +97,22 @@
     private final String name;
     private final Converter<T, String> valueConverter;
     private final boolean encoded;
+    private final Method method;
+    private final int p;
+
 
-    Path(String name, Converter<T, String> valueConverter, boolean encoded) {
+    Path(Method method, int p, String name, Converter<T, String> valueConverter, boolean encoded) {
       this.name = checkNotNull(name, "name == null");
```
vitorhugods: Would it be better if these `checkNotNull` were replaced with the `Utils.parameterError` as well?

JakeWharton: If they're impossible we actually usually prefer `AssertionError`. I would leave them as-is because they should never be observable by users.

PR #3011, opened 2019-01-22
Mark HttpException.response() as nullable

### LINKED ISSUE #2983: HttpException.response() should be marked w/ @Nullable
What kind of issue is this?

 - [x] Bug report

Code for that method in Retrofit2 2.5 is:

```
/**
   * The full HTTP response. This may be null if the exception was serialized.
   */
  public Response<?> response() {
    return response;
  }
```

JakeWharton:
If the exception was serialized this value is not retained.

Closes #2983

### DIFF
--- a/retrofit/src/main/java/retrofit2/HttpException.java
+++ b/retrofit/src/main/java/retrofit2/HttpException.java
@@ -15,6 +15,8 @@
  */
 package retrofit2;
 
+import javax.annotation.Nullable;
+
 import static retrofit2.Utils.checkNotNull;
 
 /** Exception for an unexpected, non-2xx HTTP response. */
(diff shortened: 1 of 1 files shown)


swankjesse:
That is quite unfortunate.

DavidMihola:
Since I am just in the process of adapting https://gist.github.com/naturalwarren/56b54759b0f690622938caa91f037be6 for our own purposes - do we need to worry about this in the context of a `CallAdapter`? Is there any scenario in which the `HttpException` would have been serialized at this point?

JakeWharton:
Nope. Not unless you had another call adapter which you delegated to which,
for some reason, roundtripped the object through serialization. In practice
this will never happen.

DavidMihola:
@JakeWharton: Perfect - thanks so much for the quick reply!

Issue #3018, opened 2019-01-25
Dependency conflicts on com.squareup.okio:okio.jar, causing unexpected hehaviors


HelloCoCooo:
Hi, there are multiple versions of **com.squareup.okio:okio:jar** in **retrofit-parent-2.5.0**. As shown in the following dependency tree, according to Maven's “nearest wins” strategy, only **com.squareup.okio:okio:jar:1.13.0** can be loaded, and **com.squareup.okio:okio:jar:1.15.0** will be shadowed.

As **com.squareup.okio:okio:jar:1.15.0** is missing during the build process, method **retrofit2.converter.moshi.MoshiConverterFactory: retrofit2.Converter responseBodyConverter(java.lang.reflect.Type,java.lang.'annotation'.Annotation[],retrofit2.Retrofit)** actually references method **okio.RealBufferedSource: int select(okio.Options)** in the unexpected version **com.squareup.okio:okio:jar:1.13.0**  via the following invocation path:

`<retrofit2.converter.moshi.MoshiConverterFactory: retrofit2.Converter responseBodyConverter(java.lang.reflect.Type,java.lang.'annotation'.Annotation[],retrofit2.Retrofit)> C:\Users\Flipped\Desktop\project\moshi\.\target\classes
<com.squareup.moshi.Moshi: com.squareup.moshi.JsonAdapter adapter(java.lang.reflect.Type,java.util.Set)> C:\Users\Flipped\.m2\repository\com\squareup\moshi\moshi\1.5.0\moshi-1.5.0.jar
<com.squareup.moshi.Moshi$Builder$2: com.squareup.moshi.JsonAdapter create(java.lang.reflect.Type,java.util.Set,com.squareup.moshi.Moshi)> C:\Users\Flipped\.m2\repository\com\squareup\moshi\moshi\1.5.0\moshi-1.5.0.jar
<com.squareup.moshi.Util: boolean isAnnotationPresent(java.util.Set,java.lang.Class)> C:\Users\Flipped\.m2\repository\com\squareup\moshi\moshi\1.5.0\moshi-1.5.0.jar
<okhttp3.Cache$2: boolean hasNext()> C:\Users\Flipped\.m2\repository\com\squareup\okhttp3\okhttp\3.12.0\okhttp-3.12.0.jar
<okio.RealBufferedSource: java.lang.String readUtf8LineStrict()> C:\Users\Flipped\.m2\repository\com\squareup\okio\okio\1.13.0\okio-1.13.0.jar
[...]

PR #3020, opened 2019-01-27
#3018 Fix dependency conflict issue for com.squareup.okio:okio

HelloCoCooo:
#3018 Fix dependency conflict issue for com.squareup.okio:okio. (conflict version:1.13.0:-1.15.0)
Dependency tree after revision:

[INFO] com.squareup.retrofit2:converter-moshi:jar:2.5.0
[INFO] +- com.squareup.retrofit2:retrofit:jar:2.5.0:compile
[INFO] |  \- com.squareup.okhttp3:okhttp:jar:3.12.0:compile
[INFO] |     **\- com.squareup.okio:okio:jar:1.15.0:compile**
[INFO] +- com.squareup.moshi:moshi:jar:1.5.0:compile
[INFO] +- com.google.code.findbugs:jsr305:jar:3.0.2:provided
[INFO] +- junit:junit:jar:4.12:test
[INFO] |  \- org.hamcrest:hamcrest-core:jar:1.3:test
[INFO] +- com.squareup.okhttp3:mockwebserver:jar:3.12.0:test
[INFO] |  +- (com.squareup.okhttp3:okhttp:jar:3.12.0:test - omitted for duplicate)
[INFO] |  \- (junit:junit:jar:4.12:test - omitted for duplicate)
[INFO] \- org.assertj:assertj-core:jar:1.7.0:test

### DIFF
--- a/retrofit-converters/moshi/pom.xml
+++ b/retrofit-converters/moshi/pom.xml
@@ -31,7 +31,7 @@
 <dependency>
     	<groupId>com.squareup.okio</groupId>
       		<artifactId>okio</artifactId>
-      		<version>2.2.1</version>
+      		<version>1.16.1</version>
     </dependency>
     <dependency>
       <groupId>junit</groupId>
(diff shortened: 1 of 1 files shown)


JakeWharton:
I don't think this is the correct solution. If Moshi winds up needing a newer version of Okio then it will fail at runtime.

JakeWharton:
Actually that seems to be what's causing the CI failure...

PR #3028, opened 2019-02-15
Add parameter annotation @Tag for setting request tags

### LINKED ISSUE #2812: Add support for OkHttp's request tags
I want to cancel all request for Activity or Fragment  during the end of their life style. In OkHttp, I can use Request.Builder.setTag method to setTag for every Request.
 
And then, in my client, I add this method:

    public static void cancelRequest(Object tag) {
        if (tag == null) {
            return;
        }
        List<Call> queuedCalls = mClient.dispatcher().queuedCalls();
        for (Call call : queuedCalls) {
            if (tag.equals(call.request().tag())) {
                call.cancel();
            }
        }
        List<Call> runningCalls = mClient.dispatcher().runningCalls();
        for (Call call : runningCalls) {
            if (tag.equals(call.request().tag())) {
                call.cancel();
            }
        }
    }
And in BaseActivity and BaseFragment's onDestroy method, I can do this:

    APIRestClient.cancelRequset(this);    

I think it's useful and convenient.
But in Retrofit, I can't use field tag. So how can I do above in Retrofit efficiently and can you give me some suggestions？ Thanks！😊

JakeWharton:
Closes #2812.

### DIFF
--- a/retrofit/src/main/java/retrofit2/RequestFactory.java
+++ b/retrofit/src/main/java/retrofit2/RequestFactory.java
@@ -55,6 +55,7 @@ import retrofit2.http.Path;
 import retrofit2.http.Query;
 import retrofit2.http.QueryMap;
 import retrofit2.http.QueryName;
+import retrofit2.http.Tag;
 import retrofit2.http.Url;
 
 import static retrofit2.Utils.methodError;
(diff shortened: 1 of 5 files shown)


retrofit/src/main/java/retrofit2/RequestFactory.java
```diff
@@ -712,6 +713,24 @@ private Headers parseHeaders(String[] headers) {
         }
         gotBody = true;
         return new ParameterHandler.Body<>(method, p, converter);
+
```
vanniktech: nit: unnecessary new line

JakeWharton: It's intentional to separate the branches.

retrofit/src/main/java/retrofit2/http/Tag.java
```diff
@@ -0,0 +1,39 @@
...
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Adds the argument instance as a request tag using the type as the key.
+ * <pre><code>
+ * &#64;GET("/")
+ * Call&lt;ResponseBody&gt; foo(@Tag String tag);
+ * </code></pre>
+ * Tag arguments may be {@code null} which will omit them from the request. Passing a parameterized
```
swankjesse: Nice docs!

neteinstein:
Awesome. 
@JakeWharton, do you have any ETA for a release with this?

JakeWharton:
No. There are other features that need added before a release will be
considered.

PR #3038, opened 2019-02-28
Add annotation for skipping the callback executor

JakeWharton:
This is going to be used for `Call` to `Call` adapter delegation which will back `suspend` functions and prevent them from inheriting the callback executor.

If I did it all again I'd eliminate the callback executor and instead make you register a call adapter which applied a callback executor based on either an annotation or to all calls. Guess that'll have to wait for 3.x.


retrofit/src/main/java/retrofit2/SkipCallbackExecutor.java
```diff
@@ -0,0 +1,50 @@
...
+ * @Get("user/{id}/token")
+ * Call<String> getToken(@Path("id") long id);
+ * }</pre>
+ *
+ * This annotation can also be used when a {@link CallAdapter.Factory} <em>explicitly</em> delegates
+ * to the built-in factory for {@link Call} via
+ * {@link Retrofit#nextCallAdapter(CallAdapter.Factory, Type, Annotation[])} in order for the
+ * returned {@link Call} to skip the executor. (Note: by default, a {@link Call} supplied directly
+ * to a {@link CallAdapter} will already skip the callback executor. The annotation is only useful
+ * when looking up the build-in adapter.)
```
swankjesse: nit: build-in -> built-in ?

retrofit/src/test/java/retrofit2/RetrofitTest.java
```diff
@@ -1341,6 +1344,62 @@
     verifyNoMoreInteractions(executor);
   }
 
+  @Test public void skippedCallbackExecutorNotUsedForSuccess() throws InterruptedException {
+    Executor executor = new Executor() {
+      @Override public void execute(Runnable command) {
+        fail();
```
swankjesse: cute

PR #3040, opened 2019-03-01
Back suspend methods with the CallAdapter for Call.

### LINKED ISSUE #3037: Ability to add custom adapter for suspend methods
As discussed in #2886:

Seemingly best way to handle server errors in Retrofit right now is to add custom adapter that intercepts all requests and then parses error body  (similarly to how [official retrofit2 sample](https://github.com/square/retrofit/blob/master/samples/src/main/java/com/example/retrofit/ErrorHandlingAdapter.java) intercepts exceptions). For example, errors can be easily handled by installing modified version of https://github.com/JakeWharton/retrofit2-kotlin-coroutines-adapter that parses and throws any server errors before creating Deferred instances.

However, with the new official coroutine support with `suspend` modifier, there is no way to attach any kind of adapter. Only way of parsing server exceptions is using `suspend Response<T>` everywhere (instead of just `suspend T`) and then manually parsing error body after every call to the retrofit methods, which is cumbersome.

@nenick has created fork of retrofit with that feature here: https://github.com/nenick/retrofit/tree/suspend-calladapter . Maybe something like that could be adopted into central Retrofit?

JakeWharton:
This allows the regular mechanism of composition of behavior without explicitly needing to model suspending functions in the API.

Closes #3037.

### DIFF
--- a/retrofit/src/test/java/retrofit2/KotlinSuspendTest.kt
+++ b/retrofit/src/test/java/retrofit2/KotlinSuspendTest.kt
@@ -33,6 +33,8 @@ import retrofit2.helpers.ToStringConverterFactory
 import retrofit2.http.GET
 import retrofit2.http.Path
 import java.io.IOException
+import java.lang.reflect.ParameterizedType
+import java.lang.reflect.Type
 
 class KotlinSuspendTest {
   @get:Rule val server = MockWebServer()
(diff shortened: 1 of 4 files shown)


retrofit/src/test/java/retrofit2/KotlinSuspendTest.kt
```diff
@@ -213,4 +215,62 @@ class KotlinSuspendTest {
...
+                      callback.onResponse(call, response)
+                    }
+                  }
+                })
+              }
+            }
+          }
+        }
+      }
+    }
```
JakeWharton: Look! It's Christmas in February!

swankjesse:
Nice example!

PR #3049, opened 2019-03-15
Upgrade to OkHttp 3.14.

swankjesse:
This causes Retrofit to itself require Java 8 or Android 5+. We may want to avoid
merging this PR until there is a feature in OkHttp 3.14 compelling enough to
break support for Android 4.x.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -50,7 +50,7 @@
 
     <!-- Dependencies -->
     <android.version>4.1.1.4</android.version>
-    <okhttp.version>3.12.0</okhttp.version>
+    <okhttp.version>3.14.0</okhttp.version>
     <kotlinx.coroutines.version>1.1.0</kotlinx.coroutines.version>
     <animal.sniffer.version>1.14</animal.sniffer.version>
 


JakeWharton:
Yeah I skipped 3.13.x for the same reason. I'd at least like to get 2.6.0 out before merging an upgrade.

swankjesse:
Lets close so this is out of the way. Can reopen later.

JakeWharton:
I should stop slacking on finishing 2.6.0...

lolsheeplol:
Would be nice to have okhttp 3.14 as it contains a fix for https://github.com/square/okhttp/issues/4570.  

That said, is there any problem with overriding the okhttp version in my own build.gradle?  

```
implementation 'com.squareup.retrofit2:retrofit:2.6.0'    //  bundles okhttp 3.12.0
implementation 'com.squareup.okhttp3:okhttp:3.14.2'
```

JakeWharton:
Go for it. I don't want to update and force Java 8 until https://github.com/square/retrofit/issues/3110 is fixed in OkHttp.

PR #3066, opened 2019-04-07
Replace unnecessary async usage by launch in KotlinSuspendTest

LouisCAD:


JakeWharton:
I prefer to keep is as `async` since it maps to the semantics of something which returns a value as a Retrofit method would.

PR #3069, opened 2019-04-07
Add BaseUrl parameter feature

nz2Dev:


JakeWharton:
Can you motivate why this is needed in addition to `@Url`?

nz2Dev:
Alows relative path in method annotations e.g `@GET("foo/path")` to be used. So in case we need different base url due to load balance, or dev/prod endpoints, etc. We then just pass it as a dependencie, and also keeps path to the method in the same place. Which in case of using `@Url` we should move out of the method declaration.
Here is example of what I'm trying to say:

```java
interface UrlInjectionUsersService { 
  @GET
  Call<ResponseBody> getUsers(@Url String url);
}

class Usage {
  void foo() {
    UrlInjectionUsersService service = ...
    service.getUsers("http://dev.example.com/api/users"); // We should take care of adding 'api/users' there
  }
}
```
```java
interface BaseUrlInjectionUsersService { 
  @GET("api/users")
  Call<ResponseBody> getUsers(@BaseUrl String baseUrl);
}

class Usage {
  void foo() {
    BaseUrlInjectionUsersService service = ...
    service.getUsers("http://dev.example.com/"); //Only changin a base url or "host" for our service
  }
}
```
`@BaseUrl` feature allows to keep our services interfaces in respect of knowing what method to use in order to get data, and leaves managing from where to get data to the outer layer. So there is no more needs to use workaround with `OkHttpInterceptor` or other approaches to change base url, which makes it less error prone. But it require extra layer of logic for that, as in case of `@Url`, but keeps relative url, and `@Path` features to work with inside the interface of service.

nz2Dev:
So I wanted it to be treated as an alternative to `@Url`.

nz2Dev:
For now, I see the next implementation applicable for changing the base URL in that way:
```java
interface WebService {
  @GET("api/method")
  Call<ResponseBody> method();
}

class WebServiceProvider {
  final Retrofit prototype = ...
  
  WebService getService(String baseUrl) {
    // can be cached by baseUrl as a key
    return prototype.newBuilder()
             .baseUrl(baseUrl)
             .build()
             .create(WebService.class);
  }
}

class Usage {
  final WebServiceProvider serviceProvider = ...  

  void foo(String baseUrl) { // base URL passed from the outer layer so we don't care
    serviceProvider.getService(baseUrl).method();
  }
}
```

But it can be simplified by removing additional complexity presented by `WebServiceProvider`
```java
interface WebService {
  @GET("api/method")
  Call<ResponseBody> method(@BaseUrl baseUrl);
}

class Usage {
  final WebService service = ...  

  void foo(String baseUrl) { // base URL sill passed from the outer layer so we don't care
    service.method(baseUrl);
  }
}
```

nz2Dev:
@JakeWharton Does that make sense for you?

iplanetcn:
The implement is elegant and which is very useful.

acefalobi:
@JakeWharton please merge this. It is very useful

karangoel16:
I really love this idea, and is somewhat required in multiple server environment.

mustafaersoyer:
Useful.

PR #3070, opened 2019-04-09
Support okhttp3.Headers for HeadersMap parameters.

### LINKED ISSUE #2696: Headers parameter annotation for okhttp3.Headers
There's no way to do `@HeaderMap Map<String, List<String>> headers` as a parameter right now. (Maybe that should be allowed? I can open a separate issue or PR later.)

How about a way to pass in okhttp3.Headers? (even easier to implement with https://github.com/square/okhttp/issues/3923)

I'm not sure what the annotation name would be. There's already the method annotation `@Headers`.

NightlyNexus:
Closes #2696.

### DIFF
--- a/retrofit/src/test/java/retrofit2/RequestFactoryTest.java
+++ b/retrofit/src/test/java/retrofit2/RequestFactoryTest.java
@@ -60,6 +60,7 @@ import retrofit2.http.QueryName;
 import retrofit2.http.Tag;
 import retrofit2.http.Url;
 
+import static java.util.Arrays.asList;
 import static java.util.Collections.emptyList;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertNull;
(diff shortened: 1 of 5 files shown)


NightlyNexus:
oops, will do.

retrofit/src/test/java/retrofit2/RequestFactoryTest.java
```diff
@@ -653,6 +654,46 @@
     }
   }
 
+  @Test public void getWithHeaders() {
+    class Example {
+      @GET("/search")
+      Call<ResponseBody> method(@HeaderMap okhttp3.Headers headers) {
```
JakeWharton: One last thing, can you also add a test that does both type to ensure they're combined instead of replaced?

NightlyNexus: Good idea. Will also fix the style violation tomorrow.

Issue #210, opened 2013-05-15
null param in @Query anotation 


kotya341:
When I try to send null value by GET request with query annotation @Query ,
I recieve  URL  like http://myUrl/?search=null&blabla 
But I think it would be correct not send null param at all.
Can I achieve this?

PR #3083, opened 2019-04-18
Adds Nullable annotation to Query documentation

ladykovington:
To reflect https://github.com/square/retrofit/issues/210


JakeWharton:
I think you've misinterpreted the original issue. The `@Nullable` annotation does not actually change behavior in any way. It's merely for documentation for your users. I don't want to give the impression that this is required so I'm going to close.

PR #3102, opened 2019-05-18
Remove serviceMethodCache synchronized lock

nitrohsu:
serviceMethodCache now is Thread-safe's ConcurrentHashMap, so the performance is better if delete the sync lock

### DIFF
--- a/retrofit/src/main/java/retrofit2/Retrofit.java
+++ b/retrofit/src/main/java/retrofit2/Retrofit.java
@@ -163,12 +163,10 @@ public final class Retrofit {
     ServiceMethod<?> result = serviceMethodCache.get(method);
     if (result != null) return result;
 
-    synchronized (serviceMethodCache) {
-      result = serviceMethodCache.get(method);
-      if (result == null) {
-        result = ServiceMethod.parseAnnotations(this, method);
-        serviceMethodCache.put(method, result);
-      }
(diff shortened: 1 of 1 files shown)


retrofit/src/main/java/retrofit2/Retrofit.java
```diff
@@ -163,12 +163,10 @@ private void eagerlyValidateMethods(Class<?> service) {
...
-    synchronized (serviceMethodCache) {
-      result = serviceMethodCache.get(method);
-      if (result == null) {
-        result = ServiceMethod.parseAnnotations(this, method);
-        serviceMethodCache.put(method, result);
-      }
+    result = serviceMethodCache.get(method);
+    if (result == null) {
+      result = ServiceMethod.parseAnnotations(this, method);
+      serviceMethodCache.put(method, result);
```
swankjesse: What we probably want here is to putIfAbsent so that if we race both callers get the same instance back.

nitrohsu: Yeah,the code block is same as putIfAbsent(), but this don't cause thread-safe problem,so 
 remove 'synchronized' is ok based on performance.

JakeWharton: It's not the same. Please update.

nitrohsu:
Error pull request

PR #3109, opened 2019-05-30
`URL` overload for `Retrofit.Builder#baseUrl`

SethX3:
Functionally does nothing, just somewhat cleaner to be able to do `baseUrl(URL)` than `baseUrl(URL#toString());` if handling a `URL` rather than `String`.
Used `HttpUrl.get(String)` as `HttpUrl.get(URL)` returns null instead of throwing an exception.


JakeWharton:
Squashed, added test, and merged. Thanks!

PR #3114, opened 2019-06-05
Add JAXB Converter in Document

joelhandwell:
By reading documentation, it gives impression that retrofit do not support JAXB without mentioning.


JakeWharton:
Thanks!

Issue #3117, opened 2019-06-07
Changelog of version 2.5.0 is misleading regarding Unit support


LouisCAD:
Quoting the changelog:
> New: Built-in support for Kotlin's Unit type. This behaves the same as Java's Void where the body content is ignored and immediately discarded.

If you try to use a suspending function returning no meaningly value (i.e. `Unit`), with a backend sending a 204 No-Content response for example, you end up with a *runtime* failure: the infamous `KotlinNullPointerException`.

To me, this is quite contradictory with changelog of version 2.5.0, and it's unclear how `Unit` is supported.

Also, [the test for Kotlin's `Unit`](https://github.com/square/retrofit/blob/74fc0ff863f771cde028e4bd36da70c568017a96/retrofit/src/test/java/retrofit2/KotlinUnitTest.java) is not actually written in Kotlin.

[The suspending functions support test](https://github.com/square/retrofit/blob/74fc0ff863f771cde028e4bd36da70c568017a96/retrofit/src/test/java/retrofit2/KotlinSuspendTest.kt) is in Kotlin, but it doesn't test no return value (`Unit`) in the service proxied by Retrofit. If it is not supported, it'd good to have it written clearly somewhere (along with new doc for Kotlin usage?), and a test that is asserting it fails with KNPE would be good. I lost hours of my time trying to understand why it doesn't pick the built-in `Unit` converter, and still don't have an answer, hours I'd have preferred to spend on something more useful. I prefer to not imagine how beginners would face this.

If support is not there yet, I can submit a doc update PR.

JakeWharton:
204 and 205 always return null and skip all converters for better or worse. Dupe of #2867.

> Also, the test for Kotlin's Unit is not actually written in Kotlin.

Retrofit doesn't care what language your service interfaces are written in since it parses the shape of Java bytecode through the Java reflection APIs.

> The suspending functions support test is in Kotlin, but it doesn't test no return value (Unit) in the service proxied by Retrofit.

There is no need to test this since converters are entirely orthogonal to adapters. The `Unit` converter will work in Kotlin the same as the `Optional` converter and every other built-in one. Testing every combination of converter and adapter would result in an exponential need.

egor-n:
For anyone looking for a sensible solution on how to keep using `Unit` with `suspend` functions and 204/205 responses:

```kotlin
@POST("...")
suspend fun someFun(): Response<Unit>
```

Just know that `body()` will return null as expected.

LouisCAD:
[...]

PR #3118, opened 2019-06-07
Specify that Unit is not supported in suspend fun

LouisCAD:
Related to #3117


JakeWharton:
This is an untrue statement.

JakeWharton:
See reply on issue.

PR #3119, opened 2019-06-07
Specify 204 and 205 don't work with suspend Unit

LouisCAD:


CHANGELOG.md
```diff
@@ -15,7 +15,7 @@ Version 2.6.0 *(2019-06-05)*
    Behind the scenes this behaves as if defined as `fun user(...): Call<User>` and then invoked with `Call.enqueue`.
    You can also return `Response<User>` for access to the response metadata.
 
-   Currently this integration only supports non-null response body types. Follow
+   **IMPORTANT: Currently this integration doesn't support Unit for HTTP 204 and 205 responses, and doesn't support nullable response body types.** Follow
```
JakeWharton: This has nothing to do with `Unit`. 204 and 205 skips all converters regardless of the declared body type. It also has little to do with `suspend`. The Rx adapters exhibit the same problem, for example.

LouisCAD: I understand. I still think it's important to note that, because many users of this library are not aware of that, including me, after years of using it, until today of course.

LouisCAD: This is about the changelog, telling what works, and what doesn't. Could be addressed for Rx adapters too in another PR if needed.

JakeWharton: I don't want to mention a bug that's neither related to nor specific to `Unit` or Kotlin suspend support in the change log for Kotlin suspend support.

LouisCAD: It's not a bug. It's warning that something API users may expect is not _supported_, with the intent to save them time as they update the library and their retrofit interfaces.

JakeWharton: Well it _is_ a bug because it can be fixed.

I don't see why we should mention this here and not in the change log entry which added the RxJava 2 adapter. Or any other adapter. It's not a problem with the new suspend feature and your ability to observe the behavior is unchanged from the previous version.

Here's another bug: non-IOException subtypes thrown from an OkHttp interceptor will cause your suspend functions to never resume because the exception is leaked to the background thread uncaught exception handler instead of being forwarded to the `Call` we wrap. Except, this isn't specific to suspend functions either. It'll happen with the RxJava adapter or the ListenableFuture adapter or just plain Call.

LouisCAD: I understand. I just edited the PR to be more specific about the cause.

PaulWoitaschek: I find it important to document this too. I assume this has cost a lot of users a lot of time and when making a post request, not declaring a return value makes totally sense.

Especially with things like this in the changelog:
`New: Built-in support for Kotlin's Unit type. This behaves the same as Java's Void where the body content is ignored and immediately discarded.`

Issue #951, opened 2015-07-09
Does Retrofit can specify a default field for POST request?


Folyd:
As we know,we can use `@Field` annotation to specify a field in a `@POST` method, like this:
```java
    @POST
    void doSomething(@Field String params,@Field String anotherParams)
```
I need specify a default value to a field in a POST request,but i have no idea to do it.

Does Retrofit support specify a default field for POST request?  If not,cloud you add a feature to support this?
```java
    @POST
    @DefaultField(params = "defaultValue")
    void doSomething(@Field String anotherParams)
```
Any comment is welcome.Thanks.:)

oliverhausler:
I am missing this as well. AWS for example uses thousands of default fields which need to be injected on every request. I admit retrofit is not a good choice for a non-REST conform api as AWS [which is not retrofit's fault], but I want to use the same library for all API endpoints in my project, and retrofit is a great choice for all others. I am currently using a custom client implementation where I add these things to the request before it is sent, but it's sort of a mess.

My suggestion would be to add a parameter to either use @DefaultField as mentioned here http://stackoverflow.com/questions/26697054/retrofit-how-to-send-a-post-request-with-constant-fields - but I would also like to see @DefaultFieldMap() where I can add a whole bunch of default fields to a request.

IlyaEremin:
+1 for this

svenjacobs:
+1

eginez:
+1!

akobylianskyi:
+1!

mix3d:
Yet another clamoring for this feature :D

tomasbjerre:
You may use [Java Method Invocation Builder](https://github.com/tomasbjerre/java-method-invocation-builder) for this.

Then you can annotate the interface like this:
```
@GenerateMethodInvocationBuilder
public interface BitBucketServerService {
 @GET("/rest/api/1.0/projects/{projectkey}/repos/{repositoryslug}/pull-requests?direction={direction}&at={at}&state={state}&order={order}&withattributes={withattributes}&withproperties={withproperties}")
 Call<BitbucketServerResponse<BitBucketServerPullRequest>> pullRequests(//
   @Default("PROJ") @Query("projectkey") String projectKey,//
   @Default("REPO") @Query("repositoryslug") String repositoryslug,//
   @Default("INCOMING") @Query("direction") String direction,//
   @Default("23") @Query("at") String at,//
   @Default("OPEN") @Query("state") String state,//
[...]

PR #3120, opened 2019-06-08
Allow default field parameters on method declarations

vitorhugods:
Satisfies #951.

I've done this more than four months ago and got me thinking how good is this approach. It solves the problem, but unfortunately I couldn't immerse myself into the project to understand it architecture better, so I'm not 101% sure if this was the best approach.

I'm very new to the project and open source contributing in big projects overall. This is my second contribution, so I opened this hoping at least I'll learn something more about Retrofit for the next times.


CoolMind:
Will this PR be merged?

vitorhugods:
@CoolMind I wouldn't put my money on it.

- I am not super happy with my implementation;
- You probably shouldn't be using Retrofit interfaces directly. You'll be happier with an abstract Repository or something. You could put the Form fields in there, instead of having them static in the annotations.

So, if you organise your code better, you won't need this anyway.
I'm closing this.

PR #3121, opened 2019-06-09
Test unit for API request

kmp90:
Test and fix for response request errors


JakeWharton:
Not sure what the goal here was?

PR #3142, opened 2019-07-02
Fix missing body in raw response for error

Jawnnypoo:
Context: for some of our tests, we use `Response.error()` to mock an error response from the server. We are attempting to then call `response.raw().body()` and it comes back as null, due to the fact that the body is not getting set on the raw OkHttp response.

### DIFF
--- a/retrofit/src/main/java/retrofit2/Response.java
+++ b/retrofit/src/main/java/retrofit2/Response.java
@@ -85,6 +85,7 @@ public final class Response<T> {
   public static <T> Response<T> error(int code, ResponseBody body) {
     if (code < 400) throw new IllegalArgumentException("code < 400: " + code);
     return error(body, new okhttp3.Response.Builder() //
+        .body(body)
         .code(code)
         .message("Response.error()")
         .protocol(Protocol.HTTP_1_1)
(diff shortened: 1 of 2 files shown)


retrofit/src/main/java/retrofit2/Response.java
```diff
@@ -85,6 +85,7 @@
   public static <T> Response<T> error(int code, ResponseBody body) {
     if (code < 400) throw new IllegalArgumentException("code < 400: " + code);
     return error(body, new okhttp3.Response.Builder() //
+        .body(body)
```
JakeWharton: This needs to be changed to mimic:

https://github.com/square/retrofit/blob/master/retrofit/src/main/java/retrofit2/OkHttpCall.java#L204

Jawnnypoo:
Okay, feedback addressed 👍

Jawnnypoo:
Any thoughts on this failure? https://travis-ci.org/square/retrofit/builds/553410971?utm_source=github_status&utm_medium=notification

JakeWharton:
Merged as 06978ccd. Thanks!

Jawnnypoo:
Awesome, thanks for taking it on 👍

PR #3145, opened 2019-07-04
Allow service interfaces to extend other interfaces

### LINKED ISSUE #504: Allow to configure retrofit to allow API interfaces to extend interfaces
The comment:

```
// Prevent API interfaces from extending other interfaces. This not only avoids a bug in
// Android (http://b.android.com/58753) but it forces composition of API declarations which is
// the recommended pattern.
```

in:

https://github.com/square/retrofit/blob/master/retrofit/src/main/java/retrofit/Utils.java

Sounds rather arbitrary. I think it should be configurable at the very least.

I for example have something very similar to the following. 

A base response class:

```
class GenericResponse {
   Boolean success
}
```

and an interface that defines how my API needs to be accessed:

```
interface GenericAPI<R extends GenericResponse> {
   void obtainData(
      String name, String version, Map<String, String> parameters, Callback<R> callback
   )
}
```

I latter extend that interface with some specific API like:

```
interface GetUsersAPI extends GenericAPI<UsersResponse> {
   @GET("/dataAPI/{version}/{name}")
   void obtainData(
      @Path("name") String name,
      @Path("version") String version,
      @QueryMap Map<String, String> parameters,
      Callback<UsersResponse> callback
   )
}
```

being UsersResponse an extension of GenericResponse. With something like that, I can implement an abstract client like:

```
abstract class GenericClient<R extends GenericResponse, C extends GenericConnector> implements Callback<R> { 

   protected void doRequest(
      Class<C> apiDefinition, 
      String name, String version, 
      Map<String, Object> params
   ) { 
      new RestAdapter.Builder().build().create(apiDefinition).obtainData(
         name, version, params, this
      )
   }
}
```

That client can work with any of my APIs, avoiding code duplication while having typing enforced by the compiler using generics. It can deal with any obtainData connector and with the common parts of every response. When implemented by a concrete class, thanks to the use of generics and inheritance, the data is transformed to the proper specific response class and when using an IDE with generics support (any, I guess :) is very convenient to use every method with the signature automatically transformed to the concrete classes declared by the implementor class.

I understand that composition is better for many things, but I don't see how to work with a generic connector the way I'm doing it, being able to assume that every connector will have an obtainData with the given parameters and having that enforced by the compiler.

I'm not saying is not possible - I'm saying I don't know how, and the restriction I'm finding is, in my opinion, quite poorly explained. 

I think something whose reason to be is something as arguable as "composition is the only possible, good way" should be configurable at least.

JakeWharton:
With a higher minSdkVersion, the bug which prevented this on Android no longer exists.

Closes #504.

### DIFF
--- a/retrofit/src/main/java/retrofit2/Utils.java
+++ b/retrofit/src/main/java/retrofit2/Utils.java
@@ -317,18 +317,6 @@ final class Utils {
     return ResponseBody.create(body.contentType(), body.contentLength(), buffer);
   }
 
-  static <T> void validateServiceInterface(Class<T> service) {
-    if (!service.isInterface()) {
-      throw new IllegalArgumentException("API declarations must be interfaces.");
-    }
-    // Prevent API interfaces from extending other interfaces. This not only avoids a bug in
-    // Android (http://b.android.com/58753) but it forces composition of API declarations which is
-    // the recommended pattern.
-    if (service.getInterfaces().length > 0) {
-      throw new IllegalArgumentException("API interfaces must not extend other interfaces.");
-    }
-  }
-
   static Type getParameterUpperBound(int index, ParameterizedType type) {
     Type[] types = type.getActualTypeArguments();
     if (index < 0 || index >= types.length) {
--- a/retrofit/src/test/java/retrofit2/RetrofitTest.java
+++ b/retrofit/src/test/java/retrofit2/RetrofitTest.java
@@ -85,6 +85,10 @@ public final class RetrofitTest {
   }
   interface Extending extends CallMethod {
   }
+  interface TypeParam<T> {
+  }
+  interface ExtendingTypeParam extends TypeParam<String> {
+  }
   interface StringService {
     @GET("/") String get();
   }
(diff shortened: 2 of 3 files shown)


retrofit/src/test/java/retrofit2/RetrofitTest.java
```diff
@@ -129,16 +129,16 @@
     assertThat(example.toString()).isNotEmpty();
   }
 
-  @Test public void interfaceWithExtendIsNotSupported() {
+  @Test public void interfaceWithExtendIsNotSupported() throws IOException {
```
ChristianBecker: Rename this method as `interfaceWithExtend` **is** supported after this change.

retrofit/src/main/java/retrofit2/Utils.java
```diff
@@ -323,18 +323,6 @@ static ResponseBody buffer(final ResponseBody body) throws IOException {
     return ResponseBody.create(body.contentType(), body.contentLength(), buffer);
   }
 
-  static <T> void validateServiceInterface(Class<T> service) {
-    if (!service.isInterface()) {
-      throw new IllegalArgumentException("API declarations must be interfaces.");
-    }
-    // Prevent API interfaces from extending other interfaces. This not only avoids a bug in
-    // Android (http://b.android.com/58753) but it forces composition of API declarations which is
-    // the recommended pattern.
```
oldergod: not anymore?

baoti:
test name is `interfaceWithExtendIsSupported`?

swankjesse:
Sadly there's more tests to write.

```
interface TacoApi extends Generic API<Taco> { }
```

JakeWharton:
I was hoping no one noticed that... because there's more work to do
implementation-wise for that as well.

swankjesse:
Wanna just explicitly forbid that case? Implemented interfaces are raw types, not parameterized types? That'll cover 98% of our users demands without needing to copy paste the type resolver code a 5th time.

JakeWharton:
For now, yes. I'll update this week.

retrofit/src/main/java/retrofit2/Utils.java
JakeWharton: 6 years of hearing people complain.. I don't care what they do at this point.

JakeWharton:
@swankjesse PTAL

joe-striedl:
Any idea on when this will make it into a release?

JakeWharton:
No estimate, no.

bullheadandplato:
still no?

PR #3178, opened 2019-07-25
Fixed range notation in docs comment

igorwojda:


JakeWharton:
This range does not include 300 hence the "up-to" parenthesis instead of the "up-to-and-including" square bracket.

PR #3183, opened 2019-07-30
Avoid synchronous checked exceptions in 'suspend fun's

### LINKED ISSUE #3128: UndeclaredThrowableException when using "suspend" modifier
Retrofit crashes with `UndeclaredThrowableException` when using `suspend` modifier and making multiple requests which are supposed to throw `IOException`.

Retrofit version: 2.6.0, 2.6.1-SNAPSHOT (20190612)

    class RetrofitUndeclaredThrowableExceptionBug {
    
       interface Service {
            @GET("/")
            suspend fun download()
        }
    
        @Test
        fun test() = runBlocking {
            val retrofit = Retrofit.Builder()
                .baseUrl("https://unresolved-host.com/") // An unresolved host to simulate IOException
                .build()
    
            val service = retrofit.create<Service>(Service::class.java)
    
            // First attempt works fine
            try {
                service.download()
            } catch (e: IOException) {
                println("Catched 1: $e")
            }
    
            // Second attempt sometimes crashes
            try {
                service.download()
            } catch (e: IOException) {
                println("Catched 2: $e")
            }
    
            // And this will crash for sure (sometimes sooner, sometimes later)
            repeat(Integer.MAX_VALUE) {
                try {
                    service.download()
                } catch (e: IOException) {
                    println("Catched 3: $e")
                }
            }
        }
    }

JakeWharton:
Because of a race condition caused by thread preemption, this can sometimes happen. Explicitly yield to ensure the exception is delivered asynchronously to the continuation.

Closes #3128.

### DIFF
--- a/retrofit/src/test/java/retrofit2/KotlinSuspendTest.kt
+++ b/retrofit/src/test/java/retrofit2/KotlinSuspendTest.kt
@@ -273,4 +273,26 @@ class KotlinSuspendTest {
     val body = runBlocking { example.body() }
     assertThat(body).isEqualTo("HiHiHiHiHi")
   }
+
+  @Test fun checkedExceptionsAreNotSynchronouslyThrown() = runBlocking {
+    val retrofit = Retrofit.Builder()
+        .baseUrl("https://unresolved-host.com/")
+        .addConverterFactory(ToStringConverterFactory())
+        .build()
+    val example = retrofit.create(Service::class.java)
+
+    server.shutdown()
+
+    // The problematic behavior of the UnknownHostException being synchronously thrown is
+    // probabilistic based on thread preemption. Running a thousand times will almost always
+    // trigger it, so we run an order of magnitude more to be safe.
+    repeat(10000) {
+      try {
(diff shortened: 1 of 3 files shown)


retrofit/src/test/java/retrofit2/KotlinSuspendTest.kt
```diff
@@ -273,4 +273,25 @@ class KotlinSuspendTest {
...
+    server.shutdown()
+
+    // The problematic behavior of the UnknownHostException being synchronously thrown is
+    // probabilistic based on thread preemption. Running a thousand times will almost always
+    // trigger it, so we run an order of magnitude more to be safe.
+    repeat(10000) {
+      try {
+        example.body()
+        fail()
+      } catch (e: IOException) {
```
swankjesse: nit: e -> `_`

JakeWharton: Done. Also added a comment since it's not clear how this test could fail.

PR #3220, opened 2019-10-01
Stream request bodies in first party converters

### LINKED ISSUE #3215: Stream bodies from first-party converters
We currently do not to guard against mutable inputs, but that doesn't seem worth optimizing for. Plus it defers even more work to the HTTP thread.

ZacSweers:
Resolves #3215 

I wasn't sure about some places that manually called flush(), so let me know if there's anything I should change in those cases

### DIFF
--- a/retrofit-converters/gson/src/main/java/retrofit2/converter/gson/GsonRequestBodyConverter.java
+++ b/retrofit-converters/gson/src/main/java/retrofit2/converter/gson/GsonRequestBodyConverter.java
@@ -20,16 +20,16 @@ import com.google.gson.TypeAdapter;
 import com.google.gson.stream.JsonWriter;
 import java.io.IOException;
 import java.io.OutputStreamWriter;
-import java.io.Writer;
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import okhttp3.MediaType;
 import okhttp3.RequestBody;
-import okio.Buffer;
+import okio.BufferedSink;
 import retrofit2.Converter;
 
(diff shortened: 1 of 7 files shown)


ZacSweers:
Rekicked this, let me know what y'all think

JakeWharton:
I'm not sure I'm comfortable making this change without a pretty significant version bump. The `Content-Length` header is no longer set and if you're doing any kind of request signature (such as those required by various AWS usage) it might up and break your whole setup.

ZacSweers:
Hmm, what about with an opt-in API like this?

```java
WireConverterFactory.create()
  .streamRequestBody() // Returns a new instance with this behavior enabled
```

JakeWharton:
Yeah it's not dissimilar from how the use of async HTTP is used in RxJava adapters. In 1.x and 2.x there's `create()` and `createAsync()`. In 3.x there's `create()` and `createSynchronous()`.

That said we're kinda on the precipice of a new major version anyway. OkHttp has abandoned supporting their Java-based version that Retrofit depends on so now so upgrading our dependency to OkHttp 4 means we get a chance to make some decisions. There's plenty I got wrong in 2.x, so is a 3.x warranted with some significant carrots like zero-reflection and KSP+MPP?

Give me a week or two to think about it. I've been mulling reflection-free and KSP+MPP for a few weeks now. At worst we can ship this then as factory overloads like you suggested.

ZacSweers:
Sounds good 👍

efrohnhoefer:
Given the current state of Retrofit, does it make sense to revisit this and create factory methods that will expose the ability to stream request bodies?

PR #3230, opened 2019-10-17
enqueue and execute using Result<T> in kotlin

demonar:
First time using mvn, will run the checks and commit changes if needed, let me know if you want to change anything in the style or functionality of the extensions. 

For reference: https://medium.com/@ar.demon/those-kotlin-handy-extensions-retrofit-meet-kotlin-and-result-6d4b149a8ee2?source=friends_link&sk=6cea08880fc71adc78dbfddc6e708120


JakeWharton:
Thanks for the PR, but we're waiting until we can use Kotlin's `Result` type directly. For now it's just awaiting stabilization here: https://github.com/square/retrofit/compare/jakew/result/2020-03-13

PR #3232, opened 2019-10-18
Lazily get gson TypeAdapter in GsonResponseBodyConverter

hqzxzwb:
`Gson.getAdapter` usually creates a TypeAdapter from `ReflectiveTypeAdapterFactory`, which uses reflection to scan all fields immediately, can cost a lot of time. And it makes thing worse that the scanning goes through types of fields recursively as well.
We usually use an asynchronous CallAdapter, because we want the time-costing work to be done on worker threads. Method `Converter.Factory.responseBodyConverter` is called on the caller thread (e.g. Android UI thread), while method `Converter.convert` is called on the scheduled worker thread. So moving the time-costing `Gson.getAdapter` call from `responseBodyConverter` to `convert` should reduce blocking time on the caller thread.

### DIFF
--- a/retrofit-converters/gson/src/main/java/retrofit2/converter/gson/GsonConverterFactory.java
+++ b/retrofit-converters/gson/src/main/java/retrofit2/converter/gson/GsonConverterFactory.java
@@ -61,8 +61,7 @@ public final class GsonConverterFactory extends Converter.Factory {
   @Override
   public Converter<ResponseBody, ?> responseBodyConverter(Type type, Annotation[] annotations,
       Retrofit retrofit) {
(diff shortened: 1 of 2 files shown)


hqzxzwb:
Is there anything else to do to get this reviewed? Do I need to add any profiling data or test cases? We were facing boot time problems in our android app related to this situation. A copy with this modification is being used in our project now and it works well. @JakeWharton

JakeWharton:
It was actually a design decision not to do this in the switch from Retrofit 1 to Retrofit 2. For one, we aren't going to just change the behavior of a single converter and would need to change them all. Since that's unlikely, we'll probably pursue a default method on `Converter.Factory` that makes any converter lazy. I just haven't had time to look into it yet.

PR #3236, opened 2019-10-30
Add request URL to HttpException's message

bherbst:
`HttpExceptions` can currently be difficult to track down depending on how an application makes HTTP calls and which call adapters the application is using.

For example, an application using the RxJava2 call adapter might see a stacktrace that looks something like this in Firebase or other similar reporting tools:

```
retrofit2.adapter.rxjava2.HttpException: HTTP 403 
       at retrofit2.adapter.rxjava2.BodyObservable$BodyObserver.onNext(BodyObservable.java:54)
       at retrofit2.adapter.rxjava2.BodyObservable$BodyObserver.onNext(BodyObservable.java:37)
       at retrofit2.adapter.rxjava2.CallEnqueueObservable$CallCallback.onResponse(CallEnqueueObservable.java:60)
       at retrofit2.OkHttpCall$1.onResponse(OkHttpCall.java:129)
       at okhttp3.RealCall$AsyncCall.execute(RealCall.java:174)
       at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:32)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
       at java.lang.Thread.run(Thread.java:764)
```

There is very little context for a developer to use to identify the cause since no client code is in the stack trace and the `HttpException` could have come from any Retrofit HTTP call.

This PR aims to address that problem by including the request URL in the exception's message. While clients could previously access that by catching the `HttpException` and performing that logging manually, this makes the default behavior for `HttpException` more helpful.

### DIFF
--- a/retrofit-adapters/guava/src/test/java/retrofit2/adapter/guava/ListenableFutureTest.java
+++ b/retrofit-adapters/guava/src/test/java/retrofit2/adapter/guava/ListenableFutureTest.java
@@ -68,7 +68,7 @@ public final class ListenableFutureTest {
       assertThat(e.getCause())
           .isInstanceOf(HttpException.class) // Required for backwards compatibility.
           .isInstanceOf(retrofit2.HttpException.class)
-          .hasMessage("HTTP 404 Client Error");
+          .hasMessage("HTTP 404 Client Error from " + server.url("/"));
     }
   }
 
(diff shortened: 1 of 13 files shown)


JakeWharton:
This has the potential to leak PII or private authentication keys.

PR #3254, opened 2019-11-22
Migrate forced suspension mechanism from yield to Dispatcher.Default

JakeWharton:
Since it's possible for certain dispatchers to completely avoid yielding, and currently the immediate dispatchers exhibit this behavior, we need an alternate mechanism of forcing suspension or UndeclaredThrowableExceptions will still be seen.

Retrofit does not have its own thread pool onto which we can defer resuming. Instead we rely to Dispatchers.Default and forcibly suspend the caller using low-level coroutine intrinsics.

Refs #3128. Refs https://github.com/Kotlin/kotlinx.coroutines/pull/1667. Cc @elizarov.

### DIFF
--- a/retrofit/src/main/java/retrofit2/KotlinExtensions.kt
+++ b/retrofit/src/main/java/retrofit2/KotlinExtensions.kt
@@ -18,8 +18,11 @@
 
 package retrofit2
 
+import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.suspendCancellableCoroutine
-import kotlinx.coroutines.yield
+import kotlin.coroutines.intrinsics.COROUTINE_SUSPENDED
+import kotlin.coroutines.intrinsics.intercepted
+import kotlin.coroutines.intrinsics.suspendCoroutineUninterceptedOrReturn
 import kotlin.coroutines.resume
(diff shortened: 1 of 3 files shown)


retrofit/src/main/java/retrofit2/KotlinExtensions.kt
```diff
@@ -97,7 +100,20 @@ suspend fun <T : Any> Call<T>.awaitResponse(): Response<T> {
...
+/**
+ * Force the calling coroutine to suspend before throwing [this].
+ *
+ * This is needed when a checked exception is synchronously caught in a [java.lang.reflect.Proxy]
+ * invocation to avoid being wrapped in [java.lang.reflect.UndeclaredThrowableException].
+ *
+ * The implementation is derived from:
+ * https://github.com/Kotlin/kotlinx.coroutines/pull/1667#issuecomment-556106349
+ */
+internal suspend fun Exception.suspendAndThrow(): Nothing {
```
swankjesse: neat

JakeWharton:
I've updated the tests to use a dispatcher which prevents `yield()` from working to avoid regressions.

PR #3295, opened 2020-02-13
Test may fail due to a different order of items in XML

testfixer:
Tests `xmlRequestBody` and `userSuppliedJaxbContext` in `JaxbConverterFactoryTest` depend on `RecordedRequest.getBody()` to get the request body. The test assertions use hard-coded strings. However, the test may fail due to a different order of serialization results. So tests may fail or pass without any modifications to source code.

This PR proposes to use XMLUnit to fix the problem. It ignores the order of XML elements to make the tests more robust.


swankjesse:
Have you observed such test failures? I’m under the impression our current set-up is deterministic in practice.

PR #3309, opened 2020-02-20
[RFC] Easier creation of new calls

christophehenry:
Creating new `Call<T>` is far too difficult. Currently, it requires creating a new class implementing `Call`. Vut this is not handy in the case you just want to modify the call at run-time according to custom annotations using adapters.

### DIFF
--- a/retrofit/src/main/java/retrofit2/OkHttpCall.java
+++ b/retrofit/src/main/java/retrofit2/OkHttpCall.java
@@ -29,8 +29,8 @@ import okio.Okio;
 
 import static retrofit2.Utils.throwIfFatal;
 
-final class OkHttpCall<T> implements Call<T> {
-  private final RequestFactory requestFactory;
+public final class OkHttpCall<T> implements Call<T> {
+  private final SimpleRequestFactory requestFactory;
   private final Object[] args;
   private final okhttp3.Call.Factory callFactory;
   private final Converter<ResponseBody, T> responseConverter;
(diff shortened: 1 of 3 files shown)


JakeWharton:
Can you explain your use case more?

christophehenry:
Sure! So I have an HTTP API that mainly work with short-lived tokens for authentication. This token must be retrieved by calling an endpoint secured by a basic authentication. I wanted to use custom annotation to:

* intercept the calls to the HTTP API and renew the token if necessary before performing the actual call
* dynamically set the authentication header when calling the token endpoint.

Now I know that the second use case can be handled by using a `@Header` parameter. But since I also have the first use case, my service implementation class has the credentials in its properties. I also know That both these use case can be handled by an interceptor, as described [here](https://futurestud.io/tutorials/retrofit-token-authentication-on-android), but not every endpoint requires the same authentication. So I would have to add a lot of logic to determine which endpoint is being called and which pre-processing it requires.

I ended up trying to distinguish the endpoints authenticated by a token from the token endpoint authenticated by a basic authentication using custom annotations.

swankjesse:
Does `Invocation` help?
https://square.github.io/retrofit/2.x/retrofit/retrofit2/Invocation.html

```

 class InvocationLogger implements Interceptor {
   @Override public Response intercept(Chain chain) throws IOException {
     Request request = chain.request();
     Invocation invocation = request.tag(Invocation.class);
     if (invocation != null) {
       System.out.printf("%s.%s %s%n",
           invocation.method().getDeclaringClass().getSimpleName(),
           invocation.method().getName(), invocation.arguments());
     }
   }
 }
```

christophehenry:
@swankjesse Oh nice! That works well! Maybe this use-case could be worth documenting?
Thank you both a lot for taking time to answer to me :+1:

PR #3310, opened 2020-02-21
FIX:  #3281 Don't call private Lookup constructor

### LINKED ISSUE #3281: Could not find class 'java.lang.invoke.MethodHandles$Lookup', referenced from method retrofit2.Platform.invokeDefaultMethod
When I added MultiDex to the Android project, In lower versions of Android like the 4.2 I get the following error
```
E/dalvikvm: Could not find class 'java.lang.invoke.MethodHandles$Lookup', referenced from method retrofit2.Platform.invokeDefaultMethod
```
Here's how I added Multiplex

My gradle module:
``` 
defaultConfig {
    multiDexEnabled = true
    // other code
}

dependencies {
    implementation 'com.android.support:multidex:1.0.3'

    implementation 'com.squareup.okhttp3:okhttp:1.12.7'

    implementation 'com.squareup.retrofit2:retrofit:2.7.0'
    implementation 'com.squareup.retrofit2:converter-gson:2.7.0'
    // other code
}

```

My application class:
```
public class BaseApplication extends MultiDexApplication {
 // other code
}
```

My Manifest:
```
<application
    android:name=".BaseApplication">
    <!-- other code -->
</application>
```

My retrofit clinte:
```
public class ApiClient {

    private static final String BASE_URL = BaseApplication.API_URL;
    private static Retrofit retrofit = null;

    public static Retrofit getClient() {

        if (retrofit == null) {
            retrofit = new Retrofit.Builder()
                    .baseUrl(BASE_URL)
                    .client(getOkHttpClient())
                    .build();
        }
        return retrofit;
    }

    public static OkHttpClient getOkHttpClient() {

        return new OkHttpClient.Builder()
                .readTimeout(20, TimeUnit.SECONDS)
                .connectTimeout(20, TimeUnit.SECONDS)
                .addInterceptor(new Interceptor() {
                                    @Override
                                    public Response intercept(@NonNull Interceptor.Chain chain) throws IOException {
                                        Request original = chain.request();
                                        Request request = original.newBuilder()
                                                .method(original.method(), original.body())
                                                .build();

                                    }
                                }
                ).build();
    }
}
```

**Some of the solutions I found in search were:**

 - Use retrofit version 2.2.0 or below

 - Disable Multiplex and resize the project to the point that it does not need to be multiplexed

But I don't want to use these methods because I always want to use the latest version of retrofit, also cannot reduce the project size

cogman:
This fixes #3281

This change is to sniff whether the current private lookup constructor is available.  If it isn't available, assume that we are running on a newer version of the JVM (in which case, the unreflectSpecial should work correctly).

I've added a test that I manually exercised using newer and older versions of the jdk.  Pre change, it fails on JVMs higher than 14.  Post change it works on all JVMs (I tested 8, 12, 13, and 14.  I did not test android, I assume it should still work.)

This change relies on the fix for https://bugs.openjdk.java.net/browse/JDK-8209005

### DIFF
--- a/retrofit/src/main/java/retrofit2/Platform.java
+++ b/retrofit/src/main/java/retrofit2/Platform.java
@@ -18,6 +18,9 @@ package retrofit2;
 import android.os.Build;
 import android.os.Handler;
 import android.os.Looper;
+
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
 import java.lang.invoke.MethodHandles.Lookup;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
@@ -48,9 +51,22 @@ class Platform {
   }
 
   private final boolean hasJava8Types;
+  private final boolean supportsInvokeSpecial;
 
   Platform(boolean hasJava8Types) {
     this.hasJava8Types = hasJava8Types;
+    if (hasJava8Types) {
+      boolean hasLookupPrivateChangedFromJava8 = false;
+      try {
+        Lookup.class.getDeclaredConstructor(Class.class, int.class);
+      } catch(NoSuchMethodException ex) {
+        hasLookupPrivateChangedFromJava8 = true;
+      }
+      supportsInvokeSpecial = hasLookupPrivateChangedFromJava8;
+    }
+    else {
+      this.supportsInvokeSpecial = false;
+    }
   }
 
   @Nullable Executor defaultCallbackExecutor() {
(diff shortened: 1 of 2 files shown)


JakeWharton:
Thanks I'll take a peek tomorrow

cogman:
Is there anything you need from me to help move this along?

JakeWharton:
Unfortunately we cannot use method handles because of Android. They rely on a new features of the Dalvik bytecode which are only available once your minimum API is 24 or 26.

I think we need to use a multi-release JAR instead so this code is only loaded on JVM 9+.

JakeWharton:
Here's a somewhat ridiculous sample we can steal from: https://github.com/apache/maven-compiler-plugin/tree/master/src/it/multirelease-patterns/singleproject-runtime

cogman:
I'm not familiar with Android development, but the code already had a fork for if this is the JVM vs Android. (That's essentially what the Boolean is doing ok the constructor).  The broken method in question isn't called by Android.

"Lookup" is only ever supposed to be accessed from the MethodHandles class (it's constructors are private).  My assumption was that this was safe to do since reflectively invoking the private lookup constructor was safe. 

Was that a bad assumption?

JakeWharton:
No it's a fine assumption! The problem is more subtle.

You are correct that at runtime (on Android) this code will not be used. The problem is compile-time. The bytecode needed to represent the MethodHandle interaction is only available on newer versions of the Dalvik bytecode format which forces the minimum supported API higher than our actual supported minimum. The compiler doesn't know that this code won't actually be executed at runtime and it has to include it which means forcing the higher version.

cogman:
Why wasn't this triggered beforehand? "unreflectSpecial" returns a MethodHandle object.  I've saved that into a temporary variable now, but the bytecode for invoking MethodHandle#bindTo existed prior to my change.

I could change it to chaining like it was before but I don't see how that avoids the bytecode problem.  I could also store off the Lookup object itself instead and go from there (as the more important thing here is how the Lookup object is created).

The main reason (which, admittedly is a bit selfish) I'm trying to avoid the multirelease jar path is that it is somewhat complex and doesn't necessarily play nice with the maven shade plugin (which we use).

cogman:
Also, what's the best way for me to test what is being forced as the minimum version?  Could this be captured with some sort of test?

JakeWharton:
Ok I looked into it more.

The usage in this PR and in the prior code uses only `MethodHandle` as API. This is fine for Android.

What doesn't work is when you have constant method handles which refer to something fixed. This is required for the JIT to see through the method handle and treat it like a normal invoke, whereas our dynamic usage is opaque to optimization. The static, constant-producing usage is what uses new bytecodes that only work with the new Dalvik bytecode format. Since we're not doing that, it's not a problem!

I will properly review this shortly and hopefully get it landed.

JakeWharton:
> what's the best way for me to test what is being forced as the minimum version

If we add a small Android sample with our minimum target the build would have failed (if we actually had the constant handle). Maven isn't supported for building Android. We'd have to migrate to Gradle, which would be fine, but I've been resisting simply because there hasn't been any reason to do such a migration.

JakeWharton:
Fixed up some things and merged in a separate PR (with attribution). I deleted the test since it's covered by the existing default method tests. Going to hopefully get CI working on 9+ soon, but it requires larger changes to the build.

Thanks for this!

cogman:
No problem.  Thanks for working with me to get this in.

PR #3311, opened 2020-02-21
Use computeIfAbsent method to simplify this method

cogman:
This is a minor change.  This relies on the internal locking mechanisms of the concurrent hash map instead of doing manual synchronization work.

Not that it matters, but it ends up being somewhat faster (this is not likely a hot path).  There is only a single key lookup instead of 2 in the worst case and it relies on CHMs internal locking scheme only (instead of having an extra one outside).

### DIFF
--- a/retrofit/src/main/java/retrofit2/Retrofit.java
+++ b/retrofit/src/main/java/retrofit2/Retrofit.java
@@ -183,17 +183,7 @@ public final class Retrofit {
   }
 
   ServiceMethod<?> loadServiceMethod(Method method) {
-    ServiceMethod<?> result = serviceMethodCache.get(method);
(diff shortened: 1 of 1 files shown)


JakeWharton:
Unfortunately this is only [available on][1] Android API 24 and newer whereas we support API 21+.

So it'll be a while before we can actually use this, despite being much nicer.

 [1]: https://developer.android.com/reference/java/util/concurrent/ConcurrentHashMap#computeIfAbsent(K,%20java.util.function.Function%3C?%20super%20K,%20?%20extends%20V%3E)

JakeWharton:
_Hopefully_ the Animal Sniffer check will fail the Travis CI build because of this...

PR #3334, opened 2020-03-12
Add Java 14+ support for default methods

### LINKED ISSUE #3281: Could not find class 'java.lang.invoke.MethodHandles$Lookup', referenced from method retrofit2.Platform.invokeDefaultMethod

JakeWharton:
The bug which required the reflective use of a hidden constructor is fixed in Java 14 and the normal lookup can just be used. Moreover, that private constructor is no longer even available.

Closes #3281. Closes #3310.

### DIFF
--- a/retrofit/src/main/java/retrofit2/Platform.java
+++ b/retrofit/src/main/java/retrofit2/Platform.java
@@ -18,6 +18,7 @@ package retrofit2;
 import android.os.Build;
 import android.os.Handler;
 import android.os.Looper;
+import java.lang.invoke.MethodHandles;
 import java.lang.invoke.MethodHandles.Lookup;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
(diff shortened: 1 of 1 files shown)


JakeWharton:
Manually testing on JDK 8-15 which requires a bunch of changes to the pom and Robolectric. Going to work on getting CI set up for 9-14 in the coming weeks.

PR #3338, opened 2020-03-18
Add Call.timeout() which returns the OkHttp Timeout

### LINKED ISSUE #2966: Support okhttp3.Call.timeout()
https://github.com/square/okhttp/blob/c273b3be385383842e44d25d6b7ca137ff6b1076/okhttp/src/main/java/okhttp3/Call.java#L84-L91

I have one endpoint in the retrofit `interface` that needs an increased timeout. Currently I have to create a modified `OkHttpClient`, and then create an additional `interface` with an endpoint that needs the increased timeout.

It would be great to have annotation for that

swankjesse:
Closes #2966

### DIFF
--- a/retrofit/src/main/java/retrofit2/OkHttpCall.java
+++ b/retrofit/src/main/java/retrofit2/OkHttpCall.java
@@ -26,6 +26,7 @@ import okio.Buffer;
 import okio.BufferedSource;
 import okio.ForwardingSource;
 import okio.Okio;
+import okio.Timeout;
 
 import static retrofit2.Utils.throwIfFatal;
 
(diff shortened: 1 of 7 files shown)


retrofit/src/main/java/retrofit2/OkHttpCall.java
```diff
@@ -58,28 +59,47 @@
...
+    } catch (IOException e) {
+      throw new RuntimeException("Unable to create call.", e);
+    }
+  }
+
+  /**
+   * Returns the raw call, initializing it if necessary. Throws if initializing the raw call throws,
+   * or has thrown in previous attempts to create it.
+   */
+  private okhttp3.Call getRawCall() throws IOException {
```
JakeWharton: Needs to be synchronized. Concurrent invocations can return different instances or see partial state.

I hope error-prone catches this because of the `@GuardedBy`...

JakeWharton: It did. Whew.

PR #3346, opened 2020-03-25
Small fixes from error-prone

JakeWharton:


retrofit/src/test/java/retrofit2/NonFatalError.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package retrofit2;
+
+final class NonFatalError extends Error {
```
Egorand: Hmm, what does error-prone complain about here?

JakeWharton: the test code was basically
```java
try {
  throw new Error();
  fail();
} catch (Error e) {
}
```
Since `fail()` will throw `AssertionError`, it will be caught by the `catch` block even if the behavior under test fails. By changing to a specific subtype and catching only that subtype we allow an `AssertionError` to propagate and correctly fail the test.

Egorand: Nice!

retrofit/src/test/java/retrofit2/RequestFactoryTest.java
```diff
@@ -793,6 +794,7 @@
     assertThat(request.body()).isNull();
   }
 
+  @Ignore("This test is valid but isn't validated by RequestFactory so it needs moved")
```
swankjesse: nit: needs to be moved ?

JakeWharton: Yep, but will do in a follow-up.

Issue #3372, opened 2020-04-25
A problem when use JacksonConverterFactory


ctlove0523:
problem：
com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "app_id" 

can modify `JacksonConverterFactory`'s private final field `mapper` like this：
~~~
mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
~~~

JakeWharton:
Create a mapper, configure this setting, and pass it into the factory through the static method which accepts a mapper.

ctlove0523:
@JakeWharton yes I understand

PR #3373, opened 2020-04-26
#3372

ctlove0523:


JakeWharton:
If you want this behavior, configure it on a mapper that you pass in. We're not going to prescribe settings that you must use. If you want it to be the default, pursuing it on Jackson is the right course of action.

PR #3387, opened 2020-05-09
add feature

CShiru:
My code will throw a IllegalArgumentException exception when user use query without string converter and the parameter type of method is not String .


swankjesse:
The current behavior is working as designed. Install the appropriate converter.

PR #3397, opened 2020-05-21
Use https for the apache license

PaulWoitaschek:


JakeWharton:
Now do the ~hundred or so others 🙏 

Thanks!

PaulWoitaschek:
Actually Im using https://github.com/jaredsburrows/gradle-license-plugin and iirc it was only in the latest version or so that it changed to non https.

PR #3400, opened 2020-05-24
Add documentation to clarify if ResponseBody.close() is necessary

### LINKED ISSUE #2950: Question: Do I need to close ResponseBody?
When Retrofit2 returns okhttp3.ResponseBody then it has to be closed in order to avoid memory leaks, right? It's something what I have just noticed in our code and am not sure.
```
@GET("accountId")
@WorkerThread
Call<ResponseBody> getUserData(@Header(NetworkConstants.Header.AUTHORIZATION) String accessToken);

...
Response<ResponseBody> response = interface.getUserData();
// do something with response
**response.close()**
```
https://stackoverflow.com/questions/53265862/do-i-need-to-close-response

tejas-n:
Closes #2950 and #2929


retrofit/src/main/java/retrofit2/Response.java
```diff
@@ -147,12 +147,23 @@ public boolean isSuccessful() {
     return rawResponse.isSuccessful();
   }
 
-  /** The deserialized response body of a {@linkplain #isSuccessful() successful} response. */
+  /**
+   * The deserialized response body of a {@linkplain #isSuccessful() successful} response.
+   *
+   * <p>If the body is of type {@link ResponseBody}, calling {@link ResponseBody#close close} is
```
swankjesse, line 153: I don't think this is helpful because it splits the type system. 

By following this doc you need to differentiate between responses that are created from streaming endpoints and responses that are not. If you write general purpose code to accept a `Response`, you need to clarify whether it's a streaming one. Also if you add `@Streaming` you create a resource leak.

It's simpler and more resilient to document that this must be closed under all circumstances.

tejas-n: Thanks for the reply. Yeah, that makes sense. So you are suggesting "If the body is of type ResponseBody, close() should always be called to avoid leaking resources"? Am I correct?

tejas-n: @swankjesse  I've updated the PR, can you let me know if that looks correct, please?

retrofit/src/main/java/retrofit2/Response.java
```diff
@@ -147,12 +147,22 @@ public boolean isSuccessful() {
     return rawResponse.isSuccessful();
   }
 
-  /** The deserialized response body of a {@linkplain #isSuccessful() successful} response. */
+  /**
+   * The deserialized response body of a {@linkplain #isSuccessful() successful} response.
+   *
+   * <p>If the body is of type {@link ResponseBody}, {@link ResponseBody#close()} must always be
+   * called to avoid leaking resources
```
JakeWharton, line 154: This is only true of `@Streaming` endpoints. I think this doc needs to live only on the `@Streaming` annotation and not here.

The one down below is still useful, however.

tejas-n: Thanks, that makes sense. I'll try to update it tonight.

PR #3404, opened 2020-05-28
Make jaxb converter able to process json content

MaximumQuiet:
Jaxb is just an API, and its implementations also may work with json (e.g. EclipseLink MOXy)

This PR adds two features:
1. Ability to provide properties for jaxb marshaller/unmarshaller, for example to [use json content with jaxb](https://www.eclipse.org/eclipselink/documentation/2.4/moxy/json002.htm)
2. Ability to choose content media-type to work with jaxb from converter factory

### DIFF
--- a/retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/ConvertingMediaType.java
+++ b/retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/ConvertingMediaType.java
@@ -2,5 +2,6 @@ package retrofit2.converter.jaxb;
 
 /** Request/response media type to convert by JAXB */
 enum ConvertingMediaType {
-  XML, JSON
+  XML,
+  JSON
 }
--- a/retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/ConvertingMediaType.java
+++ b/retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/ConvertingMediaType.java
@@ -1,5 +1,6 @@
 package retrofit2.converter.jaxb;
 
+/** Request/response media type to convert by JAXB */
 enum ConvertingMediaType {
   XML, JSON
 }
--- a/retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/ConvertingMediaType.java
+++ b/retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/ConvertingMediaType.java
@@ -1,7 +1,0 @@
-package retrofit2.converter.jaxb;
-
-/** Request/response media type to convert by JAXB */
-enum ConvertingMediaType {
-  XML,
-  JSON
-}
--- a/retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/ConvertingMediaType.java
+++ b/retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/ConvertingMediaType.java
@@ -1,0 +1,5 @@
+package retrofit2.converter.jaxb;
+
+enum ConvertingMediaType {
+  XML, JSON
+}
--- a/retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbConverterFactory.java
+++ b/retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbConverterFactory.java
@@ -30,8 +30,8 @@ import retrofit2.Converter;
 import retrofit2.Retrofit;
 
 /**
- * A {@linkplain Converter.Factory converter} which uses JAXB for XML and JSON. Concrete content media type
- * sets by {@link ConvertingMediaType}. All validation events are ignored.
+ * A {@linkplain Converter.Factory converter} which uses JAXB for XML and JSON. Concrete content
+ * media type sets by {@link ConvertingMediaType}. All validation events are ignored.
  */
 public final class JaxbConverterFactory extends Converter.Factory {
   static final MediaType XML = MediaType.get("application/xml; charset=utf-8");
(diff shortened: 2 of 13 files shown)


retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/ConvertingMediaType.java
```diff
@@ -0,0 +1,6 @@
+package retrofit2.converter.jaxb;
+
+/** Request/response media type to convert by JAXB */
+enum ConvertingMediaType {
+  XML, JSON
```
JakeWharton: If it's just an API then I don't see why we should restrict the formats to these two? For the [kotlinx.serialization converter][1] you can supply a custom `MediaType` since it's entirely format-agnostic.

 [1]: https://github.com/JakeWharton/retrofit2-kotlinx-serialization-converter/#usage

retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbConverterFactory.java
```diff
@@ -28,29 +30,94 @@
...
+    return new JaxbConverterFactory(context, null, marshalProps, unmarshalProps);
+  }
+
+  /** Create an instance using {@code context} for conversion with custom (un)marshaller properties. */
+  @SuppressWarnings("ConstantConditions")
+  public static JaxbConverterFactory create(
+          JAXBContext context,
+          ConvertingMediaType convertingMediaType,
+          Map<String, Object> marshalProps,
+          Map<String, Object> unmarshalProps) {
```
JakeWharton: These maps need checked for null

retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbConverterFactory.java
```diff
@@ -28,29 +30,94 @@
...
+  private final ConvertingMediaType convertingMediaType;
+  private final Map<String, Object> marshalProps;
+  private final Map<String, Object> unmarshalProps;
 
-  private JaxbConverterFactory(@Nullable JAXBContext context) {
+  private JaxbConverterFactory(
+      JAXBContext context,
+      @Nullable ConvertingMediaType convertingMediaType,
+      @Nullable Map<String, Object> marshalProps,
+      @Nullable Map<String, Object> unmarshalProps) {
```
JakeWharton, line 125: I would prefer we move default media type and the `emptyMap()`s to the call sites and keep all of these argument non-null.

retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbConverterFactory.java
```diff
@@ -60,7 +127,14 @@ private JaxbConverterFactory(@Nullable JAXBContext context) {
       Annotation[] methodAnnotations,
       Retrofit retrofit) {
     if (type instanceof Class && ((Class<?>) type).isAnnotationPresent(XmlRootElement.class)) {
```
JakeWharton, line 139: Is this annotation still present for JSON? That's a little weird.

retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbConverterFactory.java
```diff
@@ -28,29 +30,94 @@
...
+  @SuppressWarnings("ConstantConditions") // Guarding public API nullability.
+  public static JaxbConverterFactory create(JAXBContext context, ConvertingMediaType convertingMediaType) {
+    if (context == null) throw new NullPointerException("context == null");
+    if (convertingMediaType == null) throw new NullPointerException("convertingMediaType == null");
+    return new JaxbConverterFactory(context, convertingMediaType,null, null);
+  }
+
+  /** Create an instance using a default {@link JAXBContext}
+   * instance for conversion with custom (un)marshaller properties. */
+  public static JaxbConverterFactory create(Map<String, Object> marshalProps, Map<String, Object> unmarshalProps) {
```
JakeWharton: Need null checks for these arguments

retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbConverterFactory.java
```diff
@@ -28,29 +30,94 @@
...
+    return new JaxbConverterFactory(null, null, marshalProps, unmarshalProps);
+  }
+
+  /** Create an instance using a default {@link JAXBContext}
+   * instance for conversion with custom (un)marshaller properties. */
+  public static JaxbConverterFactory create(
+      ConvertingMediaType convertingMediaType,
+      Map<String, Object> marshalProps,
+      Map<String, Object> unmarshalProps) {
```
JakeWharton, line 84: Need null checks on these maps.

retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbConverterFactory.java
```diff
@@ -28,29 +30,94 @@
...
+          Map<String, Object> marshalProps,
+          Map<String, Object> unmarshalProps) {
+    if (context == null) throw new NullPointerException("context == null");
+    if (convertingMediaType == null) throw new NullPointerException("convertingMediaType == null");
+    return new JaxbConverterFactory(context, convertingMediaType, marshalProps, unmarshalProps);
   }
 
   /** If null, a new JAXB context will be created for each type to be converted. */
   private final @Nullable JAXBContext context;
+  /** If null, will be XML by default */
```
JakeWharton, line 115: This field is never null

retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbConverterFactory.java
```diff
@@ -60,7 +127,14 @@ private JaxbConverterFactory(@Nullable JAXBContext context) {
       Annotation[] methodAnnotations,
       Retrofit retrofit) {
     if (type instanceof Class && ((Class<?>) type).isAnnotationPresent(XmlRootElement.class)) {
-      return new JaxbRequestConverter<>(contextForType((Class<?>) type), (Class<?>) type);
+      switch (convertingMediaType) {
+        case XML:
+          return new JaxbXmlRequestConverter<>(contextForType((Class<?>) type), marshalProps, (Class<?>) type);
```
JakeWharton: You can extract the `contextForType(..)` code outside the switch.

retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbConverterFactory.java
```diff
@@ -69,7 +143,14 @@ private JaxbConverterFactory(@Nullable JAXBContext context) {
   public @Nullable Converter<ResponseBody, ?> responseBodyConverter(
       Type type, Annotation[] annotations, Retrofit retrofit) {
     if (type instanceof Class && ((Class<?>) type).isAnnotationPresent(XmlRootElement.class)) {
-      return new JaxbResponseConverter<>(contextForType((Class<?>) type), (Class<?>) type);
+      switch (convertingMediaType) {
+        case XML:
+          return new JaxbXmlResponseConverter<>(contextForType((Class<?>) type), unmarshalProps, (Class<?>) type);
```
JakeWharton: You can extract the `contextForType(..)` code outside the switch.

retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/ConvertingMediaType.java
MaximumQuiet: You're right. I kept in mind only an example with EclipseLink MOXy, xml & json.

retrofit-converters/jaxb/src/main/java/retrofit2/converter/jaxb/JaxbConverterFactory.java
MaximumQuiet: Just discovered that MOXy supports an option for an object marshal without @XmlRootElement. Then I will set this as the default behavior for json processing, and if the user's jaxb implementation does not support it, the user should add an annotation for successful marshalling

MaximumQuiet:
I created universal converters for non-xml content, however I didn't find other formats besides json supported by jaxb implementations

retrofit-converters/jaxb/src/test/resources/META-INF/services/javax.xml.bind.JAXBContext
```diff
@@ -0,0 +1 @@
+org.eclipse.persistence.jaxb.JAXBContextFactory
```
swankjesse, line 1: What does this do?

MaximumQuiet: By this file I changed an implementation of JAXBContextFactory to EclipseLink's one, because it supports both XML & JSON, so I was able to write tests for both formats.

https://docs.oracle.com/javase/9/docs/api/java/util/ServiceLoader.html

MaximumQuiet:
Yeah, I've no problem making a 3rd party converter, sounds good to me

PR #3411, opened 2020-06-02
Binary support implemented in JacksonConverterFactory

### LINKED ISSUE #3403: Support CBOR in JacksonConverterFactory
I am using [CBOR](http://cbor.io/) for request/response encoding. I am creating Jackson `ObjectMapper` with `CBORFactory` like this:

```java
ObjectMapper mapper = new ObjectMapper(new CBORFactory())
	.setVisibility(PropertyAccessor.FIELD, Visibility.ANY);
```

I then tried to use the existing `JacksonConverterFactory` like this:

```java
	.client(client)
	.baseUrl("http://" + host + ":" + port + "/")
	.addConverterFactory(JacksonConverterFactory.create(mapper))
	.build();
```

But this failed with an exception, because `JacksonConverterFactory` insists on reading/writing documents as text while CBOR requires binary I/O.

I have implemented custom converter as a workaround. It would be nice to have CBOR supported simply by passing in custom `ObjectMapper`. Switching to binary I/O should have no negative impact on JSON serialization AFAIK.

istvankovacs-priv:
Closes #3403

### What?
Binary support implemented in `JacksonConverterFactory` implemented.

### Why?
Currently, in `JacksonResponseBodyConverter` a `charStream` is used to convert the content of the response body. This requires a text formatted content to be able to serialize, making it impossible to send binary data. It is possible to resolve this with workarounds, but it would be nice to support this simply by passing in a custom ObjectMapper.

### How?
- `charStream` changed to `byteStream` in `JacksonResponseBodyConverter`
- `mediaType` made changeable via a constructor parameter in `JacksonRequestBodyConverter`
- an overload with two parameters of the `create` function is implemented in `JacksonConverterFactory`, where it is possible to set the `mediaType` via a nullable second parameter (other overloads calls this new function, to support backward compatibility)

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -20,6 +20,7 @@ robovm = "2.3.14"
 kotlinx-serialization = "1.6.3"
 autoService = "1.1.1"
 incap = "1.0.0"
+jackson = "2.16.1"
 
 [libraries]
 androidPlugin = { module = "com.android.tools.build:gradle", version = "8.2.2" }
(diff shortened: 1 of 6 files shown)


JakeWharton:
This looks pretty good. Needs some tests though before we can merge to prevent regression. If you can't get to them then I will eventually.

navtsiscavok:
Many thanks for the review! Feel free to implement the tests.

wellmonge (approved):

PR #3424, opened 2020-06-12
Prevent use of Java 8 classes on RoboVM

clydebarrow:
Retrofit used to work on RoboVM, but not all Java 8 features are supported by its VM, so Retrofit2 fails with "class not found" errors. This PR checks the VM and suppresses the use of Java 8 features when running on RoboVM.

### DIFF
--- a/retrofit/src/main/java/retrofit2/Platform.java
+++ b/retrofit/src/main/java/retrofit2/Platform.java
@@ -39,9 +39,14 @@ class Platform {
   }
 
   private static Platform findPlatform() {
-    return "Dalvik".equals(System.getProperty("java.vm.name"))
-        ? new Android() //
-        : new Platform(true);
+    switch (System.getProperty("java.vm.name")) {
+      case "Dalvik":
+        return new Android();
+      case "RoboVM":
+        return new Platform(false);
+      default:
+        return new Platform(true);
+    }
   }
 
   private final boolean hasJava8Types;
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,12 +64,26 @@ jobs:
         env:
           API_LEVEL: ${{ matrix.api-level }}
 
+  robovm:
+    runs-on: macos-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Validate Gradle Wrapper
+        uses: gradle/wrapper-validation-action@v1
+
+      - name: Run Tests
+        run: ./gradlew retrofit:robovm-test:robovmTest
+
   publish:
     runs-on: ubuntu-latest
     if: github.ref == 'refs/heads/master'
     needs:
       - jvm
       - android
+      - robovm
 
     steps:
       - name: Checkout
--- a/build.gradle
+++ b/build.gradle
@@ -7,6 +7,7 @@ buildscript {
       'okhttp': '3.14.9',
       'protobuf': '3.10.0',
       'jaxb': '2.3.1',
+      'robovm': '2.3.9',
   ]
   ext.deps = [
       'kotlinStdLib': "org.jetbrains.kotlin:kotlin-stdlib:${versions.kotlin}",
@@ -34,6 +35,7 @@ buildscript {
       'simpleXml': 'org.simpleframework:simple-xml:2.7.1',
       'wireRuntime': 'com.squareup.wire:wire-runtime:2.2.0',
       'jsoup': 'org.jsoup:jsoup:1.12.1',
+      'robovm': "com.mobidevelop.robovm:robovm-rt:${versions.robovm}",
   ]
 
   dependencies {
@@ -44,6 +46,7 @@ buildscript {
     classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.12'
     classpath 'ru.vyarus:gradle-animalsniffer-plugin:1.5.0'
     classpath 'gradle.plugin.com.github.sherter.google-java-format:google-java-format-gradle-plugin:0.9'
+    classpath "com.mobidevelop.robovm:robovm-gradle-plugin:${versions.robovm}"
   }
   repositories {
     mavenCentral()
--- a/retrofit/robovm-test/build.gradle
+++ b/retrofit/robovm-test/build.gradle
@@ -1,0 +1,20 @@
+apply plugin: 'java'
+apply plugin: 'robovm'
+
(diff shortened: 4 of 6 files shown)


JakeWharton:
How can we ensure we don't break this in the future?

retrofit/src/main/java/retrofit2/Platform.java
```diff
@@ -39,9 +39,14 @@ static Platform get() {
   }
 
   private static Platform findPlatform() {
-    return "Dalvik".equals(System.getProperty("java.vm.name"))
-        ? new Android() //
-        : new Platform(true);
+    switch(System.getProperty("java.vm.name")) {
```
JakeWharton, line 42:
```suggestion
    switch (System.getProperty("java.vm.name")) {
```
I suspect this is the source of the build failures

clydebarrow:
> How can we ensure we don't break this in the future?

What sort of scenario do you envisage would break it? I used the system property since it was already being used to identify Android. I do not expect the property value for RoboVM will change, though it is possible a future version will support Java 8.

Another option besides checking the system property would be to check for Java 8 specific classes or methods - the particular failure I encountered was `java.lang.reflect.Method.isDefault` so that could be used in a try/catch to determine if it is available and set `hasJava8Types` accordingly.

JakeWharton:
Well I obviously broke it before, and I could simply delete the `case` statement after this PR and it would break again. If we want it to keep working, we need something which validates the correct behavior.

clydebarrow:
Well, then it needs a test phase with a 1.7 JRE. Testing against RoboVM itself is probably impractical, so that suggests to me that the condition should identify the running JRE as pre 1.8 rather than as RoboVM in particular. I can amend the PR to do that. I have no idea at this point of how to go about adding a JRE1.7 test phase but I assume that it will be a case of copying the 1.8 pattern and changing the JRE.

clydebarrow:
So adding 1.7 to the list of Java versions in build.yml seemed like the obvious thing to do, but this fails because gradle 6.3 requires Java 8. Downgrading gradle would not be feasible, but it should be possible to run gradle with 1.8 and run the tests with 1.7. I'm not sure how to do that in the github actions context though.

JakeWharton:
In general we don't support Java 1.7 in either Retrofit or OkHttp because of how old it is. Is there some other RoboVM-specific build we can run?

clydebarrow:
> Is there some other RoboVM-specific build we can run?

 No, RoboVM doesn't actually use a JVM - it has an AOT compiler that compiles to native code. It does support Java8 language features, but not the full API. Its runtime supports iOS (which I assume is impractical to test via Github CI) and MacOs, but the latter is not actively maintained so I have no idea if it would be worth trying to use that.

> In general we don't support Java 1.7 

Well, yeah, but Android is supported back to API 21, which is 1.7. So the code will apparently run on 1.7 as long as the `Platform.hasJava8Types` flag is unset. The RoboVM Java API is very close to what's supported by Android 21.

I realise that a future version of Retrofit might abandon any pretense at supporting older Java versions, but right now it does work, so unless you have any strong objections I think adding a 1.7 test phase will be sufficient for the objective of preserving RoboVM compatibility.

I am fiddling with this in my own fork and should have a solution shortly to add a 1.7 specific test phase. I have not found any show-stoppers yet, I'm mainly just climbing the Github workflow learning curve, which is useful anyway.

clydebarrow:
Well there is a show-stopper. Running the tests with JVM1.7 is not possible due to class file version. Building with a lower target version than 1.8 isn't possible either. So back to square one.

swankjesse:
My interpretation here is that RoboVM is a new platform altogether; different from the JVM and from Android. Cool!

Retrofit and our other projects might support this platform by accident, but unless we test on it I don't want to give our users the impression that it's anything more than that. In particular I fear somebody adopts Retrofit on RoboVM and we break them on a patch release. Or we have a security problem on that platform because we make an assumption that doesn't hold for TLS or DNS or the filesystem on that platform.

I don't think we should official support RoboVM; at least not yet. In particular I don't think we should give our users the expectation that things working on RoboVM today will continue working in the future. And we should close bugs that reproduce there but not on our supported platforms. This isn't intended to discourage RoboVM use or development; only to be transparent in our ability to test and support a platform we don't use.

Next steps: I don't mind landing changes that help RoboVM but don't hurt other platforms. But if the RoboVM could find a way to make our tests fail if we regress, that's much better. Perhaps a CI target or an AnimalSniffer target?

clydebarrow:
Yeah, this is definitely intended to be viewed only as "best effort". But while the support for older Android versions is maintained the code should work fine on RoboVM since the two VMs have similar capabilities - Android 21 supports the 1.8 language features via desugaring, RoboVM via its compiler. Neither supports the full 1.8 API.

So my thoughts now are to revert to a specific check  for RoboVM when creating the `Platform` object, and add a test case that simply confirms that the `nonJava8Types` flag is set when that check's conditions are met.

The other options from my point of view would be to revert to Retrofit v1 - but since that now has a support horizon I would like to be able to use v2 - or to maintain my own fork that differs by just a few lines of code. I'm willing to do that but would rather not.

JakeWharton:
We can definitely run iOS and Mac OS tests on GitHub actions. So if the former is what's supported the most, let's make a small integration test similar to the ones in retrofit/android/ in retrofit/robovm/ and run them on CI.

clydebarrow:
Well, that was easier than I expected. The only odd thing is that when I verified that the CI run fails if the RoboVM test is broken, the CI log was odd - it shows the build step as cancelled when in fact it completed successfully and it was the Run Tests step that failed (non-zero exit status from the test program.)

I did see some references elsewhere to this behaviour specifically relating to the MacOs environment in Github actions. It would be nicer if the log showed the test as failing.

Screenshot 2020-06-15 at 12 00 17 pm

clydebarrow:
Seems happier if the test is run from gradle rather than as a separate step.

This should be ready to merge now.

JakeWharton:
Thanks! This is enough for me to run with. I think I'm going to change it so that we run a simple HTTP server locally on the host and make an actual full end-to-end HTTP request. I greatly prefer that over just checking we choose the right platform setup since it means we can guarantee the whole stack works.

clydebarrow:
Sounds good. Let me know what else I can do.

JakeWharton:
I decided it wasn't worth my time and what you had was fine. Tiny tweaks, squashed, and will merge on green.

JakeWharton:
Thanks!

clydebarrow:
Great!

PR #3431, opened 2020-06-27
Change forced suspension mechanism to try yield first

1zaman:
Instead of dispatching directly on the default dispatcher, try yielding first. If yield does not suspend, then fall back to the previous approach.

Context: #3254

### DIFF
--- a/retrofit/src/main/java/retrofit2/KotlinExtensions.kt
+++ b/retrofit/src/main/java/retrofit2/KotlinExtensions.kt
@@ -20,8 +20,11 @@ package retrofit2
 
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlinx.coroutines.yield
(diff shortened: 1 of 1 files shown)


JakeWharton:
Why?

1zaman:
Better performance, less context switches. On the other hand, I suppose this is a rare case, and doesn't need optimization. Feel free to close it if you think this adds unnecessary complexity.

JakeWharton:
> On the other hand, I suppose this is a rare case, and doesn't need optimization.

If this codepath were taken on every exception I would consider it, but it's only taken when an exception is thrown so quickly that it actually happens before the caller can suspend. This effectively only happens when your network is disconnected on requests after the first one where the DNS failure is cached locally.

There's more details here: https://jakewharton.com/exceptions-and-proxies-and-coroutines-oh-my/

Since this case is _extremely_ rarely hit, and it's already in an exception path, I don't think it's worth the added complexity of maintaining a more elaborate mechanism.

PR #3468, opened 2020-10-02
Allowing Unit return type

JohnOberhauser:
### Problem to solve
When making a request and you are expecting a 204, you have to use `Response<Unit>` for your return type, otherwise a `KotlinNullPointerException` is thrown.  Error handling with `Response<Unit>` requires checking `isSuccessful` whereas error handling with a data model uses a try catch block since an `HttpException` is thrown on error.  It would be nice to have the option to use try catch error handling for calls that are expecting a 204 without any workarounds.

This does not work for nullable types, only strictly `Unit`.

### Description
- checks if the raw return type is `Unit`.  If it is, set `continuationBodyNullable` to true so that the function `KotlinExtensions.awaitNullable` is used instead of `KotlinExtensions.await`
- added unit tests for this case

### DIFF
--- a/retrofit/src/main/java/retrofit2/HttpServiceMethod.java
+++ b/retrofit/src/main/java/retrofit2/HttpServiceMethod.java
@@ -23,6 +23,7 @@ import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import javax.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


lassem:
Yes please! Can we have this one. 

We've semi circumvented this problem by utilizing a nullable object which we then just ignore.

JakeWharton:
Fixed by #3702

PR #3471, opened 2020-10-04
Lowered sdk_int from 24 to 23

SnoopyCodeX:


JakeWharton (commented):
This is used to gate default methods which only work on 24+

Issue #3474, opened 2020-10-05
When using suspending async/await, stacktrace contains no application code


ghus-raba:
Can be reproduced by simple test:

```kotlin
  @Test fun await404() {
    val retrofit = Retrofit.Builder()
            .baseUrl(server.url("/"))
            .addConverterFactory(ToStringConverterFactory())
            .build()
    val example = retrofit.create(Service::class.java)

    server.enqueue(MockResponse().setResponseCode(404))

    try {
      runBlocking {
        val deferred = async { example.body()  }

        deferred.await()
      }
      fail()
    } catch (e: HttpException) {
      val writer = StringWriter()
      e.printStackTrace(PrintWriter(writer))

      val trace = writer.toString()

      assertThat("KotlinSuspendTest" in trace).isTrue()
      assertThat("await404" in trace).isTrue()
    }
  }
```

By default the stacktrace will look like:

```
retrofit2.HttpException: HTTP 404 Client Error
	at retrofit2.KotlinExtensions$await$2$2.onResponse(KotlinExtensions.kt:53)
	at retrofit2.OkHttpCall$1.onResponse(OkHttpCall.java:161)
	at okhttp3.RealCall$AsyncCall.execute(RealCall.java:174)
	at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:32)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```
This is not very helpful, as we don't know which call caused the exception. That can be fixed by `HttpException` implementing [CopyableThrowable](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-copyable-throwable/), which would lead to following trace:

```
retrofit2.HttpException: HTTP 404 Client Error
	(Coroutine boundary)
	at retrofit2.KotlinSuspendTest$await404$1$deferred$1.invokeSuspend(KotlinSuspendTest.kt:190)
Caused by: retrofit2.HttpException: HTTP 404 Client Error
	at retrofit2.KotlinExtensions$await$2$2.onResponse(KotlinExtensions.kt:53)
	at retrofit2.OkHttpCall$1.onResponse(OkHttpCall.java:161)
	at okhttp3.RealCall$AsyncCall.execute(RealCall.java:174)
[...]

PR #3475, opened 2020-10-05
Make HttpException implement CopyableThrowable + add test

ghus-raba:
Fixes https://github.com/square/retrofit/issues/3474


retrofit/src/main/java/retrofit2/HttpException.java
```diff
@@ -15,11 +15,13 @@
...
 package retrofit2;
 
+import kotlinx.coroutines.CopyableThrowable;
+
 import java.util.Objects;
 import javax.annotation.Nullable;
 
 /** Exception for an unexpected, non-2xx HTTP response. */
-public class HttpException extends RuntimeException {
+public class HttpException extends RuntimeException implements CopyableThrowable<HttpException> {
```
ghus-raba, line 24: Ok, found out this is not working for pure java build.

retrofit/src/main/java/retrofit2/KotlinHttpException.kt
```diff
@@ -0,0 +1,14 @@
+package retrofit2
+
+import kotlinx.coroutines.CopyableThrowable
+
+class KotlinHttpException(private val response: Response<*>) : HttpException(response),
+    CopyableThrowable<KotlinHttpException> {
+
+    override fun createCopy(): KotlinHttpException {
+        val result = KotlinHttpException(response)
+        result.initCause(this)
```
JakeWharton, line 10: Why do you pass `this` as the cause here? Shouldn't this be `cause`?

JakeWharton (commented):
Unfortunately `CopyableThrowable` is not stable API so we cannot depend on it. As soon as it goes stable we can land this change.

retrofit/src/main/java/retrofit2/KotlinHttpException.kt
JakeWharton: I made this change, but I see the coroutines docs use `this`, so I'll revert.

iamarjun:
@ferinagy @frantisek-nagy just created a fork with your changes and it seems like I still don't get the stack trace of the call site, am I missing anything?

Goooler (approved):

ferinagy:
> @ferinagy @frantisek-nagy just created a fork with your changes and it seems like I still don't get the stack trace of the call site, am I missing anything? Not sure how tests are working fine.

Hi, maybe you are hitting this case? https://github.com/Kotlin/kotlinx.coroutines/issues/2607

iamarjun:
> > @ferinagy @frantisek-nagy just created a fork with your changes and it seems like I still don't get the stack trace of the call site, am I missing anything? Not sure how tests are working fine.
> 
> Hi, maybe you are hitting this case? [Kotlin/kotlinx.coroutines#2607](https://github.com/Kotlin/kotlinx.coroutines/issues/2607)

Oh, I see, any workaround for this?

EnginnerDog:
I think I just saw a unicorn in this PR.

PR #3477, opened 2020-10-07
Added OSGi metadata via build

swimmesberger:
This is analogous to the implementation in https://github.com/square/okhttp/pull/6282

For retrofit there are currently following issues:
- okhttp/okio is still on a version which do not have OSGi metadata (therefore the test fails)
- The kotlinx.coroutines (https://github.com/Kotlin/kotlinx.coroutines) stuff is not OSGi compatible, therefore I made those packages optional for now
- I can't get both "retrofit-converters:guava" and "retrofit-adapters:guava" work together in my test because gradle seems to only copy one of those. It seems as if both of these are defined as project(...) dependencies gradle only adds the first one as an retrievable dependency (maybe a hasdCode/equals conflict?). I have found a similar issue on [stackoverflow](https://stackoverflow.com/questions/28458511/gradle-confusion-dependency-with-duplicate-sub-project-names)
- I can't test moshi/simplexml/wire in the junit test because I haven't found OSGi bundles for those dependencies

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,6 +8,7 @@ buildscript {
       'protobuf': '3.10.0',
       'jaxb': '2.3.1',
       'robovm': '2.3.9',
+      'bnd': '5.1.2'
   ]
   ext.deps = [
       'kotlinStdLib': "org.jetbrains.kotlin:kotlin-stdlib:${versions.kotlin}",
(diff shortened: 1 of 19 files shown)


p4535992:
+1 for this

Issue #3231, opened 2019-10-17
Publish a BOM


mikeholler:
What kind of issue is this?

 - [ ] Question. This issue tracker is not the place for questions. If you want to ask how to do
       something, or to understand why something isn't working the way you expect it to, use Stack
       Overflow. https://stackoverflow.com/questions/tagged/retrofit

 - [ ] Bug report. If you’ve found a bug, spend the time to write a failing test. Bugs with tests
       get fixed. Here’s an example: https://gist.github.com/swankjesse/6608b4713ad80988cdc9

 - [x] Feature Request. Start by telling us what problem you’re trying to solve. Often a solution
       already exists! Don’t send pull requests to implement new features without first getting our
       support. Sometimes we leave features out on purpose to keep the project small.
--------

Hey there! Love the project, but I would love it if there was a BOM file that is published for each version. If you're not familiar with a BOM,it stands for "Bill of Materials" it helps your Gradle (in my case, but it's originally a Maven thing) file look like this:

```kotlin
implementation(platform("com.squareup.retrofit2:retrofit-bom:2.6.1") )
implementation("com.squareup.retrofit2:retrofit")
implementation("com.squareup.retrofit2:converter-jackson")
```

Instead of this:

```kotlin
val retrofitVersion by extra { "2.6.1" }

implementation("com.squareup.retrofit2:retrofit:$retrofitVersion")
implementation("com.squareup.retrofit2:converter-jackson:$retrofitVersion")
```

Notice how you just leave the versions off when you use a BOM. This can help reduce the number of dependency compatibility surprises one can encounter, especially if a transitive dependency brings in a newer version of one of the components (it'll be reduced to the BOM's version).

Here's a deeper dive into how a BOM works:

https://howtodoinjava.com/maven/maven-bom-bill-of-materials-dependency/

tobias-:
Just a friendly reminder that according to https://github.com/square/retrofit/pull/2944, BoM was wating for Gradle conversion. Now that the Gradle conversion is done, I'd love for a platform/BoM project to be published.

Could possibly be as easy as:
```
plugins {
    `java-platform`
}

dependencies {
    constraints {
        rootProject.subprojects.onEach {
            api("${it.group}:${it.name}:$version")
        }
    }
}
```

JakeWharton:
PR welcome!

kid1412621:
finally! When is it available in maven repo?

majid-vaghari:
[...]

PR #3487, opened 2020-10-31
Add platform module for publishing BOM

nsuriadi:
[#3231](https://github.com/square/retrofit/issues/3231).

I've updated all subprojects to use the `maven-publish` plugin so the platform module can use the `POM_ARTIFACT_ID` of each subproject as opposed to project name when generating the BOM and its dependencies. 

I tried to keep existing properties and functionality from the previous maven publish intact (e.g. excluding directories named `internal` and suppressing doclint warnings). Let me know if this isn't ideal.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -95,7 +95,7 @@ jobs:
           java-version: 14
 
       - name: Upload Artifacts
-        run: ./gradlew uploadArchives
+        run: ./gradlew publish
         env:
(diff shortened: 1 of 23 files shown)


madorb:
any movement on this? would love to see a bom

realdadfish:
Ah, found that too late. Figured out that the existing plugin can eventually be used starting with 0.17.0 and created another PR for it. But I guess going native Gradle might be the better option actually...

Issue #3451, opened 2020-08-13
R8 rules setup outdated / incoherent on website


4face-studi0:
What kind of issue is this?

 - [ ] Question. This issue tracker is not the place for questions. If you want to ask how to do
       something, or to understand why something isn't working the way you expect it to, use Stack
       Overflow. https://stackoverflow.com/questions/tagged/retrofit

 - [ ] Bug report. If you’ve found a bug, spend the time to write a failing test. Bugs with tests
       get fixed. Here’s an example: https://gist.github.com/swankjesse/6608b4713ad80988cdc9

 - [ ] Feature Request. Start by telling us what problem you’re trying to solve. Often a solution
       already exists! Don’t send pull requests to implement new features without first getting our
       support. Sometimes we leave features out on purpose to keep the project small.

- [x] Other: outdated / incoherent documentation on https://square.github.io/retrofit/

--- 

## Expected behavior
[website](https://square.github.io/retrofit/) and GitHub readme to give same instruction about R8 rules setup, probably same as stated in the GitHub readme

## Current behavior
Website says
> If you are using R8 or ProGuard add the options from this file.

GitHub readme says
> If you are using R8 the shrinking and obfuscation rules are included automatically.
>
> ProGuard users must manually add the options from retrofit2.pro.

_Sorry if this isn't the proper place for this, just didn't know any better one_

PR #3489, opened 2020-11-02
Update website Proguard rule section to be same as Readme

kozaxinan:
Website and Readme have different information about shrinking rules. Information in readme is more accurate. This pr updates website to be same as readme.

Fixes https://github.com/square/retrofit/issues/3451


JakeWharton:
Thanks! Deployed!

PR #3501, opened 2020-12-08
POC of logging api result

Jintin:
We need a high level logging of api result above okhtttp logging.
Since some api call is binary like protobuf, it will be good if retrofit can provide a feature to enable object level of logging like the poc pr here.


Jintin:
btw I guess Android related build failed is normal?

PR #3562, opened 2021-05-17
Create CODE_OF_CONDUCT.md

sambhavsaxena:


JakeWharton:
We are covered under Square's blanket open source code of conduct: https://developer.squareup.com/blog/open-source-code-of-conduct/. But we could probably do a better job of reporting that on all of our projects.

PR #3563, opened 2021-05-21
Add keep rule for kotlin.coroutines.Continuation

mkj-gram:
I am unsure if the factory that checks for suspend functions relies on the name. If not, and the library do not need the name or location, this could be `-keep,allowobfuscation` instead.


JakeWharton:
Sorry I missed this. We do not rely on the name so we can allow obfuscation. Also you'll need to rebase since Christoffer also sent a PR to add a rule.

The check for coroutines is the very last attempt to match a method signature so we compare against a class literal which has a try/catch:
```java
try {
  if (Utils.getRawType(parameterType) == Continuation.class) {
    isKotlinSuspendFunction = true;
    return null;
  }
} catch (NoClassDefFoundError ignored) {
}
throw parameterError(method, p, "No Retrofit annotation found.");
```

mkj-gram:
I rebased (I think) and also allowed for obfuscation and shrinking.

JakeWharton:
Thanks!

Simay15:
android.enableR8.fullMode=true

PR #3574, opened 2021-06-15
HttpException no longer extends RuntimeException

gyrraf:
I don't see a reason why HttpException has to be a RuntimeException. To me the HttpException looks like an exception that should be handled in any app that uses this library and therefore shouldn't be a RuntimeException.


JakeWharton:
This is a source and binary incompatible change that we cannot make even if we wanted to.

I don't think this change is particularly useful otherwise since the exception is not used in a `throws` portion of a method declaration by the library. If you are using it like that in your application it's probably best to coerce it to your own exception type which does not carry raw response data.

PR #3579, opened 2021-06-22
Add keep rule to retain generic signature of Call

### LINKED ISSUE #3580: Inadequate keep rules with R8 full mode version 3.0
R8 version 3.0 started to remove generic signatures from items that are not matched by a -keep rule in full mode (see also https://r8.googlesource.com/r8/+/744d742137a82656b8ee27513f975e0528aecd78). The motivation for this change is to avoid that all generic signatures are retained in programs, when only a few are typically required for reflection.

As a result of this change, an extra `-keep` rule for `retrofit2.Call` is required to keep its generic signature.

I have opened PR https://github.com/square/retrofit/pull/3579 to resolve this.

christofferqa:
R8 version 3.0 started to remove generic signatures from items that are not matched by a `-keep` rule in full mode (see also https://r8.googlesource.com/r8/+/744d742137a82656b8ee27513f975e0528aecd78). The motivation for this change is to avoid that all generic signatures are retained in programs, when only a few are typically required for reflection.

This change adds a rule that causes the generic signature of `retrofit2.Call` to be retained.

Related resources:
- https://github.com/chrisbanes/tivi/commit/9720aa34f6c8b9e5260129ab6d05b44e84d3721c
- https://issuetracker.google.com/issues/188703877

Closes #3580

### DIFF
--- a/retrofit/src/main/resources/META-INF/proguard/retrofit2.pro
+++ b/retrofit/src/main/resources/META-INF/proguard/retrofit2.pro
@@ -5,6 +5,9 @@
 # Retrofit does reflection on method and parameter annotations.
 -keepattributes RuntimeVisibleAnnotations, RuntimeVisibleParameterAnnotations
 
+# Keep annotation default values (e.g., retrofit2.http.Field.encoded).
+-keepattributes AnnotationDefault
+
 # Retain service method parameters when optimizing.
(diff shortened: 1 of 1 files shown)


JakeWharton:
Thank you!

PR #3596, opened 2021-08-05
Add proguard config for R8 to keep type information for RxJava stuff

neworld:
The problem is similar to: https://issuetracker.google.com/issues/188703877
And the solution is similar to #3563

I tested only the case with Single of RxJava2. But other classes should work


JakeWharton:
Thanks!

ottne:
In addition to the above rules, I had to add

```-keep,allowobfuscation,allowshrinking class retrofit2.Response```

because a method like this one

```
interface WebApi {
@GET("/api/data")
    fun getData(): Single<Response<MyData>>
}
```

would throw this ` java.lang.IllegalArgumentException: Unable to create call adapter for io.reactivex.Single<retrofit2.Response>`.

Although I'm not sure if I'm simply missing an unreleased rule from Retrofit base. My rule is also possibly overly broad.

JakeWharton:
Yes, we also needed that rule in order to keep the generic signature of `Response`. It should be added to the library. Can you create a new issue? Or if you're up to it you could send a PR.

PR #3606, opened 2021-08-10
Bump maven-publish-plugin to 0.17.0

Goooler:


JakeWharton:
This is intentionally on an old version. The newer versions actually fail your build if you have the old env vars which is ridiculous behavior because we still use them on some projects. Therefore I refuse to upgrade until that check is removed.

Goooler:
> This is intentionally on an old version. The newer versions actually fail your build if you have the old env vars which is ridiculous behavior because we still use them on some projects. Therefore I refuse to upgrade until that check is removed.

Ok.

JakeWharton:
I filed a bug to remove the check. Hopefully we can upgrade to a hypothetical 0.18.0 as a result.

PR #3607, opened 2021-08-11
Update dependencies

Goooler:
I think it's time to bump bundled okHttp to latest stable version.


JakeWharton:
The OkHttp version is the last written in Java. We do not want to impose the Kotlin standard library dependency on our consumers.

Goooler:
> The OkHttp version is the last written in Java. We do not want to impose the Kotlin standard library dependency on our consumers.

Reverted.

PR #3610, opened 2021-08-12
Bump rxJava2 to 2.2.21

Goooler:
Merged from #2771, can close that after merging this one.


retrofit-adapters/rxjava2/src/test/java/retrofit2/adapter/rxjava2/FlowableTest.java
```diff
@@ -136,14 +136,14 @@ public void responseFailure() {
   public void responseRespectsBackpressure() {
```
JakeWharton, line 136: These tests can all be removed, and I suspect there's code in the implementation of the adapter that can also be removed if an initial 0 request is no longer valid.

JakeWharton: I meant the individual test cases. Not the whole file.

But like I said there's larger changes needed anyway.

PR #3613, opened 2021-08-16
Add keep rules for protobuf and Wire

### LINKED ISSUE #3564: Add R8 rules for converter-protobuf
What kind of issue is this?
 - [ ] Question. 
 - [ ] Bug report.
 - [x] Feature Request.
---
[`ProtoConverterFactory`](https://github.com/square/retrofit/blob/master/retrofit-converters/protobuf/src/main/java/retrofit2/converter/protobuf/ProtoConverterFactory.java) works by accessing the generated `parser()` method through reflection. If R8 is not correctly configured it will potentially remove or rename the method while doing code code shrinking or obfuscation. These days AARs and JARs can ship with the required R8 configuration which is then applied automatically. I believe it should be something like:

```
-keepclassmembers public class * extends com.google.protobuf.MessageLite {
  public static ** parser();
 }
```
The downside with this simple rule is that it would keep the `parser()` method of all kept `MessageLite` classes, even those that are not used by Retrofit. I believe that is not a significant concern for most apps. Let me know if this sounds reasonable.

JakeWharton:
Closes #3564.

### DIFF
--- a/retrofit-converters/protobuf/src/main/resources/META-INF/proguard/retrofit2-proguard-converter.pro
+++ b/retrofit-converters/protobuf/src/main/resources/META-INF/proguard/retrofit2-proguard-converter.pro
@@ -1,0 +1,7 @@
+# Parser looked up reflectively.
+-keepclassmembers class * implements com.google.protobuf.MessageLite {
+  public static com.google.protobuf.Parser parser();
+
+  # Fallback for v2.x.
+  public static com.google.protobuf.Parser PARSER;
+}
--- a/retrofit-converters/wire/src/main/resources/META-INF/proguard/retrofit2-wire-converter.pro
+++ b/retrofit-converters/wire/src/main/resources/META-INF/proguard/retrofit2-wire-converter.pro
@@ -1,0 +1,4 @@
+# ADAPTER field is looked up reflectively.
+-keepclassmembers class * extends com.squareup.wire.Message {
+  public static com.squareup.wire.ProtoAdapter ADAPTER;
+}


retrofit-converters/wire/src/main/resources/META-INF/proguard/retrofit2-wire-converter.pro
```diff
@@ -0,0 +1,4 @@
+# ADAPTER field is looked up reflectively.
+-keepclassmembers class * extends com.squareup.wire.Message {
```
mattprecious, line 2: I assume this doesn't prevent stripping the whole class if it's unused?

mattprecious (approved):

retrofit-converters/wire/src/main/resources/META-INF/proguard/retrofit2-wire-converter.pro
JakeWharton: Correct

retrofit-converters/protobuf/src/main/resources/META-INF/proguard/retrofit2-proguard-converter.pro
```diff
@@ -0,0 +1,7 @@
+# Parser looked up reflectively.
```
ChristianBecker, line 1: Should this file be named `retrofit2-protobuf-converter.pro`?

JakeWharton: Yep want to send a PR?

ChristianBecker: #3623

PR #3617, opened 2021-08-18
Update ErrorHandlingAdapter.java

shahwaiz90:
set ranges in the when statement, that way more optimised to catch error.


JakeWharton:
The code is in Java

PR #3636, opened 2021-09-21
Remove backpressure support from RxJava 2 and 3 adapters

JakeWharton:
The RS spec states that you must call request and that its value is strictly positive meaning that you can never actually request 0. As such, switch our implementation to reduce its overhead by removing a backpressure strategy.

Update the RxJava 2 and 3 deps to their latest. The latest RxJava 2 dep correctly validates this value as positive mandating this change, but the RxJava 3 one does not. Additionally, the RxJava 2 bump changes how undeliverable errors are propagated to plugins affecting our tests.

Closes #3610


mattprecious (approved):

Goooler:
Can close #2771 also.

PR #3648, opened 2021-10-28
Add support for Unit response type in HEAD method requests.

forrestpangborn:
This adds support for `Unit` (additionally to preexisting `Void` support) response types for HEAD method requests.


forrestpangborn:
@JakeWharton is there anything else you'd recommend or prefer for this? 

I'm willing to change the approach it uses for determining if `Unit` is available at runtime. I can also add more tests if you think thats appropriate and if you can suggest what would also be good to verify.

PR #3661, opened 2021-11-26
Allow headers with non-ascii values

### LINKED ISSUE #3662: Headers don't allow non-ASCII values
What kind of issue is this?

Problem:
Retrofit headers annotations don't have the option to allow non-ASCII values, even though Okhttp header class allows them. 

Scenario:
In our app, we were trying to send the content-related data in headers for the purpose of error analytics in our API. And having multi-lingual support while sending the data in headers, the API call was failing due to lack of support in retrofit to allow non-ASCII values

vipulasri:
Okhttp allows sending non-ASCII characters in headers but retrofit doesn't have the provision to allow them via annotations.
This PR request adds the possibility to allow non-ASCII characters via all Header annotations.
Affected Annotations:
1. Header
2. Headers
3. HeaderMap

Closes #3662

### DIFF
--- a/retrofit/src/main/java/retrofit2/RequestFactory.java
+++ b/retrofit/src/main/java/retrofit2/RequestFactory.java
@@ -241,11 +241,12 @@ final class RequestFactory {
         HTTP http = (HTTP) annotation;
         parseHttpMethodAndPath(http.method(), http.path(), http.hasBody());
       } else if (annotation instanceof retrofit2.http.Headers) {
-        String[] headersToParse = ((retrofit2.http.Headers) annotation).value();
+        retrofit2.http.Headers headers = (retrofit2.http.Headers) annotation;
+        String[] headersToParse = headers.value();
         if (headersToParse.length == 0) {
           throw methodError(method, "@Headers annotation is empty.");
         }
-        headers = parseHeaders(headersToParse);
+        this.headers = parseHeaders(headersToParse, headers.allowUnsafeNonAsciiValues());
       } else if (annotation instanceof Multipart) {
         if (isFormEncoded) {
           throw methodError(method, "Only one encoding annotation is allowed.");
@@ -293,7 +294,7 @@ final class RequestFactory {
       this.relativeUrlParamNames = parsePathParameters(value);
     }
 
-    private Headers parseHeaders(String[] headers) {
+    private Headers parseHeaders(String[] headers, boolean allowUnsafeNonAsciiValues) {
       Headers.Builder builder = new Headers.Builder();
       for (String header : headers) {
         int colon = header.indexOf(':');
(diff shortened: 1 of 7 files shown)


Goooler (approved):

retrofit/src/main/java/retrofit2/RequestFactory.java
```diff
@@ -524,15 +528,20 @@ private ParameterHandler<?> parseParameterAnnotation(
           ParameterizedType parameterizedType = (ParameterizedType) type;
           Type iterableType = Utils.getParameterUpperBound(0, parameterizedType);
           Converter<?, String> converter = retrofit.stringConverter(iterableType, annotations);
-          return new ParameterHandler.Header<>(name, converter).iterable();
+          return new ParameterHandler.Header<>(
+                  name, converter, ((Header) annotation).allowUnsafeNonAsciiValues())
```
JakeWharton, line 532:
```suggestion
                  name, converter, header.allowUnsafeNonAsciiValues())
```

retrofit/src/main/java/retrofit2/RequestFactory.java
```diff
@@ -524,15 +528,20 @@ private ParameterHandler<?> parseParameterAnnotation(
...
+              .iterable();
         } else if (rawParameterType.isArray()) {
           Class<?> arrayComponentType = boxIfPrimitive(rawParameterType.getComponentType());
           Converter<?, String> converter =
               retrofit.stringConverter(arrayComponentType, annotations);
-          return new ParameterHandler.Header<>(name, converter).array();
+          return new ParameterHandler.Header<>(
+                  name, converter, ((Header) annotation).allowUnsafeNonAsciiValues())
```
JakeWharton, line 539:
```suggestion
                  name, converter, header.allowUnsafeNonAsciiValues())
```

retrofit/src/main/java/retrofit2/RequestFactory.java
```diff
@@ -524,15 +528,20 @@ private ParameterHandler<?> parseParameterAnnotation(
...
+              .array();
         } else {
           Converter<?, String> converter = retrofit.stringConverter(type, annotations);
-          return new ParameterHandler.Header<>(name, converter);
+          return new ParameterHandler.Header<>(
+              name, converter, ((Header) annotation).allowUnsafeNonAsciiValues());
```
JakeWharton, line 544:
```suggestion
              name, converter, header.allowUnsafeNonAsciiValues());
```

retrofit/src/main/java/retrofit2/RequestFactory.java
```diff
@@ -245,7 +245,9 @@ private void parseMethodAnnotation(Annotation annotation) {
         if (headersToParse.length == 0) {
           throw methodError(method, "@Headers annotation is empty.");
         }
-        headers = parseHeaders(headersToParse);
+        headers =
+            parseHeaders(
+                headersToParse, ((retrofit2.http.Headers) annotation).allowUnsafeNonAsciiValues());
```
JakeWharton, line 250: Let's extract the cast value to a local rather than casting twice.

JakeWharton (approved):

Aishwar-wynk:
@JakeWharton Please let me know version contain this fix

hugonardo:
There is any plan to release this feature?

JakeWharton:
It is merged so it goes out with the next release, whenever that happens. For now you can use a snapshot build or deploy a local version in the mean time.

PR #3663, opened 2021-12-01
Exposes scalar converter internals for extension

Dragas:
This pull request exposes scalar converter internals so that they could be extended. Common scenario is to force preference of scalar converter using marker annotations when there are multiple converters available in the retrofit instance. For example

```java
import java.lang.annotation.Documented;
import java.lang.annotation.Retention;
import java.lang.annotation.Target;

import static java.lang.annotation.ElementType.METHOD;
import static java.lang.annotation.RetentionPolicy.RUNTIME;

@Documented
@Target(METHOD)
@Retention(RUNTIME)
public @interface ScalarResponse {
}

public interface RetrofitInterface {
    @GET("")
    @ScalarResponse
    Call<String> getText();
}

public class ConditionalScalarConverterFactory extends retrofit2.converter.scalars.ScalarsConverterFactory {
    public static ConditionalScalarResponseFactory create() {
        return new ConditionalScalarResponseFactory();
    }
    public Converter<?, RequestBody> requestBodyConverter(
            Type type,
            Annotation[] parameterAnnotations,
            Annotation[] methodAnnotations,
            Retrofit retrofit) {
        return null;
    }
     @Override
     public Converter<ResponseBody, ?> responseBodyConverter(
        for (Annotation methodAnnotation : annotations) {
            if (methodAnnotation.annotationType().equals(ScalarResponse.class)) {
                  return super.responseBodyConverter(type, annotations, retrofit);
            }
        }
    }
}

// somewhere in code which builds retrofit

Retrofit retrofitInterface = new Retrofit.Builder()
        .baseUrl(baseurl)
        .validateEagerly(true)
        .addConverterFactory(ConditionalScalarResponseFactory.create())
        .addConverterFactory(someOtherFactory) // another factory that overlaps with scalars factory
        .build();

```

Without this change I would need to copy over entire scalars converter module into my codebase and later maintain it in case original changes.

### DIFF
--- a/retrofit-converters/scalars/src/main/java/retrofit2/converter/scalars/ScalarRequestBodyConverter.java
+++ b/retrofit-converters/scalars/src/main/java/retrofit2/converter/scalars/ScalarRequestBodyConverter.java
@@ -20,9 +20,9 @@ import okhttp3.MediaType;
 import okhttp3.RequestBody;
 import retrofit2.Converter;
 
-final class ScalarRequestBodyConverter<T> implements Converter<T, RequestBody> {
+public final class ScalarRequestBodyConverter<T> implements Converter<T, RequestBody> {
   static final ScalarRequestBodyConverter<Object> INSTANCE = new ScalarRequestBodyConverter<>();
-  private static final MediaType MEDIA_TYPE = MediaType.get("text/plain; charset=UTF-8");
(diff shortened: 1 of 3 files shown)


JakeWharton:
See my reply on the issue as to why we don't actually need to make this change to support this functionality.

PR #3671, opened 2021-12-17
Put deps into version catalog

Goooler:
After this, we can convert kts easily.

PR #3672, opened 2021-12-19
Adopt vanniktech maven publish 0.18.0

Goooler:
Follow up commit for #3671, as `uploadArchives` task was removed from `vanniktech maven publish 0.18.0` to lead to publishing failure.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -97,7 +97,7 @@ jobs:
           java-version: 14
 
       - name: Upload Artifacts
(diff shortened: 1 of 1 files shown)


JakeWharton:
I didn't even notice this version had changed in the build refactor. Was anything else changed? I would strongly prefer separation of version bumps from any build refactoring in the future.

Goooler:
Yep, just missed that one, as the `libs.versions.toml` was copied from KotlinPoet's, forgot to revert that version back.

PR #3675, opened 2021-12-23
Update sonatype urls to s01

Goooler:
Left overs from #3673


Goooler:
Forgot to modify docs, sorry.

PR #3688, opened 2022-02-03
Prevent Kotlin callers from creating services with a nullable class type

bryansills:
[Everything is non-null](https://github.com/square/retrofit/blob/master/retrofit/src/main/java/retrofit2/internal/EverythingIsNonNull.java) so Kotlin users shouldn't be able to pass in a nullable class type into the Kotlin `create()` extension function.

I have signed the CLA. Hopefully I used all the correct verbiage here in my commit message and PR description.


JakeWharton:
Thanks!

Papmalik999:
C mon meilleure ka ʻōlelo

PR #3693, opened 2022-02-11
Run suspending calls on `Dispatchers.Default`

jingibus:
Run suspending calls within an optional `CoroutineDispatcher`, or `Dispatchers.Default` if none is provided.

Our first call to any Retrofit method is blocked by the initialization of our `OkHttpClient` (a prerequisite for the creation of a `Call`). On `suspend` APIs in Android, this is a problem: up until the invocation of `call.enqueue`, the caller will be running synchronously on the UI thread, which means the UI thread is blocked waiting on `OkHttpClient` initialization. This has been causing ANRs for us, even after attempting to work around by triggering initialization earlier in the app's lifecycle.

This fixes the issue by running all calls on a `CoroutineDispatcher`, either one provided in `Retrofit.Builder` or, failing that, `Dispatchers.Default`.

Since this changes forces all exceptions off of the calling stack frame and onto a `CoroutineDispatcher`, the hacks to get around limitations in exception throwing from Java Proxy objects can be safely removed.


Egorand (approved):
@JakeWharton any thoughts on this?

jingibus:
Modified to remove the public API to set the `CoroutineDispatcher`. This keeps Kotlin out of the ABI. Instead, always use `Dispatchers.Default`.

fsgjlp:
Looking forward to merging this commit.

retrofit/src/main/java/retrofit2/KotlinExtensions.kt
```diff
@@ -30,57 +30,63 @@ import kotlin.coroutines.resumeWithException
 inline fun <reified T: Any> Retrofit.create(): T = create(T::class.java)
 
 suspend fun <T : Any> Call<T>.await(): T {
```
Goooler, line 32: Use these would be less git changes, and needs to rebase.

```kt
suspend fun <T : Any> Call<T>.await(): T = withContext(Dispatchers.Default) {
```

```kt
suspend fun <T : Any> Call<T?>.await(): T? = withContext(Dispatchers.Default) {
```

```kt
suspend fun <T> Call<T>.awaitResponse(): Response<T> = withContext(Dispatchers.Default) {
```

juliocbcotta:
How about to instead of hardcoding a dispatcher, allow devs to provide a dispatcher..

PR #3702, opened 2022-02-25
Improve support for Unit as the return type

ansman:
Specifically this change fixes two cases:
- Using Unit for with suspending functions for responses without a body.
- Using Unit for HEAD requests.

This fixes #2888 but in a more compatible way. Closes #3648. Closes #3470.

### DIFF
--- a/retrofit/src/main/java/retrofit2/KotlinExtensions.kt
+++ b/retrofit/src/main/java/retrofit2/KotlinExtensions.kt
@@ -20,6 +20,7 @@ package retrofit2
 
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.suspendCancellableCoroutine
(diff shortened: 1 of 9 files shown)


retrofit/src/main/java/retrofit2/KotlinExtensions.kt
```diff
@@ -83,6 +84,12 @@ suspend fun <T : Any> Call<T?>.await(): T? {
   }
 }
 
+@JvmName("awaitUnit")
+suspend fun Call<Unit>.await() {
```
ansman, line 88: If you prefer this could be made `internal` but the type resolution seems to work fine so it seems like a nice addition.

JakeWharton:
This looks good. I'll review in depth next week. Retrofit is high up on my backlog of things to work on.

PR #3713, opened 2022-03-28
Run tests on each major JDK version

JakeWharton:
We only need to build on a single JDK (the latest, 18) and target the lowest supported (8).

### DIFF
--- a/retrofit/build.gradle
+++ b/retrofit/build.gradle
@@ -26,3 +26,26 @@ jar {
     attributes  'Automatic-Module-Name': 'retrofit2'
   }
 }
+
+// Create a test task for each supported JDK.
+(8..17).each { majorVersion ->
+  // Adoptium JDK 9 cannot extract on Linux or Mac OS.
+  if (majorVersion == 9) return
+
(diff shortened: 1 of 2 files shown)


retrofit/build.gradle
```diff
@@ -26,3 +26,26 @@ jar {
     attributes  'Automatic-Module-Name': 'retrofit2'
   }
 }
+
+// Create a test task for each supported JDK.
+(8..17).each { majorVersion ->
```
JakeWharton, line 31: 18 hasn't been deployed by Adoptium yet, but the 18 on the CI runner will be used by the normal 'test' task.

retrofit/build.gradle
```diff
@@ -26,3 +26,26 @@ jar {
...
+  // Adoptium JDK 9 cannot extract on Linux or Mac OS.
+  if (majorVersion == 9) return
```
JakeWharton, line 33: I _think_ this is a Gradle bug, but I'm not sure. I don't really care since it's only one really old JDK.

oldergod (approved):
:niiice:

PR #3715, opened 2022-03-30
Migrate Kotlin tests to separate project

### LINKED ISSUE #3714: Run non-Kotlin tests in JVM without Kotlin stdlib/coroutines
This is already needed for our limited Kotlin support, but as it grows having all codepaths covered in a VM where the types are not present will be important.

Refs #3702

JakeWharton:
This is easier than messing with classpaths.

Closes #3714


retrofit/gradle.properties
```diff
@@ -1,3 +1,5 @@
 POM_ARTIFACT_ID=retrofit
 POM_NAME=Retrofit
 POM_DESCRIPTION=A type-safe HTTP client for Android and Java.
+
+kotlin.stdlib.default.dependency=false
```
JakeWharton, line 5: Our version of Kotlin is old enough that the stdlib isn't added by default. This is a hedge against future upgrades.

Goooler (approved):

mattprecious (approved):

PR #3716, opened 2022-03-30
Add documentation for using suspend functions on a retrofit interface

JesseCorbett:
I found myself frequently revisiting the source to double check suspend function behavior, so figured it might warrant formal documentation (plus it doesn't seem to be mentioned anywhere at all that retrofit supports suspend functions at all!).

I put together a succinct addition to the website under the API Declaration section about using suspend functions and their general happy path and non-2XX error behavior


Narcos80:
46c94e916d10cfd212bd3708b3b9b0370652c9c9

Narcos80:
I will check it out 😉

PR #3717, opened 2022-03-30
Optimize CI a bit

Goooler:
Link https://github.com/square/okhttp/pull/7138.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,6 +17,8 @@ jobs:
           distribution: 'zulu'
           java-version: 18
 
+      - uses: gradle/gradle-build-action@v2
+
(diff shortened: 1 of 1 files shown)


JakeWharton:
I have pushed the first two commits to `master`. Thanks!

The gradle wrapper validation is not duplicated. Those three shards all run in parallel from the start of the build and as such they all need to validate the wrapper. Otherwise a compromised jar would be allowed to run on two of the three shards.

I prefer the existing formatting with the API levels one per-line. This allows changes to be better reflected in the git history.

I don't think caching the AVD adds any speed. It seems to actually take longer than what we currently do. The build is fast enough that I would prefer not having to maintain all that setup for meager gains.

PR #3723, opened 2022-04-19
Add execute version of Call.await

pablisco:
We found a situation where we wanted to execute a call synchronously and still make use of coroutines. 

Long story short: we were doing async calls inside our Authenticator. This can cause a lot of problems if the Dispatcher is full as it's rather easy to get into a dead lock situation. So, the solution we came up with was to make the call synchronously from our repository/datasource.

Being suspended it was very easy to integrate with existing coroutine setups. 

The only thing I am not sure is if it's worth wrapping it all in a `withContext(Dispatchers.IO)` to make sure users we don't use it on Main. For out case we are already somewhere far from Main, but may make it easier to use.

### DIFF
--- a/retrofit/src/main/java/retrofit2/KotlinExtensions.kt
+++ b/retrofit/src/main/java/retrofit2/KotlinExtensions.kt
@@ -20,12 +20,12 @@ package retrofit2
 
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.suspendCancellableCoroutine
-import java.lang.reflect.ParameterizedType
 import kotlin.coroutines.intrinsics.COROUTINE_SUSPENDED
 import kotlin.coroutines.intrinsics.intercepted
 import kotlin.coroutines.intrinsics.suspendCoroutineUninterceptedOrReturn
 import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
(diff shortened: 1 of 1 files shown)


JakeWharton:
Right now this function is equivalent of just calling `Call.execute()` directly so I'm not sure the purpose of wrapping it in a suspending function which never suspends and blocks the caller for a very long time.

When the code inside of `suspendCancellableCoroutine` invokes the `continuation` synchronously the value or exception is intercepted and is returned or thrown from the function synchronously. This is because coroutines are not allowed to invoke their continuation in the same stack frame and must instead return or throw when the result is known synchronously.

Mandojr1974:
This Armando Navarro Jr the true owner of this email holder of this phone
device so I say to you all that are giving there assistance to my help with
my issue I truly appreciate all your hard work everything else that you're
doing to fix and correct the problems that my device is going though I can
tell you for a fact that my phone device is experiencing problems like
being hacked and coded and it looks like all of my phone device and
personal info is compromised that is the whole entire truth that I'm
telling you all that are helping me and the person or individual that has
done this is very close I even believe that this person or this individual
is local I have done my inquiries and research on this issue I know that
this individual is posing as myself Armando Navarro Jr and receiving all of
my messages and information about all my apps so I leave you with this very
important information so you my execute version of Call.await (PR#3723)

PR #3744, opened 2022-06-01
Add service class & instance to Invocation

### LINKED ISSUE #3730: Add Service instance to `Invocation`
What kind of issue is this?

Since Retrofit supports service interface hierarchies, the `method()` & `arguments()` of the `Invocation` are no more sufficient to distinguish the invocation. A method might be declared in a super class with Retrofit annotations but the subclass is the interface which gets created.

```kt
interface Super {
  @GET("/super") fun getSuper()
}

// This interface gets instantiated via Retrofit.create(Sub::class)
interface Sub: Super {
  fun getSuperFallback(error: Throwable) { ... }
}
```

The primary use case right now is to be able to support/implement Resilience4j's CircuitBreaker fallback mechanism but there are [other library use cases](https://github.com/LianjiaTech/retrofit-spring-boot-starter/issues/109) and it might open the door to many others.

efemoney:
Fixes #3730

### DIFF
--- a/retrofit/java-test/src/test/java/retrofit2/InvocationTest.java
+++ b/retrofit/java-test/src/test/java/retrofit2/InvocationTest.java
@@ -39,6 +39,8 @@ public final class InvocationTest {
         @Path("p1") String p1, @Query("p2") String p2, @Body RequestBody body);
   }
 
+  interface ExampleSub extends Example {}
+
   @Test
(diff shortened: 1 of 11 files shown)


efemoney:
Seems CI wasn't triggered on the last push 🤔 Is there something I can do?

PR #3759, opened 2022-08-03
@HeaderMap parameter type must be Map or Headers

### LINKED ISSUE #3758: "@HeaderMap parameter type must be Map" is not precise enough, since HeaderMap supports Headers too !
Bug report.

"@HeaderMap parameter type must be Map" is not precise enough, since HeaderMap supports Headers too !

```java
        } else if (annotation instanceof HeaderMap) {
        if (type == Headers.class) {
          return new ParameterHandler.Headers(method, p);
        }

        validateResolvableType(p, type);
        Class<?> rawParameterType = Utils.getRawType(type);
        if (!Map.class.isAssignableFrom(rawParameterType)) {
          throw parameterError(method, p, "@HeaderMap parameter type must be Map.");
        }
```

Goooler:
Fixes #3758.

### DIFF
--- a/retrofit/src/test/java/retrofit2/RequestFactoryTest.java
+++ b/retrofit/src/test/java/retrofit2/RequestFactoryTest.java
@@ -609,10 +609,10 @@ public final class RequestFactoryTest {
   }
 
   @Test
-  public void headerMapMustBeAMap() {
+  public void headerMapMustBeAMapOrHeaders() {
     class Example {
       @GET("/")
-      Call<ResponseBody> method(@HeaderMap List<String> headers) {
+      Call<ResponseBody> method(@HeaderMap okhttp3.Headers headers, @HeaderMap List<String> headerMap) {
         return null;
       }
     }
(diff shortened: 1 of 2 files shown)


retrofit/src/test/java/retrofit2/RequestFactoryTest.java
```diff
@@ -609,10 +609,10 @@ Call<ResponseBody> method(@HeaderMap Map<String, Object> headers) {
   }
 
   @Test
-  public void headerMapMustBeAMap() {
+  public void headerMapMustBeAMapOrHeaders() {
     class Example {
       @GET("/")
-      Call<ResponseBody> method(@HeaderMap List<String> headers) {
+      Call<ResponseBody> method(@HeaderMap okhttp3.Headers headers, @HeaderMap List<String> headerMap) {
```
JakeWharton, line 615: Why was the `Headers` parameter added? If we want to test it works then it should be done separately from testing a type which does not work.

Goooler: Cause I see the workable ones already exist.

https://github.com/square/retrofit/blob/296fcd962956c1c806b3bcf08b91edb96a8ada75/retrofit/src/test/java/retrofit2/RequestFactoryTest.java#L715-L737

https://github.com/square/retrofit/blob/296fcd962956c1c806b3bcf08b91edb96a8ada75/retrofit/src/test/java/retrofit2/RequestFactoryTest.java#L739-L763

JakeWharton:
Thanks!

PR #3766, opened 2022-08-28
Run tests on JDK 19

Goooler:
Follow up #3713.

https://docs.gradle.org/7.6/release-notes.html


retrofit/build.gradle
```diff
@@ -26,7 +26,7 @@ jar {
 }
 
 // Create a test task for each supported JDK.
-(8..17).each { majorVersion ->
+(8..JavaVersion.current().majorVersion.toInteger()).each { majorVersion ->
```
JakeWharton, line 29: Doesn't this mean that if I'm running JDK 11 to build the project that I only get tests for 8-11?

Goooler: Yep

Goooler: The default test task running on the 18, and you added 8 ~ 17, so all tests running on 8~18? 
If so I will update this PR after JDK 19 release and change here to 18.

Goooler: Would like to wait for JDK 19 & Gradle 7.6 release, seems the Gradle team will support using Zulu in `jvmToolchain`.

JakeWharton: Can we set the toolchain explicitly to 18 rather than letting it be implicit? This would ensure that anyone building the project always builds on 18 and gets test tasks from 8-17.

Goooler: 5d58a10f5a73a0e6f5149138e575668952471a5b

PR #3782, opened 2022-11-10
Add proguard rules to keep inherited services

### LINKED ISSUE #3777: Update proguard rules for inherited services
What kind of issue is this?

 - [X] Feature Request. Start by telling us what problem you’re trying to solve. Often a solution

An issue on retrofit was reported to R8 with a small example showing R8 to strip out a subclass of a defined service interface:
https://b.corp.google.com/issues/248203227

The solution for inheritance may be to adjust the existing keep rule:
```
# With R8 full mode, it sees no subtypes of Retrofit interfaces since they are created with a Proxy
# and replaces all potential values with null. Explicitly keeping the interfaces prevents this.
-if interface * { @retrofit2.http.* <methods>; }
-keep,allowobfuscation interface <1>
```
to be:
```
# With R8 full mode, it sees no subtypes of Retrofit interfaces since they are created with a Proxy
# and replaces all potential values with null. Explicitly keeping the interfaces prevents this.
-if interface * { @retrofit2.http.* <methods>; }
-keep,allowobfuscation interface <1>

-if interface * { @retrofit2.http.* <methods>; }
-keep,allowobfuscation interface * extends <1>
```
I've created this a feature request and not a pull request because I am not certain what specification the retrofit libary places on inheritance of services.

Goooler:
Closes #3777.

### DIFF
--- a/retrofit/src/main/resources/META-INF/proguard/retrofit2.pro
+++ b/retrofit/src/main/resources/META-INF/proguard/retrofit2.pro
@@ -31,7 +31,7 @@
 -if interface * { @retrofit2.http.* <methods>; }
 -keep,allowobfuscation interface <1>
 
-# Keep inherited services
+# Keep inherited services.
 -if interface * { @retrofit2.http.* <methods>; }
 -keep,allowobfuscation interface * extends <1>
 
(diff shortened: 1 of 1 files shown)


retrofit/src/main/resources/META-INF/proguard/retrofit2.pro
```diff
@@ -31,6 +31,10 @@
 -if interface * { @retrofit2.http.* <methods>; }
 -keep,allowobfuscation interface <1>
 
+# Keep inherited services
```
JakeWharton, line 34:
```suggestion
# Keep inherited services.
```
https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

JakeWharton (approved):

Issue #2180, opened 2017-01-30
Retrofit 3.x Things


JakeWharton:
Currently no plans for a Retrofit 3.x. Just tracking changes in a single issue instead of many.

 * Mock behavior should invoke the implementation method for each `Call.execute()` / `Call.enqueue()`. Currently it only invokes it once and reuses the result for every cloned `Call`.

TobiasReich:
Will there be a simple way in order to retry a call n times?
This would come very handy when there is a call that should be retried after an error.
I was hoping for something like: request.retry(5);

naturalwarren:
To do this with Observables we use a custom transformer that gets tacked onto request invocations that takes in the number of retries.

Jawnnypoo:
Curious if for a 3.x it would be appropriate to have `HttpException` be a part of the `retrofit` module, instead of creating it separately in each adapter that is created. Is there a reason for this currently?

JakeWharton:
There's no use for it in the main module at present which is why each
adapter defines its own. This change wouldn't need to wait for 3.x as it
can be added in a backwards compatible-way.

Jawnnypoo:
Gotcha. So do you think that each should continue to have its own copy? The benefits I could see with unifying mostly come from just reducing the duplication. Is this something you would be interested in seeing in a PR, or do you prefer keeping them separate?

JakeWharton:
I'd be okay with a PR. Basically copy paste it into the main module and
then change all the existing ones to extend. I'm not happy it has to be
non-final in the main module, but it's only an exception so not a big deal.

JakeWharton:
I created #2194 to track that. Feel free to send a PR over the weekend. Otherwise I'll grab it on Monday. I need to release really soon but I'll wait for this.

JakeWharton:
@TobiasReich no plans for retry. You can already call `clone()` on a `Call` to retry and so implementing that at the application level seems more appropriate. You can write a `CallAdapter` to return a custom type that provides a nice API for retry. Because of the behavior in sync vs. async mode and what executor the delay happens on being unclear I don't think I want it in the library.

soodakshay:
Hey @JakeWharton  I am a big fan of yours. Can I get some code sample so that I can learn a good coding style? I am an Android Developer and I want to improve my coding styles and learn about the quality code.
Hope you will help me
Thanks :)

ylfzq:
You can add code below to your okhttp interceptor to archieve retry mechanism.

```java
[...]

PR #3793, opened 2022-12-25
Add ability to set base url per service

agamula90:
I found this feature - https://github.com/square/retrofit/issues/2180#issuecomment-327833876 to be useful in android apps.
For example, when backend team decided to move and maybe refactor their code to other remote. And it's not one day change, so android app need to use two base urls - old one and new one. 
It can be done with prepending base url per each request for new service or use 2 retrofit instances. 
I'd not consider 2nd solution at all because it's not memory efficient + service creation pattern is error prone - you can accidentally create service from wrong retrofit instance.
1st solution works, but it requires lot of base url duplicates.

From other side this feature can be used in proxy servers, that use multiple remotes to fetch/operate.

PS.: I didn't completely understood from linked comment if you consider these use cases as valid, so please, accept my apologies if I missed something from this long v3 thread. 
But I don't see any pitfalls in changes made to support this feature, so I created this PR 😄

### DIFF
--- a/retrofit/kotlin-test/src/test/java/retrofit2/KotlinRequestFactoryTest.java
+++ b/retrofit/kotlin-test/src/test/java/retrofit2/KotlinRequestFactoryTest.java
@@ -3,6 +3,8 @@ package retrofit2;
 import kotlin.Unit;
 import okhttp3.Request;
 import org.junit.Test;
+
+import retrofit2.http.BaseUrl;
 import retrofit2.http.HEAD;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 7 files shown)


JakeWharton:
Using two instances is the correct way to do this.
```java
var builder = new Retrofit.Builder()
    .client(client)
    // adapters, converters, etc.
    ;

var one = builder
    .baseUrl("https://one/")
    .build()
    .create(Service.class);

var two = builder
    .baseUrl("https://two/")
    .build()
    .create(Service.class);
```
This shares the client, adapter, converters, etc. which is no more resource intensive than your solution but is more flexible since you can alter other properties.

We don't want to introduce a third mechanism for setting the base URL and having to need to define precedence rules for them.

Issue #3558, opened 2021-05-10
Add support for Kotlin's Result


GSala:
Now that inline classes are stable in Kotlin 1.5.0, it would be nice to have API endpoints return [Result](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/) to wrap the outcome (success/failure) of a retrofit call, instead of returning a value or throwing an exception.

For example:

```
@GET("me")
suspend fun getUser(): Result<User>
```

kobito-kaba:
I doubt if Retrofit should support kotlin Result at this moment, because returning kotlin Result from functions is not the default behavior.

Anyway, [this call adapter](https://gist.github.com/kobito-kaba/1c209f01a74d190beb771ba7567a4d91) will work for your use case.
note:
That call adapter will work for `suspend fun foo() : Result<Foo>` , but won't work for `fun foo(): Result<Foo>` in some reasons.

tejas-n:
Initially I thought that might be a good addition, but I guess you can always use [runCatching](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/run-catching.html) if you want the type wrapped in `Result`

```
val user: Result<User> = runCatching {
    api.getUser()
 }
```

youarenot667543:
> val user: Result\u003CUser> = runCatching {\n    api.getUser()\n }

agamula90:
You can try my solution - https://github.com/agamula90/RetrofitKx, hosted on mavenCentral, based on retrofit + moshi + ksp if you are are looking for better kotlin api

Goooler:
As a workaround, you can try out https://github.com/connyduck/networkresult-calladapter.

PR #3873, opened 2023-04-05
Add support for Kotlin's Result

tomridder:
add the ResultCallAdapter to return [Result](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/) to wrap the outcome (success/failure) of a retrofit call, instead of returning a value or throwing an exception. #3558 

like

@GET("me")
suspend fun getUser(): Result<User>

all the testcase in Kotlin-test package and  retrofit-adapter moudle's test case has been passed

### DIFF
--- a/retrofit/kotlin-test/src/test/java/retrofit2/KotlinSuspendTest.kt
+++ b/retrofit/kotlin-test/src/test/java/retrofit2/KotlinSuspendTest.kt
@@ -26,8 +26,6 @@ import okhttp3.mockwebserver.MockWebServer
 import okhttp3.mockwebserver.SocketPolicy.DISCONNECT_AFTER_REQUEST
 import okhttp3.mockwebserver.SocketPolicy.NO_RESPONSE
 import org.assertj.core.api.Assertions.assertThat
-import org.junit.Assert.assertTrue
-import org.junit.Assert.fail
 import org.junit.Ignore
 import org.junit.Rule
 import org.junit.Test
@@ -39,6 +37,8 @@ import java.io.IOException
 import java.lang.reflect.ParameterizedType
 import java.lang.reflect.Type
 import kotlin.coroutines.CoroutineContext
+import org.junit.Assert.*
+import retrofit2.converter.gson.GsonConverterFactory
 
 class KotlinSuspendTest {
   @get:Rule val server = MockWebServer()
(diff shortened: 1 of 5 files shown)


tomridder:
@JakeWharton @codebutler hello  ?anyone there

liu-wanshun (approved):

retrofit/kotlin-test/src/test/java/retrofit2/KotlinSuspendTest.kt
```diff
@@ -39,6 +37,8 @@ import java.io.IOException
 import java.lang.reflect.ParameterizedType
 import java.lang.reflect.Type
 import kotlin.coroutines.CoroutineContext
+import org.junit.Assert.*
```
elliptic1, line 40: Don't use star imports if this project wasn't already using them

retrofit/src/main/java/retrofit2/HttpServiceMethod.java
```diff
@@ -23,6 +23,8 @@
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import javax.annotation.Nullable;
+
+import kotlin.Result;
```
elliptic1, line 27: extra space and new import not alphabetized

retrofit/src/main/java/retrofit2/ResultCallAdapterFactory.kt
```diff
@@ -0,0 +1,88 @@
...
+        callback.onResponse(this@ResultCall, Response.success(Result.failure(t)))
+      }
+    })
+  }
+
+  override fun execute(): Response<Result<T>> {
+    return try {
+      val response = delegate.execute()
+      val result = if (response.isSuccessful) {
+        Result.success(response.body()!!)
```
elliptic1, line 64: Don't use !! - there is always a better way to fail gracefully into the Result.failure case if this is null.

retrofit/src/main/java/retrofit2/ResultCallAdapterFactory.kt
```diff
@@ -0,0 +1,88 @@
...
+      } else {
+        Result.failure(IOException("Unexpected error: ${response.errorBody()?.string()}"))
+      }
+      Response.success(result)
+    } catch (e: IOException) {
+      Response.success(Result.failure(e))
```
elliptic1, line 70: This does the same job as the more general Exception case, so it's not needed

PR #3886, opened 2023-05-02
Add keep rule for keeping generic signature of return types

mkj-gram:
The additional rule generalises keeping the generic signature of return type of service methods. The rule effectively fixes an issue with RxJava3 Observables as reported in b/280277628.


JakeWharton:
I suspect this means we can delete https://github.com/square/retrofit/blob/master/retrofit-adapters/rxjava3/src/main/resources/META-INF/proguard/retrofit2-rxjava3-adapter.pro and the others whose goal was to mitigate this. I'll do that and merge tomorrow.

JakeWharton:
Based on the linked bug, the rules which were already embedded in the adapters should have kept the signature.

Regardless, your rule is superior since it covers all potential return values instead of us having to remember to include it for every type.

PR #3893, opened 2023-05-11
synchronized on serviceClass to avoid lock competition  when loading services of different serviceClass

Knight-ZXW:
synchronized on serviceClass to avoid lock competition when loading methods of different serviceClass


retrofit/src/main/java/retrofit2/Retrofit.java
```diff
@@ -163,7 +163,7 @@ public <T> T create(final Class<T> service) {
                 Platform platform = Platform.get();
                 return platform.isDefaultMethod(method)
                     ? platform.invokeDefaultMethod(method, service, proxy, args)
-                    : loadServiceMethod(method).invoke(args);
+                    : loadServiceMethod(service,method).invoke(args);
```
JakeWharton, line 166:
```suggestion
                    : loadServiceMethod(service, method).invoke(args);
```

PR #3957, opened 2023-09-13
Close open return points of the response and response body

### LINKED ISSUE #3956: Response Body Not Closed When Exception Thrown
I've been investigating the log stated below in our services. We are using Retrofit to make calls between our microservices, sometimes our calls result with the specified log. I searched all of our codebase to find non closed response body and finally came across with the OkHttpCall#parseResponse method stated below. I think we should add some kind of close logic to catch block at line 245.

`
A connection to {my_url} was leaked. Did you forget to close a response body? To see where this was allocated, set the OkHttpClient logger level to FINE: Logger.getLogger(OkHttpClient.class.getName()).setLevel(Level.FINE);
`

https://github.com/square/retrofit/blob/d278bc8bc4541ab51b2a71297c7d8a72eb839e71/retrofit/src/main/java/retrofit2/OkHttpCall.java#L245C21-L245C21

haydin505:
We receive the following logs from our services. This pull request fixes the issue. I'm aware that `retrofit2.CallTest#responseBodyStreams` test fails but I think we should refactor it too.
```
java.lang.Throwable: response.body().close()
	at okhttp3.internal.platform.Platform.getStackTraceForCloseable(Platform.kt:145)
	at okhttp3.internal.connection.RealCall.callStart(RealCall.kt:170)
	at okhttp3.internal.connection.RealCall.execute(RealCall.kt:151)
	at retrofit2.OkHttpCall.execute(OkHttpCall.java:204)
	at retrofit2.adapter.rxjava3.CallExecuteObservable.subscribeActual(CallExecuteObservable.java:46)
	at io.reactivex.rxjava3.core.Observable.subscribe(Observable.java:13262)
	at retrofit2.adapter.rxjava3.BodyObservable.subscribeActual(BodyObservable.java:35)
	at io.reactivex.rxjava3.core.Observable.subscribe(Observable.java:13262)
	at io.reactivex.rxjava3.core.Observable.blockingFirst(Observable.java:5453)
```
Closes #3956

### DIFF
--- a/retrofit/src/main/java/retrofit2/OkHttpCall.java
+++ b/retrofit/src/main/java/retrofit2/OkHttpCall.java
@@ -234,8 +234,11 @@ final class OkHttpCall<T> implements Call<T> {
     }
 
     if (code == 204 || code == 205) {
-      rawBody.close();
-      return Response.success(null, rawResponse);
+      try {
+        return Response.success(null, rawResponse);
+      } finally {
+        rawBody.close();
+      }
     }
 
     ExceptionCatchingResponseBody catchingBody = new ExceptionCatchingResponseBody(rawBody);
@@ -245,6 +248,7 @@ final class OkHttpCall<T> implements Call<T> {
     } catch (RuntimeException e) {
       // If the underlying source threw an exception, propagate that rather than indicating it was
       // a runtime exception.
+      catchingBody.close();
       catchingBody.throwIfCaught();
       throw e;
     }


JakeWharton:
The test is correct. Please change your implementation to accommodate it.

haydin505:
> The test is correct. Please change your implementation to accommodate it.

I've changed the implementation. I suspect that RuntimeException is thrown somewhere and caught at interceptors, as a result, response returns properly to the user. That's why I closed the body in the catch block.

haydin505:
We recently added a network interceptor. After changing NetworkInterceptor to standard Interceptor, the log disappeared.

Issue #3972, opened 2023-09-26
Distinguishing between a network error and a conversion error


dbyron-sf:
What kind of issue is this?

 - [ ] Question. This issue tracker is not the place for questions. If you want to ask how to do
       something, or to understand why something isn't working the way you expect it to, use Stack
       Overflow. https://stackoverflow.com/questions/tagged/retrofit

 - [ ] Bug report. If you’ve found a bug, spend the time to write a failing test. Bugs with tests
       get fixed. Here’s an example: https://gist.github.com/swankjesse/6608b4713ad80988cdc9

 - [x] Feature Request. Start by telling us what problem you’re trying to solve. Often a solution
       already exists! Don’t send pull requests to implement new features without first getting our
       support. Sometimes we leave features out on purpose to keep the project small.

I can't quite tell if https://github.com/square/retrofit/issues/3749 already covers this, or perhaps it's a big enough change that it belongs in [the retrofit 3.x issue](https://github.com/square/retrofit/issues/2180).  What I'm looking for is a way to tell explicitly when there's a "conversion exception".  As in, when [responseConverter.convert](https://github.com/square/retrofit/blob/dbbfc5b186e0e59e6a20da8422390be41ab6c340/retrofit/src/main/java/retrofit2/OkHttpCall.java#L243) throws an exception.  [The interface](https://github.com/square/retrofit/blob/dbbfc5b186e0e59e6a20da8422390be41ab6c340/retrofit/src/main/java/retrofit2/Converter.java#L43) specifies an IOException, but because [Call.execute](https://github.com/square/retrofit/blob/dbbfc5b186e0e59e6a20da8422390be41ab6c340/retrofit/src/main/java/retrofit2/OkHttpCall.java#L204) calls both parseResponse and [okhttp's Call.execute](https://github.com/square/okhttp/blob/parent-4.9.3/okhttp/src/main/kotlin/okhttp3/Call.kt#L55) which also throws an IOException, I can't see a way for callers of Call.execute to be able to tell the difference.

The motivation is to be able to tell whether it's worth retrying a request or not.  I wouldn't expect conversion exception to be worth retrying -- they'd fail every time.  But other IOExceptions would be worth retrying as they might succeed.

Defining an exception class that extends from IOException, and then changing Converter.convert to throw that feels like it would solve this, but there could totally be better ways, or things I haven't thought of.  Happy to make a PR if this seems like a reasonable approach.

dbyron-sf:
[...]

PR #3974, opened 2023-09-29
feat(converters): change Converter.convert to throw ConversionException instead of IOException

dbyron-sf:
so callers of Call.execute can distinguish between retryable exceptions (IOException) and non-retryable exceptions (ConversionException).

closes https://github.com/square/retrofit/issues/3972.


JakeWharton:
Thanks for attempting a PR, but please see my comment on the issue about why this won't actuall work.

PR #3996, opened 2023-11-16
use target method as Lock object

Knight-ZXW:


JakeWharton:
So I was going to suggest this on the initial PR, but I wasn't sure that instance was guaranteed to be the same.

Knight-ZXW:
> So I was going to suggest this on the initial PR, but I wasn't sure that instance was guaranteed to be the same.

Your guess is correct!
I tested this case, and the result showing that they are not same object.   

I will close thie PR.

image
image

PR #3997, opened 2023-11-23
Update Dependency com.squareup.okhttp3

OmerSec:
Updated com.squareup.okhttp3 version to latest stable 4.12.0
Throw EOF exception in RequestBuilder to compile with latest okhttp version


JakeWharton:
OkHttp version 4.x and 5.x are written in Kotlin, and we're not ready to force that dependency onto our consumers yet. We will in Retrofit 3.x, though, sometime early next year.

PR #4001, opened 2023-11-26
Refactoring some code to remove design and implementation smells

Nikunj0601:
This pull request contains refactored code. The purpose of this refactoring is to improve understandability, and readability and remove some implementation and design smells.

Refactoring using rename variable/method:
I have renamed the method create() in GsonConvertorFactory.java to createGsonInstance() as this name justifies the method more.

Refactoring by decomposing conditional:
I have created a new boolean variable for if condition in HttpServiceMethod.java at line 51, for more readability.

Refactoring by extract method:
I have extracted few lines of logic from addQueryParam method of RequestBuilder.java and created new method for that extracted logic.

Refactoring by introducing explaining variable:
I have introduced 3 new boolean variables in equals() method of Utils.java class to increase the readability and understandibity of the code.

Refactoring by replacing conditional polymorphism:
I have replaced if else condition in getRawType method of Utils.java class by created new TypeHandler class and adding an appropriate sub class for implementation  of each branch of if else condition logic.

Refactoring by extracting class:
I have extracted the new class ErrorsUtils from Utils.java as this class is big and has many responsibility. I have moved methods methodError , parameterError methods which was for errors in Utils class into new class ErrorsUtils.

Refactoring by move method refactoring:
I have pulled down parseAnnotations of ServiceMethod class to its subclass HttpServiceMethod as it calls the method of HttpServiceMethod which also creates a cyclic hierarchy. Hence, by moving down the method cyclic hierarchy is also resolved.

### DIFF
--- a/retrofit-converters/gson/src/main/java/retrofit2/converter/gson/GsonConverterFactory.java
+++ b/retrofit-converters/gson/src/main/java/retrofit2/converter/gson/GsonConverterFactory.java
@@ -38,8 +38,8 @@ public final class GsonConverterFactory extends Converter.Factory {
    * Create an instance using a default {@link Gson} instance for conversion. Encoding to JSON and
    * decoding from JSON (when no charset is specified by a header) will use UTF-8.
    */
-  public static GsonConverterFactory create() {
-    return create(new Gson());
+  public static GsonConverterFactory createGsonInstance() {
+    return createGsonInstance(new Gson());
   }
 
   /**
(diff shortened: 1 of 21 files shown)


JakeWharton:
Hello. Thanks for attempting to improve our project, but this type of PR is not acceptable as-is. At the very least, you are renaming public API which is a breaking change and introducing needless types into the public API surface of the library.

This is not a good way to approach an open source project, especially one that is a decade old and well established. Changes proposed to any project should always be as small as possible, have clear motivations, and ideally are tested.

I'm not sure if this is part of an assignment, you are demonstrating a tool, or a combination of both. Using these code quality tools and applying their advice blindly is not a good way to have your changes integrated.

If you are following the instruction of a professor in the creation of this PR, please forward my feedback to them. They are doing you a great disservice. There is plenty of opportunity to contribute to open source as an education experience, but this is not the way.

I will integrate a3edb8f8f0c26acc1424ff336ce6a1aa87d35451 via #4004 as it seems like the only change of the bunch that is acceptable as-is, but in general we aren't looking for changes like this. But if you were to propose it in isolation, it's _possible_ that it would be accepted.

PR #4013, opened 2023-12-18
Reject suspend function with `Call` return type

### LINKED ISSUE #4012: Eagerly reject `Call` return type when using `suspend`
This is a common user error. We can trivially detect and suggest they remove the wrapper type.

Goooler:
Closes #4012.

### DIFF
--- a/retrofit/src/main/java/retrofit2/HttpServiceMethod.java
+++ b/retrofit/src/main/java/retrofit2/HttpServiceMethod.java
@@ -53,6 +53,12 @@ abstract class HttpServiceMethod<ResponseT, ReturnT> extends ServiceMethod<Retur
         responseType = Utils.getParameterUpperBound(0, (ParameterizedType) responseType);
         continuationWantsResponse = true;
       } else {
+        if (getRawType(responseType) == Call.class) {
+          throw methodError(
+              method,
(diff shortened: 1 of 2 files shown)


retrofit/src/main/java/retrofit2/HttpServiceMethod.java
```diff
@@ -53,6 +53,12 @@ static <ResponseT, ReturnT> HttpServiceMethod<ResponseT, ReturnT> parseAnnotatio
         responseType = Utils.getParameterUpperBound(0, (ParameterizedType) responseType);
         continuationWantsResponse = true;
       } else {
+        if (getRawType(responseType) == Call.class) {
+          throw methodError(
+              method,
+              "Unable to create call adapter for suspend function with Call return type.");
```
JakeWharton, line 59: I would like this message to be a bit more instructive.

How about

> Suspend functions should not return Call, as they already execute asynchronously.
>
> Change its return type to ${..}

Where the expression is the parameterized type of Call.

Goooler: Make sense!

JakeWharton:
Thanks!

PR #4017, opened 2023-12-19
Reject suspend function with Future return type

Goooler:
Follow up #4013.

### DIFF
--- a/retrofit/src/main/java/retrofit2/HttpServiceMethod.java
+++ b/retrofit/src/main/java/retrofit2/HttpServiceMethod.java
@@ -22,6 +22,7 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
+import java.util.concurrent.Future;
 import javax.annotation.Nullable;
 import kotlin.Unit;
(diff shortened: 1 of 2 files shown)


retrofit/src/main/java/retrofit2/HttpServiceMethod.java
```diff
@@ -53,11 +54,14 @@ static <ResponseT, ReturnT> HttpServiceMethod<ResponseT, ReturnT> parseAnnotatio
         responseType = Utils.getParameterUpperBound(0, (ParameterizedType) responseType);
         continuationWantsResponse = true;
       } else {
-        if (getRawType(responseType) == Call.class) {
+        Class<?> rawType = getRawType(responseType);
+        if (rawType == Call.class || Future.class.isAssignableFrom(rawType)) {
```
Goooler, line 58: If we should check RxJava types as well?

JakeWharton: We can't because the dependency isn't here. And we can't go by string name because of obfuscation possibility. I'm actually okay just doing Call and not doing Future or anything else. I don't think people mix and match other async types with suspend to the degree that they do Call (if at all).

PR #4021, opened 2023-12-29
Support encoding plus char in path

### LINKED ISSUE #3653: Why do not encode ‘+’  in path?
What kind of issue is this?

 - [x] Question. This issue tracker is not the place for questions. If you want to ask how to do

 - [ ] Feature Request. Start by telling us what problem you’re trying to solve. Often a solution

## retrofit-v2.9.0

the api docs
https://github.com/square/retrofit/blob/7158698314daa138e993fac6a590ed19d78a8599/retrofit/src/main/java/retrofit2/http/Path.java#L48

```java
 @GET("/user/{name}")
 Call<ResponseBody> encoded(@Path("name") String name);

 @GET("/user/{name}")
 Call<ResponseBody> notEncoded(@Path(value="name", encoded=true) String name);
 
Calling foo.encoded("John+Doe") yields /user/John%2BDoe
whereas foo.notEncoded("John+Doe") yields /user/John+Doe.
```

https://github.com/square/retrofit/blob/fbf1225e28e2094bec35f587b8933748b705d167/retrofit/src/main/java/retrofit2/RequestBuilder.java#L133

```java
  private static String canonicalizeForPath(String input, boolean alreadyEncoded) {
    int codePoint;
    for (int i = 0, limit = input.length(); i < limit; i += Character.charCount(codePoint)) {
      codePoint = input.codePointAt(i);
      if (codePoint < 0x20
          || codePoint >= 0x7f
          || PATH_SEGMENT_ALWAYS_ENCODE_SET.indexOf(codePoint) != -1
          || (!alreadyEncoded && (codePoint == '/' || codePoint == '%'))) {
        // Slow path: the character at i requires encoding!
        Buffer out = new Buffer();
        out.writeUtf8(input, 0, i);
        canonicalizeForPath(out, input, i, limit, alreadyEncoded);
        return out.readUtf8();
      }
    }

    // Fast path: no characters required encoding.
    return input;
  }
```
The canonicalizeForPath method do not encode '+'

Goooler:
Closes #3653.

### DIFF
--- a/retrofit/src/main/java/retrofit2/RequestBuilder.java
+++ b/retrofit/src/main/java/retrofit2/RequestBuilder.java
@@ -139,7 +139,7 @@ final class RequestBuilder {
       if (codePoint < 0x20
           || codePoint >= 0x7f
           || PATH_SEGMENT_ALWAYS_ENCODE_SET.indexOf(codePoint) != -1
-          || (!alreadyEncoded && (codePoint == '/' || codePoint == '%'))) {
+          || (!alreadyEncoded && (codePoint == '/' || codePoint == '%' || codePoint == '+'))) {
         // Slow path: the character at i requires encoding!
         Buffer out = new Buffer();
         out.writeUtf8(input, 0, i);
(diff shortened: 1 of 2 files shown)


JakeWharton:
https://github.com/square/retrofit/issues/3653#issuecomment-1911312533

Goooler:
If so, may have to tweak
https://github.com/square/retrofit/blob/ee71a6858dfddcc4650f6ff4d71112911bdf6ca7/retrofit/src/main/java/retrofit2/http/Path.java#L52-L53

JakeWharton:
Yeah the docs are probably from OkHttp 1.x days, or even before we used OkHttp.

Zhuinden:
This bug is causing quite a bit of headache now in 2025. 

Retrofit claims to be URL-encoding parameters, but it refuses to do it for `+` in a path parameter, even though according to URL-encode specification, it should be encoded. Unfortunate for me because it makes my situation a little tricky. 😅

PR #3582, opened 2021-06-29
BOM publishing

realdadfish:
Alternative to #3487

PR #4022, opened 2023-12-31
Support BOM publishing

### LINKED ISSUE #3231: Publish a BOM

Goooler:
- Closes #3231.
- Closes #3487.
- Closes #3582.

### DIFF
--- a/retrofit-bom/build.gradle
+++ b/retrofit-bom/build.gradle
@@ -1,0 +1,26 @@
+plugins {
+  id 'java-platform'
+  alias libs.plugins.mavenPublish
+}
+
+// TODO: This could be simplified after updating maven publish plugin, see https://vanniktech.github.io/gradle-maven-publish-plugin/what/#java-platform.
+publishing {
+  publications {
+    maven(MavenPublication) {
+      from components.javaPlatform
+    }
+  }
+}
+
+dependencies {
+  constraints {
+    rootProject.subprojects { subproject ->
+      subproject.plugins.withId(libs.plugins.mavenPublish.get().pluginId) {
+        // Exclude self project from BOM.
+        if (subproject != this.project) {
+          api subproject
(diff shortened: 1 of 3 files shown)


retrofit-bom/build.gradle
```diff
@@ -0,0 +1,26 @@
...
+    maven(MavenPublication) {
+      from components.javaPlatform
+    }
+  }
+}
+
+dependencies {
+  constraints {
+    rootProject.subprojects { subproject ->
+      subproject.plugins.withId(libs.plugins.mavenPublish.get().pluginId) {
```
JakeWharton, line 18: Very nice!

JakeWharton (approved):

mikeholler:
Thanks so much @Goooler!

PR #4055, opened 2024-01-25
Remove the use of toolchains

JakeWharton:
There no reason to use JDK 8 to run compilation when all JDKs can cross-compile to that target. Moreover, this actually prevents upgrading error-prone which wants a newer javac than 8.

This is mostly a revert of 72fc72630a3fb10ce8c6ebe200c17d0acf8c05ef


liu-wanshun:
@JakeWharton  Why not remove the use of toolchains?

PR #4060, opened 2024-01-26
Ban star imports

JakeWharton:
I don't know why GJF doesn't at least fail on these, but whatever EP to the rescue. Editor config still seems to be allowing IntelliJ to create these, though. Future problem.


Goooler:
And we can add rules:

```editorconfig
ij_java_class_count_to_use_import_on_demand = 999
ij_kotlin_name_count_to_use_star_import = 999
ij_kotlin_name_count_to_use_star_import_for_members = 999
```

JakeWharton:
Oh I thought I sent those. I have a bunch on a branch somewhere. Still have to figure out the minimal set to fix the import order to agree with GJF.

Goooler:
If we can add an option in Spotless JavaExtension?

JakeWharton:
I'd prefer it be in the editorconfig if possible

PR #4065, opened 2024-01-26
Clean up `Platform` to be more maintainable

JakeWharton:
Break up functionality into groups to simplify their configuration and make it easier to see how things change on each platform.

For the interface default method handling, replace reflection on newer JVMs with a multi-release jar and version-based variants of the JVM support. In order to sufficiently test this, the Java tests are now in a sibling module similar to Android, Kotlin, and RoboVM so that they consume the final jar rather than exploded class directories.


retrofit/build.gradle
```diff
@@ -2,6 +2,35 @@ apply plugin: 'java-library'
 apply plugin: 'org.jetbrains.kotlin.jvm'
 apply plugin: 'com.vanniktech.maven.publish'
 
+def addMultiReleaseSourceSet(int version) {
```
swankjesse, line 5: This is fantastic power

retrofit/java-test/README.md
```diff
@@ -0,0 +1,6 @@
+# Retrofit Java Tests
+
+These are in a separate module for two reasons:
+
+- It ensures optional dependencies (Kotlin stuff) are completely absent.
+- It uses the multi-release jar on the classpath rather than only the classes folder.
```
swankjesse, line 6: Thanks for calling this out

retrofit/src/main/java/retrofit2/DefaultMethodSupport.java
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  @Nullable
+  Object invokeDefaultMethod(
+      Method method, Class<?> declaringClass, Object proxy, @Nullable Object[] args)
+      throws Throwable {
+    throw new AssertionError();
+  }
+
+  @TargetApi(24)
```
swankjesse, line 41: Comment why this isn’t `@TargetApi(26)` ?

swankjesse (approved):

retrofit/src/main/java/retrofit2/DefaultMethodSupport.java
JakeWharton: Done

PR #4067, opened 2024-01-29
Import response type keeper

### LINKED ISSUE #3588: Add annotation processor for generating keep rules for types mentioned only inside generics
Otherwise they will be removed by R8 since it cannot determine whether we are doing reflection on them.

Old issue below:

----

Using com.squareup.retrofit2:retrofit:2.9.0

After upgrading gradle plugin from 

```
com.android.tools.build:gradle:4.1.3
```

to

```
com.android.tools.build:gradle:4.2.2
```

I'm getting following error when running with minifyEnabled true:

```
java.lang.IllegalArgumentException: Method return type must not include a type variable or wildcard
```

Here's how looks the method, which generates the error:

```
    @FormUrlEncoded
    @POST("view_ad") //NON-NLS
    Call<GenericDataResponse<EmptyData>> viewAd(
            @Field("session_id") String sessionId, //NON-NLS
            @Field("id") int id //NON-NLS
    );

    public class EmptyData {

        public EmptyData(){}

    }

    public class GenericDataResponse<T> {

        @SerializedName("result") //NON-NLS
        public String result;

        @SerializedName("data") //NON-NLS
        public T data;

        @SerializedName("mess") //NON-NLS
        public String[] messages;
    }
```

In order to fix this I have to either:
- downgrade to the previous gradle plugin version
- add proguard rule: -keep class com.example.app.EmptyData
- disable proguard/r8

I've read the docs about gradle 4.2 changes and nothing seems to be related to proguard directly. 
Can someone explain what's going on?

JakeWharton:
This annotation processor parses all types used as response bodies in Retrofit service methods and adds a keep rule for them. This ensures that even if the type isn't used by callers, it is kept and used to parse the body.

Closes #3588 

We have been using this internally on Cash for a few years now. I've expanded its support to cover all Retrofit annotations as well as written actual tests (gasp!).

### DIFF
--- a/retrofit-response-type-keeper/src/main/kotlin/retrofit2/keeper/RetrofitResponseTypeKeepProcessor.kt
+++ b/retrofit-response-type-keeper/src/main/kotlin/retrofit2/keeper/RetrofitResponseTypeKeepProcessor.kt
@@ -1,0 +1,106 @@
+/*
+ * Copyright (C) 2024 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package retrofit2.keeper
+
+import com.google.auto.service.AutoService
+import javax.annotation.processing.AbstractProcessor
+import javax.annotation.processing.Processor
+import javax.annotation.processing.RoundEnvironment
+import javax.lang.model.SourceVersion
+import javax.lang.model.element.ExecutableElement
+import javax.lang.model.element.TypeElement
+import javax.lang.model.type.DeclaredType
+import javax.lang.model.type.TypeMirror
+import javax.lang.model.type.WildcardType
+import javax.tools.StandardLocation.CLASS_OUTPUT
+import net.ltgt.gradle.incap.IncrementalAnnotationProcessor
+import net.ltgt.gradle.incap.IncrementalAnnotationProcessorType.ISOLATING
+
+@AutoService(Processor::class)
+@IncrementalAnnotationProcessor(ISOLATING)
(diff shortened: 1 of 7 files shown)


retrofit-response-type-keeper/src/main/kotlin/retrofit2/keeper/RetrofitResponseTypeKeepProcessor.kt
```diff
@@ -0,0 +1,106 @@
...
+      val executableElement = method as ExecutableElement
+
+      val serviceType = method.enclosingElement as TypeElement
+      val referenced = elementToReferencedTypes.getOrPut(serviceType, ::LinkedHashSet)
+
+      val returnType = executableElement.returnType as DeclaredType
+      returnType.recursiveParameterizedTypesTo(referenced)
+
+      // Retrofit has special support for 'suspend fun' in Kotlin which manifests as a
+      // final Continuation parameter whose generic type is the declared return type.
```
swankjesse, line 69: Nice comment

retrofit-response-type-keeper/src/main/kotlin/retrofit2/keeper/RetrofitResponseTypeKeepProcessor.kt
```diff
@@ -0,0 +1,106 @@
...
+        .lastOrNull()
+        ?.asType()
+        ?.takeIf { types.erasure(it).toString() == "kotlin.coroutines.Continuation" }
+        ?.let { (it as DeclaredType).typeArguments.single() }
+        ?.recursiveParameterizedTypesTo(referenced)
+    }
+
+    for ((element, referencedTypes) in elementToReferencedTypes) {
+      val typeName = element.qualifiedName.toString()
+      val outputFile = "META-INF/proguard/retrofit-response-type-keeper-$typeName.pro"
```
swankjesse, line 80: This is surprising but makes sense.

retrofit-response-type-keeper/src/test/kotlin/retrofit2/keeper/RetrofitResponseTypeKeepProcessorTest.kt
```diff
@@ -0,0 +1,162 @@
...
+        |-keep,allowobfuscation,allowoptimization class test.Three
+        |-keep,allowobfuscation,allowoptimization class test.Two
+        |
+        """.trimMargin(),
+      )
+  }
+
+  @Test
+  fun kotlinSuspend() {
+    val service = JavaFileObjects.forSourceString(
```
swankjesse, line 127: Sneaky

swankjesse (approved):

PR #4114, opened 2024-03-22
Gracefully handle parse failure with locking

### LINKED ISSUE #4113: Object cannot be cast to class retrofit2.ServiceMethod
I ran into this with Retrofit 2.10.

```
java.lang.ClassCastException: class java.lang.Object cannot be cast to class retrofit2.ServiceMethod (java.lang.Object is in module java.base of loader 'bootstrap'; retrofit2.ServiceMethod is in unnamed module of loader 'app')
	at retrofit2.Retrofit.loadServiceMethod(Retrofit.java:246)
	at retrofit2.Retrofit$1.invoke(Retrofit.java:177)
	at jdk.proxy3/jdk.proxy3.$Proxy139.addPullRequestLabel(Unknown Source)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at misk.client.ClientInvocationHandler.invoke(ClientInvocationHandler.kt:125)
	at jdk.proxy3/jdk.proxy3.$Proxy139.addPullRequestLabel(Unknown Source)
	at com.squareup.cash.treelot.github.PullRequestMaker.addLabels(PullRequestMaker.kt:237)
```

JakeWharton:
Before we would poison the map with the lock object so the second caller would get a weird ClassCastException.

Closes #4113 

---

### DIFF
--- a/retrofit/src/main/java/retrofit2/Retrofit.java
+++ b/retrofit/src/main/java/retrofit2/Retrofit.java
@@ -212,38 +212,53 @@ public final class Retrofit {
   }
 
   ServiceMethod<?> loadServiceMethod(Class<?> service, Method method) {
-    // Note: Once we are minSdk 24 this whole method can be replaced by computeIfAbsent.
-    Object lookup = serviceMethodCache.get(method);
+    while (true) {
+      // Note: Once we are minSdk 24 this whole method can be replaced by computeIfAbsent.
+      Object lookup = serviceMethodCache.get(method);
 
-    if (lookup instanceof ServiceMethod<?>) {
-      // Happy path: method is already parsed into the model.
-      return (ServiceMethod<?>) lookup;
-    }
+      if (lookup instanceof ServiceMethod<?>) {
+        // Happy path: method is already parsed into the model.
+        return (ServiceMethod<?>) lookup;
+      }
 
-    if (lookup == null) {
-      // Map does not contain any value. Try to put in a lock for this method. We MUST synchronize
-      // on the lock before it is visible to others via the map to signal we are doing the work.
-      Object lock = new Object();
-      synchronized (lock) {
-        lookup = serviceMethodCache.putIfAbsent(method, lock);
-        if (lookup == null) {
-          // On successful lock insertion, perform the work and update the map before releasing.
-          // Other threads may be waiting on lock now and will expect the parsed model.
-          ServiceMethod<Object> result = ServiceMethod.parseAnnotations(this, service, method);
-          serviceMethodCache.put(method, result);
-          return result;
+      if (lookup == null) {
+        // Map does not contain any value. Try to put in a lock for this method. We MUST synchronize
+        // on the lock before it is visible to others via the map to signal we are doing the work.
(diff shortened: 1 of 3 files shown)


retrofit/src/main/java/retrofit2/Retrofit.java
```diff
@@ -212,38 +212,53 @@ private void validateServiceInterface(Class<?> service) {
...
+            // On successful lock insertion, perform the work and update the map before releasing.
+            // Other threads may be waiting on lock now and will expect the parsed model.
+            ServiceMethod<Object> result;
+            try {
+              result = ServiceMethod.parseAnnotations(this, service, method);
+            } catch (Throwable e) {
+              // Remove the lock on failure. Any other locked threads will retry as a result.
+              serviceMethodCache.remove(method);
+              throw e;
+            }
```
JakeWharton, line 240: Change is here

retrofit/src/main/java/retrofit2/Retrofit.java
```diff
@@ -212,38 +212,53 @@ private void validateServiceInterface(Class<?> service) {
...
+      // the lock. Once we can take the lock, the map is guaranteed to contain the model or null.
+      // Note: There's a chance that our effort to put a lock into the map has actually returned a
+      // finished model instead of a lock. In that case this code will perform a pointless lock and
+      // redundant lookup in the map of the same instance. This is rare, and ultimately harmless.
+      synchronized (lookup) {
+        Object result = serviceMethodCache.get(method);
+        if (result == null) {
+          // The other thread failed its parsing. We will retry (and probably also fail).
+          continue;
+        }
```
JakeWharton, line 259: And here (plus the enclosing `while (true)`)

swankjesse (approved):

retrofit/src/main/java/retrofit2/Retrofit.java
Endeavour233: I am thinking about whether this situation is possible: `thread 1` put its `lock1`to the cache, granted the right to parse the method, but failed and threw an exception. Then, `thread 2`, which was waiting for `thread 1`  to parse the method, synchronize `lock1` successfully, find the `result` is null, continue the while loop and put its `lock2` to the cache. Then, `Thread 3` synchronize `lock1`. However, the `result` retrieved from the cache by `thread 3` is `lock2` instead of `null`. Thus, `thread 3` try to cast object to ServiceMethod and a `ClassCastException` happens. Can we check `result instanceof ServiceMethod<?>` here? What's more, can we check `lookup instanceof ServiceMethod<?>` before synchronize `lookup` to eliminate the pointless lock and redundant lookup you've metioned in comment(line 253)? Thanks.

PR #4116, opened 2024-03-23
all: remove repetitive words

deterclosed:
---


JakeWharton:
Thanks! This is in the Javadoc of v1 so it could get overwritten if we import it again, but probably not.

PR #4129, opened 2024-04-15
Recompile Java-test/src/test before running it for each supported jdk

### LINKED ISSUE #4072: JDK test tasks do not have dependencies set up correctly
Re-running a test task will not recompile main sources, for example. Probably need to abandon copying the regular test task inputs and set things up ourselves.

Endeavour233:
This pr might resolves #4072 partially. 


shubhamdevgupta (approved):

JakeWharton:
I think #4651 solved this without recompile. In either case, I think I'm going to change to running outside of Gradle when varying the JDK.

PR #4159, opened 2024-05-29
Conditionally-logging interceptor sample

### LINKED ISSUE #4156: Set logging interceptor based on API
As projects grow the number of API requests sent by the application also grows, and hence the debugging of APIs gets more complicated. It would be much convenient to be able to specify API calls by some annotation to output okhttp log. Something like this:

```
@LogInterceptor(DefaultHttpLogInterceptor())
@POST("/users/sign-in-phone")
    suspend fun loginPhone(@Body body: LoginPhoneRequestDto): Response<LoginResponseDTO.LoginSuccessfulResponseDTO>
```
This would print out the log of this specific API instead of whole network connection to server. 
Is it feasible to add such thing to Retrofit?

JakeWharton:
Closes #4156 

---

### DIFF
--- a/samples/src/main/java/com/example/retrofit/ConditionalLoggingInterceptor.kt
+++ b/samples/src/main/java/com/example/retrofit/ConditionalLoggingInterceptor.kt
@@ -1,0 +1,81 @@
+/*
+ * Copyright (C) 2024 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.example.retrofit
+
+import okhttp3.Interceptor
+import okhttp3.OkHttpClient
+import okhttp3.Response
+import okhttp3.ResponseBody
+import okhttp3.logging.HttpLoggingInterceptor
+import okhttp3.mockwebserver.MockResponse
+import okhttp3.mockwebserver.MockWebServer
+import retrofit2.Invocation
+import retrofit2.Retrofit
+import retrofit2.create
+import retrofit2.http.GET
+
+suspend fun main() {
+  val server = MockWebServer()
+  val client = OkHttpClient.Builder()
+    .addInterceptor(
+      ConditionalLoggingInterceptor(
+        HttpLoggingInterceptor(::println).setLevel(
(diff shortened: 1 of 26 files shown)


JakeWharton:
Output:
```
May 29, 2024 11:40:53 AM okhttp3.mockwebserver.MockWebServer$3 execute
INFO: MockWebServer[63842] starting to accept connections
May 29, 2024 11:40:53 AM okhttp3.mockwebserver.MockWebServer$4 processOneRequest
INFO: MockWebServer[63842] received request: GET /one HTTP/1.1 and responded: HTTP/1.1 200 OK
May 29, 2024 11:40:53 AM okhttp3.mockwebserver.MockWebServer$4 processOneRequest
INFO: MockWebServer[63842] received request: GET /two HTTP/1.1 and responded: HTTP/1.1 200 OK
--> GET http://localhost:63842/two
--> END GET
<-- 200 OK http://localhost:63842/two (0ms)
Content-Length: 0
<-- END HTTP (0-byte body)
```

samples/src/main/java/com/example/retrofit/ConditionalLoggingInterceptor.kt
```diff
@@ -0,0 +1,81 @@
...
+  @LogInterceptor
+  @GET("two")
+  suspend fun two(): ResponseBody
+}
+
+/**
+ * Retrofit service functions which are annotated with this class will have their HTTP calls
+ * logged. You must add [ConditionalLoggingInterceptor] to your [OkHttpClient] for this to work.
+ */
+annotation class LogInterceptor
```
mattprecious: nit: Should this just be `Log`? It's kind of weird to think about interceptors as a concept in your service definition.

samples/src/main/java/com/example/retrofit/ConditionalLoggingInterceptor.kt
```diff
@@ -0,0 +1,81 @@
...
+
+class ConditionalLoggingInterceptor(
+  private val loggingInterceptor: HttpLoggingInterceptor,
+) : Interceptor {
+  override fun intercept(chain: Interceptor.Chain): Response {
+    val request = chain.request()
+    request.tag(Invocation::class.java)?.let { invocation ->
+      if (invocation.method().isAnnotationPresent(LogInterceptor::class.java)) {
```
mattprecious: Does the annotation need to have a specific retention for this to reliably work? Also, only an Android issue, but would r8 rules be needed as well?

mattprecious (approved):

samples/src/main/java/com/example/retrofit/ConditionalLoggingInterceptor.kt
JakeWharton: The default is runtime retention which should be fine. These are only samples so R8 is an exercise left for the reader. It should generally be okay without any rules, I think.

samples/src/main/java/com/example/retrofit/ConditionalLoggingInterceptor.kt
JakeWharton: I started with `Log` and it looked weird, but I don't care. Put back.

PR #4199, opened 2024-08-02
Change Gradle depedency on site to be Kotlin friendly

Jawnnypoo:
Change the websites Gradle dependency example to be Kotlin friendly for folks who are using Kotlin Gradle script.


Jawnnypoo:
@JakeWharton Looks like the site never regenerated after this commit, does some additional step need to be done to generate the site with this change?

JakeWharton:
Seems like it's only updated on releases. I have a couple changes to land and then I'll make a new release soon.

Jawnnypoo:
Sounds good, thanks!

Issue #4205, opened 2024-08-12
Retrofit  Kotlin examples


Amir-yazdanmanesh:
Why doesn't Retrofit have any Kotlin examples? Can I write some sample examples and submit a PR?

shubhamdevgupta:
yes you can , just create  request and response class for the api and create a pr

JakeWharton:
There are no Kotlin examples for the same reason there are no Scala or Clojure examples: Retrofit is fundamentally a Java library. Aside from `suspend fun` support, all of the features are the same in Java as they are in Kotlin, and users can trivially convert the Java examples into whichever JVM language they want. The only example I'd accept a PR for in Kotlin would be for showing off `suspend fun`.

PR #4234, opened 2024-10-03
Implement kotlin sample

Amir-yazdanmanesh:
---
Issue Reference: [Retrofit Kotlin examples #4205](https://github.com/square/retrofit/issues/4205)

This PR introduces a Kotlin sample that demonstrates how to use Retrofit in conjunction with RxJava to fetch GitHub repositories for a specified user. 

-  Added a Kotlin example that fetches repositories from the GitHub API using Retrofit and RxJava.


kungpaogao:
+1, this would be nice to have, but could we also use coroutines as well to make it even more Kotlin? (edit: just found this PR #3500)

Also, it would be good to have Kotlin examples on the [homepage](https://square.github.io/retrofit/).

Amir-yazdanmanesh:
> +1, this would be nice to have, but could we also use coroutines as well to make it even more Kotlin? (edit: just found this PR #3500)
> 
> Also, it would be good to have Kotlin examples on the [homepage](https://square.github.io/retrofit/).

While using coroutines would indeed make the implementation more idiomatic to Kotlin, I decided to go with the current libraries in the module to avoid introducing new dependencies. Adding a new library to a project like Retrofit requires the owner's permission, and in this case, I wanted to demonstrate how we can work within the existing setup, using RxJava, which is already integrated. This keeps the implementation aligned with the current project dependencies while still showcasing Kotlin features.

PR #4238, opened 2024-10-11
Fix documentation typo

mrlem:
Self explanatory.

---


JakeWharton:
General consensus is to always use a comma https://english.stackexchange.com/q/16172/7247

mrlem:
@JakeWharton My bad for not checking the rule (which I didn't know, so thank you). In my defence, besides my incomplete knowledge, I had checked it against another occurrence in the same page that doesn't use the comma.

JakeWharton:
Well I mostly have just been copying what I've seen used, so I didn't know either. This was the first time I tried looking up an actual rule because I just as easily could have been the one doing it wrong.

Issue #4297, opened 2025-02-04
Replace use of synchronized in OkHttpCall with ReentrantLock


bendb-instacart:
We use Retrofit in our backend, and I notice that `retrofit2.OkHttpCall` relies extensively on `synchronized(this)`, which can pin host threads (when using virtual threads).  Would you be open to a PR replacing use of `synchronized` here with use of `ReentrantLock`?

JakeWharton:
It might be better to replace it with a compare and swap loop with an atomic? We're not guarding an operation, just performing creation where only a single instance can win. Plus the idea of contention here is near zero, and the cost of double allocation before a winner is determined is negligible.

bendb-instacart:
Thanks for the pointer - I'll take a look!

PR #4301, opened 2025-02-24
Refactor OkHttpCall to use CAS instead of synchronized (#4297)

qwexter:
Replace synchronized blocks with atomic compare-and-swap logic. 
It prevent virtual thread pinning and reduce contention. 
See https://github.com/square/retrofit/issues/4297 for more details.

---


retrofit/src/main/java/retrofit2/OkHttpCall.java
```diff
@@ -107,10 +103,12 @@ private okhttp3.Call getRawCall() throws IOException {
 
     // Create and remember either the success or the failure.
     try {
-      return rawCall = createRawCall();
+      final okhttp3.Call newCall = createRawCall();
+      rawCallRef.compareAndSet(null, newCall);
+      return rawCallRef.get();
```
JakeWharton, line 108:
```suggestion
      return rawCallRef.compareAndSet(null, newCall)
          ? newCall
          : rawCallRef.get();
```

retrofit/src/main/java/retrofit2/OkHttpCall.java
```diff
@@ -119,22 +117,19 @@ private okhttp3.Call getRawCall() throws IOException {
...
-        }
+    if (!executed.compareAndSet(false, true)) throw new IllegalStateException("Already executed.");
+
+    okhttp3.Call call = rawCallRef.get();
+    Throwable failure = creationFailureRef.get();
+
+    if (call == null && failure == null) {
+      try {
+        call = createRawCall();
+        rawCallRef.compareAndSet(null, call);
```
JakeWharton, line 128: This code is now racy. Another thread could be performing the same operation in parallel and its `okhttp3.Call` might "win" causing this instance to be a duplicate.

JakeWharton (commented):
Still think this needs some work. I would probably look to simplify things, such as merge the failure and call reference into one, and potentially even include the cancel state as `new Object()` tombstone reference. I need to spend more time with this code before I can move forward with it.

retrofit/src/main/java/retrofit2/OkHttpCall.java
harrytmthy, line 128: Circling back on this in case it's still active. As @JakeWharton noted, another thread could enter this block and call `createRawCall()`. Even though only one `compareAndSet(...)` might succeed, the other still ends up creating a duplicate instance.

Hi @qwexter, just curious if this race condition was considered acceptable here, or if there's a plan to mitigate it without reintroducing locking?

qwexter: Yes, I agree. I think I agree that the winner-takes-all approach mentioned in the discussion is too straightforward.

I didn't have any issues with my unit tests, but I might have missed something.

qwexter: I agree. If you have some ideas, pls share it. I'll hope I can get back to this with more resources at the end of September. If it will be needed at that time.

Issue #4226, opened 2024-09-19
CWE-295 | Improper Certificate Validation


vivek-np:
What kind of issue is this?

 ## CWE-295 | Improper Certificate Validation
 
 -checkmarx Vulnerability detected for Retrofit Converters  **com.squareup.retrofit2:converter-gson @ 2.11.0**
 
 This is happening as internally it depends on  **com.squareup.okhttp3:okhttp @ 3.14.9**
 
 

> In "verifyHostName" method of "OkHostnameVerifier.java", there is a possible way to accept a certificate for the wrong domain due to improperly used crypto. This could lead to remote information disclosure with no additional execution privileges needed. User interaction is not needed for exploitation.Product:

JakeWharton:
We are not upgrading OkHttp at this time. See https://github.com/square/retrofit/issues/4020#issuecomment-1870586604 for the plan.

You are welcome to upgrade the version of OkHttp in your builds.

PR #4313, opened 2025-03-05
Bump okhttp to resolve security vulnerabilities

shybovycha:
Bump `okhttp3` to resolve security vulnerabilities introduced by its transitive dependency, `okio`: [`okhttp3:3.14.9`](https://mvnrepository.com/artifact/com.squareup.okhttp3/okhttp/3.14.9) -> [`okio:1.17.2`](https://mvnrepository.com/artifact/com.squareup.okio/okio/1.17.2)

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -14,7 +14,7 @@
 
 [versions]
 kotlin = "2.1.10"
-okhttp = "3.14.9"
+okhttp = "4.12.0"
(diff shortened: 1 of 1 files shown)


JakeWharton:
Please see https://github.com/square/retrofit/issues/4226#issuecomment-2360997965 and then the comment linked from it as to why OkHttp is still on 3.x. This does not prevent you from upgrading the version in your builds.

PR #4462, opened 2025-08-08
feature: SSE endpoints

### LINKED ISSUE #1029: Support for server sent events
Hello,
I am trying to consume the following using retrofit

https://mesosphere.github.io/marathon/docs/rest-api.html#get-v2-events
I could not find a suitable command/documentation for consuming server sent events. Does retrofit currently support it ? If not is it on the roadmap for future releases.

Thansk!

MukjepScarlet:
Prototype (my demand):
```kotlin
@SSE
@GET("events")
suspend fun fetchEvents(): Flow<ServerSentEvent>

data class ServerSentEvent(
    val id: String?,
    val type: String?
    val data: String,
)
```

Pending:
1. Should it be migrated into base module? (e.g. `@SSE` can replace `@GET`)
2. What kind of result should be supported?
3. Write in Kotlin or Java?
4. Auto-reconnecting like JS? Or improve okhttp-sse module.
5. How can we handle connection events like open/disconnected here?
6. How to handle `id`, `type`, `data` for example `Int?` for `id`, enum for `type` and JSON `data`? 

`java.util.concurrent.Flow` (Java 9+) / `java.util.stream.Stream` (Java 8+) might be the integrated ones... But `Stream` is not designed for tasks like this.

These questions need to be answered before actual works.

---

closes https://github.com/square/retrofit/issues/1029


JakeWharton:
I think this can probably be entirely implemented in a call adapter with no changes to the core, but I'd have to try it.

MukjepScarlet:
We might have more than one adapter for SSE (for example RxJava & Kotlinx coroutines), so the basic definition like `ServerSentEvent` (or `ServerSentEvent<T>`) probably need to be in the core module.

JakeWharton:
Call adapters can delegate to each other. They do not need to be in the core.

MukjepScarlet:
I planned to add a callback style for `core` module but it requires some improvements on `okhttp-sse`, another PR for that will be opened later.

MukjepScarlet:
@JakeWharton Please review current work of this for further change/improvement/completion.

MukjepScarlet:
Current usage:
```kotlin
@Streaming
@GET
fun ktxFlow(): Flow<ServerSentEvent<String, String, String>>

@Streaming
@GET
fun jucFlow(): Flow.Publisher<ServerSentEvent<String, String, String>>

@Streaming
@GET
fun callback(): EventSource<String, String, String>
```

The fields of `ServerSentEvent` will be converted as a String ResponseBody.

PR #4590, opened 2025-12-07
Fix charset in README example for Retrofit

CoryCharlton:
The correct value for charset is "utf-8". Some systems may tolerate "utf8" but not the all of them.

Sources:

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Type#syntax

---


JakeWharton (approved):
Thanks! I suspect I copy/pasted this from somewhere. Hopefully the original place was also updated at some point!

PR #4595, opened 2025-12-11
Support QUERY method

Goooler:
~~We can't name it to `QUERY` due to the case-insensitive file systems on MacOS and Windows. Should we rename the annotation to `SEARCH` or something else?~~

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
  - Add explicit keep rules for RxJava `Result` types to prevent their generic information from being removed.
  - Add `allowoptimization` flags for most kept types.
  - Add `Invocation.annotationUrl` which returns the original URL from the method annotation.
+ - Support `QUERY` method (annotate your methods with `@QUERY_METHOD`).
 
 **Changed**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
  - Add explicit keep rules for RxJava `Result` types to prevent their generic information from being removed.
  - Add `allowoptimization` flags for most kept types.
  - Add `Invocation.annotationUrl` which returns the original URL from the method annotation.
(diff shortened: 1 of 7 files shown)


JakeWharton:
You can write it in Kotlin where the file name doesn't need to match the class name. The case collision isn't a problem for people consuming the jar.

Goooler:
I tried, grouped `@Query` and `@QUERY` in the same kt file like:

```kt
@MustBeDocumented
@Target(
  AnnotationTarget.FUNCTION,
  AnnotationTarget.PROPERTY_GETTER,
  AnnotationTarget.PROPERTY_SETTER,
)
@Retention(AnnotationRetention.RUNTIME)
annotation class QUERY(
  val value: String = ""
)

@MustBeDocumented
@Target(AnnotationTarget.VALUE_PARAMETER)
@Retention(AnnotationRetention.RUNTIME)
annotation class Query(
  val value: String,
  val encoded: Boolean = false,
)
```

and ran `gw assemble`:

```
/Users/goooler/IdeaProjects/retrofit/retrofit/src/main/java/retrofit2/Converter.java:33: error: cannot find symbol
import retrofit2.http.Query;
                     ^
  symbol:   class Query
  location: package retrofit2.http
```

Goooler:
We can't leave the single `QUERY` in the `QUERY_METHOD.kt`, as the generated `QUERY.class` still conflicts with `Query.class`. May need to add a new uber jar configuration to fix the classes bundling.

Goooler:
I added a new source set and moved the `QUERY.java` into it to avoid the conflict.

JakeWharton:
Ugh gotta do something for RoboVM

Goooler:
I took a look at the RoboVM Gradle plugin and it doesn't seem to offer an easy way to support case-insensitive features.

JakeWharton:
I'm just deleting it. It's time.

PR #4623, opened 2026-01-14
Fix malformed URL error when relative path contains colon in first segment

### LINKED ISSUE #3080: Invalid Malformed URL
The path of a URL contains a colon symbol which is causing the call to fail.

Using version com.squareup.retrofit2:retrofit:2.5.0

Code:
```
public interface ClientApi {
    @PUT("user:email={email}/login")
    Call<ResponseBody> login(@Path("email") String email, @Body String pass);

    @PUT("./user:email={email}/login")
    Call<ResponseBody> loginRelative(@Path("email") String email, @Body String pass);
...
```

**Regular path**
Stack trace:
```
Caused by: java.lang.IllegalArgumentException: Malformed URL. Base: https://test.com/restapi/, Relative: user:email=me@test.com/login
        at retrofit2.RequestBuilder.get(RequestBuilder.java:221)
        at retrofit2.RequestFactory.create(RequestFactory.java:111)
        at retrofit2.OkHttpCall.createRawCall(OkHttpCall.java:190)
        at retrofit2.OkHttpCall.execute(OkHttpCall.java:173)
        at Service.login(Service.java:38)
```
Think the code needs to treat the relative portion of the URL like an absolute URL when parsing.

**Relative path**
Using a retaliative path  annotation `./` does not solve this issue.
Stack trace:
```
Caused by: java.lang.IllegalArgumentException: @Path parameters shouldn't perform path traversal ('.' or '..'):
        at retrofit2.RequestBuilder.addPathParam(RequestBuilder.java:113)
        at retrofit2.ParameterHandler$Path.apply(ParameterHandler.java:99)
        at retrofit2.RequestFactory.create(RequestFactory.java:108)
        at retrofit2.OkHttpCall.createRawCall(OkHttpCall.java:190)
        at retrofit2.OkHttpCall.execute(OkHttpCall.java:173)
```

Bug report unit test
https://gist.github.com/nathancassano/f675e2fd1c64d4b6002afea37bf5fee4

abdellahi-brahim:
## Summary

This change addresses #3080, where Retrofit rejects a relative endpoint whose first path segment contains a colon.

For example:

```java
@PUT("user:email={email}/login")
Call<ResponseBody> login(@Path("email") String email);
```

After substituting the path parameter, Retrofit asks OkHttp to resolve:

```text
user:email=me@test.com/login
```

URI parsers interpret `user:` as a scheme rather than part of a relative path, so OkHttp cannot resolve it against the HTTP base URL.

RFC 3986 recommends prefixing this kind of relative path with `./`. The implementation now performs that disambiguation internally for annotation URLs containing `@Path` parameters:

```text
user:email=me@test.com/login
./user:email=me@test.com/login
```

The `./` segment disappears during URL resolution. With a base URL of `http://example.com/api/`, the final URL is:

```text
http://example.com/api/user:email=me@test.com/login
```

The colon remains literal. It is not changed to `%3A`.

## Implementation

`RequestFactory` records whether the annotation URL contains path parameters and passes that information to `RequestBuilder`.

Before resolving a templated relative URL, `RequestBuilder` checks whether its first path segment contains a colon. If it does, Retrofit prefixes the URL with `./`.

This happens in both request-building paths:

- Direct URL resolution when no query parameters are added.
- `HttpUrl.Builder` creation when Retrofit adds query parameters.

The change also permits users to write the RFC form explicitly:

```java
@PUT("./user:email={email}/login")
```

The path traversal check ignores this static leading `./`, but it continues rejecting `.` or `..` supplied through an `@Path` value.

## Preserved behavior

The disambiguation does not apply to:

- Dynamic `@Url` values.
- Absolute HTTP or HTTPS annotation URLs.
- URLs beginning with `/`.
- Colons after the first path segment.
- Colons in the query or fragment.

Base URL path segments are preserved.

## Open question

The string itself cannot always reveal the caller’s intent.

For example:

```java
@GET("mailto:{value}")
```

Because it contains an `@Path` parameter, the current implementation treats this as a relative HTTP path and resolves it as:

```text
http://example.com/mailto:user@example.com
```

Without a path parameter, `mailto:user@example.com` remains an unsupported URI and fails.

The design choice is whether Retrofit should infer that templated scheme-like strings are relative paths, or require callers to write the unambiguous `./` prefix themselves. Requiring `./` is stricter and follows RFC 3986 directly, but it would not make the original annotation from #3080 work unchanged.

## Validation

The tests cover literal colon preservation, base URL paths, added query parameters, explicit `./`, traversal rejection, absolute URLs, and query or fragment boundaries.

The following checks pass:

```bash
./gradlew \
  :retrofit:spotlessJavaCheck \
  :retrofit:java-test:spotlessJavaCheck \
  :retrofit:java-test:testJdk21
```

### DIFF
--- a/retrofit/java-test/src/test/java/retrofit2/RequestFactoryTest.java
+++ b/retrofit/java-test/src/test/java/retrofit2/RequestFactoryTest.java
@@ -1431,21 +1431,85 @@ public final class RequestFactoryTest {
 
   @Test
   public void getWithColonInRelativeUrlFirstSegment() {
-    // Regression test for https://github.com/square/retrofit/issues/3080
-    // A colon in the first segment of a relative URL (before the first slash) can be
-    // misinterpreted as a URL scheme separator. This test ensures such colons are encoded.
     class Example {
       @PUT("user:email={email}/login") //
-      Call<ResponseBody> method(@Path("email") String email, @Body String pass) {
+      Call<ResponseBody> method(@Path("email") String email) {
         return null;
       }
     }
 
-    Request request = buildRequest(Example.class, "me@test.com", "password");
+    Retrofit.Builder retrofitBuilder = new Retrofit.Builder().baseUrl("http://example.com/api/");
+    Request request = buildRequest(Example.class, retrofitBuilder, "me@test.com");
     assertThat(request.method()).isEqualTo("PUT");
-    // Colon in first path segment encoded as %3A to prevent scheme misinterpretation
     assertThat(request.url().toString())
-        .isEqualTo("http://example.com/user%3Aemail=me@test.com/login");
+        .isEqualTo("http://example.com/api/user:email=me@test.com/login");
+  }
+
+  @Test
+  public void getWithExplicitDotSegmentAndColonInRelativeUrlFirstSegment() {
+    class Example {
+      @PUT("./user:email={email}/login") //
+      Call<ResponseBody> method(@Path("email") String email) {
+        return null;
+      }
+    }
+
+    Request request = buildRequest(Example.class, "me@test.com");
+    assertThat(request.url().toString())
+        .isEqualTo("http://example.com/user:email=me@test.com/login");
+  }
+
+  @Test
+  public void explicitDotSegmentStillRejectsPathParameterTraversal() {
+    class Example {
+      @GET("./{path}") //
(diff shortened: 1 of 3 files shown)


JakeWharton:
Per our contribution document we do not allow LLM submissions.

PR #4648, opened 2026-02-03
Update base URL in Retrofit docs

ColtonIdle:
Jake on slack said

> You should always have a trailing slash on the base URL, because paths are relative and are resolved against the base URL. Thus, anything without a trailing slash is effectively ignored.

> Consider a base URL of https://example.com/api (no slash) and then a path of foo. This would result in https://example.com/foo which makes the api part of the original base URL useless.

> So either you meant https://example.com/ or you meant https://example.com/api/

I think the very first example users see should have the trailing /

---

### DIFF
--- a/website/src/content/docs/index.md
+++ b/website/src/content/docs/index.md
@@ -15,7 +15,7 @@ The `Retrofit` class generates an implementation of the `GitHubService` interfac
 
 ```java
 Retrofit retrofit = new Retrofit.Builder()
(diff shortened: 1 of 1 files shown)


JakeWharton:
I was wrong that this was invalid. OkHttp will parse that URL which lacks a path component but still report that its path component is `/` when asked. However, there's no harm in specifying it explicitly.

PR #4656, opened 2026-02-15
Replace synchronized blocks in OkHttpCall with atomic CAS   

### LINKED ISSUE #4297: Replace use of synchronized in OkHttpCall with ReentrantLock

Amir-yazdanmanesh:
## Summary                                                                                                                                                               
  - Replace `synchronized(this)` blocks in `OkHttpCall` with lock-free atomic operations to avoid pinning virtual threads (Project Loom)                                   
  - Merge `rawCall` and `creationFailure` fields into a single `AtomicReference<Object> rawCallState` with state transitions `null → Call` or `null → Throwable`           
  (write-once semantics)
  - Replace `boolean executed` with `AtomicBoolean` using `compareAndSet` for the execute-once guard                                                                       
                                                                                                                                                                           
  Closes #4297

  ## Motivation
  `synchronized` blocks pin virtual threads to platform threads, defeating the purpose of Project Loom. Since the guarded code protects single-instance creation with
  minimal contention, atomic CAS is a better fit than `ReentrantLock`. See #4297.

  ## Design
  The key insight is merging `rawCall` + `creationFailure` into one `AtomicReference`. This avoids the race condition from #4301 where separate references could get out of
   sync. Once the reference is set to a non-null value (either a `Call` or a `Throwable`), it never changes.

  If two threads race to create the call, both call `createRawCall()` but only one CAS wins. The loser uses the winner's result. This is safe because `createRawCall()` is
  side-effect-free.

  ## Test plan
  - [x] `./gradlew :retrofit:java-test:check` — all tests pass across JDK 8, 10-24 (including concurrency tests in CallTest.java)

### DIFF
--- a/retrofit/java-test/build.gradle
+++ b/retrofit/java-test/build.gradle
@@ -25,10 +25,9 @@ tasks.withType(JavaCompile).configureEach {
     description = "Runs the test suite on JDK $majorVersion"
     group = LifecycleBasePlugin.VERIFICATION_GROUP
 
-    // Copy inputs from normal Test task.
-    def testTask = tasks.getByName("test")
-    classpath = testTask.classpath
-    testClassesDirs = testTask.testClassesDirs
+    // Wire directly to the test source set outputs for proper task dependencies.
+    testClassesDirs = sourceSets.test.output.classesDirs
+    classpath = sourceSets.test.runtimeClasspath
   }
   tasks.named("check").configure {
     dependsOn(jdkTest)
--- a/retrofit/src/main/java/retrofit2/OkHttpCall.java
+++ b/retrofit/src/main/java/retrofit2/OkHttpCall.java
@@ -19,8 +19,9 @@ import static retrofit2.Utils.throwIfFatal;
 
 import java.io.IOException;
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 import javax.annotation.Nullable;
-import javax.annotation.concurrent.GuardedBy;
 import okhttp3.MediaType;
 import okhttp3.Request;
 import okhttp3.ResponseBody;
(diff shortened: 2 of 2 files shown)


JakeWharton:
The locks don't guard long-running work, only state changes. I don't think this is a big deal. Has anyone demonstrated an actual problem or are we just speculating?

Amir-yazdanmanesh:
> The locks don't guard long-running work, only state changes. I don't think this is a big deal. Has anyone demonstrated an actual problem or are we just speculating?


  You're right, the synchronized blocks here are short and don't guard long-running work, so pinning should be brief. I don't have benchmarks demonstrating an actual
  throughput problem.

  The change was motivated by the general guidance to avoid synchronized with virtual threads, but I understand that without evidence of a real issue, the added complexity
   of CAS may not be justified.

  Happy to close this if you'd prefer to keep the current approach until a concrete problem is reported.

JakeWharton:
I need to review it more closely. The last attempt at this had subtle races because it used two atomic states. This one is different, but it still warrants exhaustively thinking through the state transitions in the context of concurrency.

Amir-yazdanmanesh:
> I need to review it more closely. The last attempt at this had subtle races because it used two atomic states. This one is different, but it still warrants exhaustively thinking through the state transitions in the context of concurrency.

Happy to walk through the state transitions while you review:
The main thing I'd flag is the cancel window, if cancel() is called after executed CAS succeeds but before rawCallState is set, it reads null and silently does nothing. The original synchronized approach probably had the same gap, but worth confirming there wasn't a canceled boolean that got checked post-creation.
The two-thread race in getRawCall() should be fine as long as createRawCall() has no side effects, both threads may call it, but only the CAS winner's result survives. I believe that holds, but interceptors/event listeners are worth a glance.
Let me know if a test for the cancel-before-create scenario would help

PR #4706, opened 2026-07-18
Update Retrofit link in README.md

mknudson94:
Update away from broken `square` link in readme


JakeWharton:
Thanks! I just merged #4704 though which covers many more than just this single link.

PR #4755, opened 2026-09-02
Replace synchronized in OkHttpCall with atomics

### LINKED ISSUE #4297: Replace use of synchronized in OkHttpCall with ReentrantLock

HuzaifaChaudary:
Closes #4297

`OkHttpCall` guarded three fields with the object monitor: the lazily created `okhttp3.Call`, the throwable that creating it threw, and the `executed` flag. Holding a monitor across call creation pins the carrier thread when the caller is a virtual thread, which is what the issue is about.

Following the suggestion on the issue, this uses atomics rather than a `ReentrantLock`.

`executed` becomes an `AtomicBoolean`, so the one-shot check in `enqueue` and `execute` is a single `compareAndSet` instead of a read and a write under a lock.

`rawCall` and `creationFailure` become a single `AtomicReference` holding whichever of the two happened. Exactly one of them does, and both have to be remembered so later callers see the same outcome, so one field models it better than two. Creation is no longer serialised: two threads arriving together will both create a call and one wins the swap. An unexecuted `okhttp3.Call` holds no connection, so discarding the loser costs nothing, which is the tradeoff you described.

`request()`, `timeout()`, `isExecuted()`, `cancel()` and `isCanceled()` no longer take the monitor at all.

No new test. The behaviour is unchanged and the existing suite already covers it, including the "Already executed." case in `CallTest`, and a concurrency test for this would be timing dependent without proving much. `:retrofit:java-test:testJdk21` passes, 343 tests across 15 classes, and nothing outside `OkHttpCall` referenced either removed field.

### DIFF
--- a/retrofit/src/main/java/retrofit2/OkHttpCall.java
+++ b/retrofit/src/main/java/retrofit2/OkHttpCall.java
@@ -19,8 +19,9 @@ import static retrofit2.Utils.throwIfFatal;
 
 import java.io.IOException;
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 import javax.annotation.Nullable;
-import javax.annotation.concurrent.GuardedBy;
 import okhttp3.MediaType;
 import okhttp3.Request;
 import okhttp3.ResponseBody;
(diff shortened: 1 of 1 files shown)


HuzaifaChaudary:
You are right, and I apologise. Closing this.

CONTRIBUTING.md says it plainly and I did not read it before opening the PR. That is on me, there is no excuse for it. I checked the issue for the `PR welcome` label and for whether anyone else had claimed it, and I never opened the contributing guide. It will not happen again on this project.

On the substance, you are right there too. You said on the issue that contention here is near zero and this is trivial CPU work, so removing the monitor was not worth the churn in the first place.

PR #4761, opened 2026-09-08
Add an optional marker interface to keep empty Retrofit services

### LINKED ISSUE #4295: Interface without any methods not handled by Retrofit consumer keep rules
A question from the R8 team. We recently handled https://issuetracker.google.com/391260908, which turned out to be an interface passed to `create` in Retrofit, which needed to be kept. The consumer keep rules already have rules to keep Retrofit interfaces, https://github.com/square/retrofit/blob/trunk/retrofit/src/main/resources/META-INF/proguard/retrofit2.pro#L31, says
```
-if interface * { @retrofit2.http.* <methods>; }
-keep,allowobfuscation interface <1>
```
However, in the case of the issue the interface had no methods (and then trivially no methods annotated with a `retrofit2.http.*` annotation).

Is an interface without any methods a relevant use case for Retrofit `create`?

iVamsi:
Fixes #4295 

---

### DIFF
--- a/retrofit/src/main/java/retrofit2/RetrofitService.java
+++ b/retrofit/src/main/java/retrofit2/RetrofitService.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright (C) 2026 Retrofit Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


retrofit/src/main/java/retrofit2/RetrofitService.java
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright (C) 2026 Retrofit Authors
```
iVamsi, line 2: Followed the same pattern as OkHttp. "Retrofit Authors" - let me know if this needs to be something different.

